PR #5, opened 2015-05-08
force one content provider per install

### LINKED ISSUE #3: If two apps are installed with this lib they'll clash contentprovider names
Meaning the second won't work?

https://github.com/novoda/download-manager/blob/master/library/src/main/AndroidManifest.xml#L14-L17

blundell:
Fixes #3 

Now if a client wishes to use Download-Manager they have to declare `manifestPlaceholders = [downloadAuthority: "WATEVER_U_WANT"]` and then this will be used as the name of the content provider authority allowing, multiple apps using this library to be installed at the same time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,19 @@ dependencies {
 
 Simple example source code can be found in this demo module: [Android Simple Demo][2]
 
+You will need to add this to your Android application build.gradle:
+
+```groovy
+
+android { 
+  defaultConfig {
+       // other things
+       
+       manifestPlaceholders = [downloadAuthority: "${applicationId}"]
+  }
+}
+```
+
 
 ## Links
 
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -24,9 +24,12 @@ android {
 
         def projects = rootProject.getProject().getChildProjects()
         projects.each {
-            def authority = it.value.android.defaultConfig.manifestPlaceholders.downloadAuthority
-            if (authority) {
-                buildConfigField "String", "DOWNLOAD_AUTHORITY", "\"${authority}\""
+            def project = it.value
+            if (project.hasProperty('android')) {
+                def authority = project.android.defaultConfig.manifestPlaceholders.downloadAuthority
+                if (authority) {
+                    buildConfigField "String", "DOWNLOAD_AUTHORITY", "\"${authority}\""
+                }
             }
         }
     }
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -21,6 +21,14 @@ android {
     defaultConfig {
         minSdkVersion 14
         targetSdkVersion 22
+
+        def projects = rootProject.getProject().getChildProjects();
+        projects.each {
+            def authority = it.value.android.defaultConfig.manifestPlaceholders.downloadAuthority
+            if(authority) {
+                buildConfigField "String", "DOWNLOAD_AUTHORITY", "\"${authority}\""
+            }
+        }
     }
 
     lintOptions {
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -25,7 +25,7 @@ android {
         def projects = rootProject.getProject().getChildProjects();
         projects.each {
             def authority = it.value.android.defaultConfig.manifestPlaceholders.downloadAuthority
-            if(authority) {
+            if (authority) {
                 buildConfigField "String", "DOWNLOAD_AUTHORITY", "\"${authority}\""
             }
         }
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -25,6 +25,19 @@ dependencies {
...
 
+You will need to add this to your Android application build.gradle:
+
+```groovy
+
+android { 
+  defaultConfig {
+       // other things
+       
+       manifestPlaceholders = [downloadAuthority: "${applicationId}"]
```
ouchadam: :100:

ouchadam:
nice fix, LGTM :+1:

library/build.gradle
```diff
@@ -21,6 +21,14 @@ android {
     defaultConfig {
         minSdkVersion 14
         targetSdkVersion 22
+
+        def projects = rootProject.getProject().getChildProjects();
+        projects.each {
+            def authority = it.value.android.defaultConfig.manifestPlaceholders.downloadAuthority
+            if(authority) {
```
xrigau: dis formatting (missing space)

library/build.gradle
```diff
@@ -21,6 +21,14 @@ android {
     defaultConfig {
         minSdkVersion 14
         targetSdkVersion 22
+
+        def projects = rootProject.getProject().getChildProjects();
```
devisnik: `;` not needed ;)

library/build.gradle
blundell: https://github.com/novoda/download-manager/commit/6d8cdf59990edf5b71bddb340648ef9e4f3125dc

library/build.gradle
blundell: https://github.com/novoda/download-manager/commit/f2060261cd1f944ce8da76862f735566f6c81ae3

library/build.gradle
```diff
@@ -21,6 +21,14 @@ android {
...
+        projects.each {
+            def authority = it.value.android.defaultConfig.manifestPlaceholders.downloadAuthority
```
devisnik: What is `value`?

library/build.gradle
```diff
@@ -21,6 +21,14 @@ android {
     defaultConfig {
         minSdkVersion 14
         targetSdkVersion 22
+
+        def projects = rootProject.getProject().getChildProjects()
+        projects.each {
+            def authority = it.value.android.defaultConfig.manifestPlaceholders.downloadAuthority
```
devisnik: I wonder if `?.` instead of just `.` might make that more robust when having additional modules.

Also `rootProject.subProjects` returns a set of projects so that might avoid the `it.value`, `it` would be the project then already. Or do you rely on the map of projects for some reason?

devisnik:
:+1: Pretty slick.

library/build.gradle
blundell: yeah I was thinking that as well, for ex non android modules.

Nope I can go with that, map was first solution found :)

blundell: `def projects = rootProject.subProjects()`

 Could not find method subProjects() for arguments [] on root project

library/build.gradle
```diff
@@ -21,6 +21,17 @@ android {
     defaultConfig {
         minSdkVersion 14
         targetSdkVersion 22
+
+        def projects = rootProject.getProject().getChildProjects()
+        projects.each {
+            def project = it.value
+            if (project.hasProperty('android')) {
```
blundell: @devisnik had to do it this way to avoid java modules

no need for the `.?` dereferencing below, we'd want it to crash if `downloadAuthority` is not set

library/build.gradle
xrigau: @blundell do you even documentation?

http://gradle.org/docs/current/dsl/org.gradle.api.Project.html#org.gradle.api.Project:subprojects

blundell: I clearly don't understand how to peruse gradle docs. I never know if it's a groovy or gradle thing

blundell: still don't get it, docs say I should be able to ref it ... ah right lowercase `P` phht

xrigau: yeah and also you need to decide use getter `.getSubprojects()` or property `.subprojects`

blundell:
This is ready to ship, PR's, iterations welcome

library/build.gradle
blundell: fact
```groovy
def projects = rootProject.subprojects
        projects.each { project ->
            if (project.hasProperty('android')) {
                def authority = project.android.defaultConfig.manifestPlaceholders.downloadAuthority
                if (authority) {
                    buildConfigField "String", "DOWNLOAD_AUTHORITY", "\"${authority}\""
                }
            }
        }
```

devisnik: :+1:

PR #10, opened 2015-05-11
Retrieve authority by reflection

amlcurran:
We need the authority to be (somehow) statically defined by the app consuming our library. In order to get going quickly, we've used reflection to achieve this.

Hopefully in the future we'll be able to generate it using a gradle plug-in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,19 @@ dependencies {
 
 Simple example source code can be found in this demo module: [Android Simple Demo][2]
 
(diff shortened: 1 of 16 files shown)


blundell:
:+1: quick and dirty till we can think of something more ... classy

xrigau:
such HAX

rock3r:
@xrigau awesome gif

PR #13, opened 2015-05-12
Feature/Use authority as a field

friedger:
:x: 
This PR is to visualise the impact of not using Reflector but defining the authority from the context.

### DIFF
--- a/demo/build.gradle
+++ b/demo/build.gradle
@@ -19,7 +19,7 @@ android {
         versionCode 1
         versionName "1.0"
 
-        manifestPlaceholders = [downloadAuthority: "${applicationId}"]
+        manifestPlaceholders = [downloadAuthority: "${applicationId}.downloads"]
     }
 
     lintOptions {
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
@@ -26,6 +26,7 @@ import java.util.concurrent.Future;
  */
 public class DownloadInfo {
     public static final String EXTRA_EXTRA = "com.novoda.download.lib.KEY_INTENT_EXTRA";
+    private String mAuthority;
 
     // TODO: move towards these in-memory objects being sources of truth, and
 
@@ -34,14 +35,16 @@ public class DownloadInfo {
     public static class Reader {
         private ContentResolver mResolver;
         private Cursor mCursor;
+        private String mAuthority;
 
-        public Reader(ContentResolver resolver, Cursor cursor) {
+        public Reader(ContentResolver resolver, Cursor cursor, String authority) {
             mResolver = resolver;
             mCursor = cursor;
+            mAuthority = authority;
         }
 
         public DownloadInfo newDownloadInfo(Context context, SystemFacade systemFacade, StorageManager storageManager, DownloadNotifier notifier) {
-            final DownloadInfo info = new DownloadInfo(context, systemFacade, storageManager, notifier);
+            final DownloadInfo info = new DownloadInfo(context, systemFacade, storageManager, notifier, mAuthority);
             updateFromDatabase(info);
             readRequestHeaders(info);
             return info;
@@ -89,7 +92,7 @@ public class DownloadInfo {
 
         private void readRequestHeaders(DownloadInfo info) {
             info.mRequestHeaders.clear();
-            Uri headerUri = Uri.withAppendedPath(info.getAllDownloadsUri(), Downloads.Impl.RequestHeaders.URI_SEGMENT);
+            Uri headerUri = Uri.withAppendedPath(info.getAllDownloadsUri(mAuthority), Downloads.Impl.RequestHeaders.URI_SEGMENT);
             Cursor cursor = mResolver.query(headerUri, null, null, null, null);
             try {
                 int headerIndex =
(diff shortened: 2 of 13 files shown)


library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
```diff
@@ -479,12 +483,12 @@ public boolean isOnCache() {
                 || mDestination == Downloads.Impl.DESTINATION_CACHE_PARTITION_PURGEABLE);
     }
 
-    public Uri getMyDownloadsUri() {
-        return ContentUris.withAppendedId(Downloads.Impl.CONTENT_URI, mId);
+    public Uri getMyDownloadsUri(String authority) {
```
blundell: this is publicly accessible, we don't want the client to have to pass their authority. Either we make this method package? or we do somethign else

library/src/main/java/com/novoda/downloadmanager/lib/DownloadProvider.java
```diff
@@ -118,27 +119,32 @@
...
-        sURIMatcher.addURI(AUTHORITY, "all_downloads", ALL_DOWNLOADS);
-        sURIMatcher.addURI(AUTHORITY, "all_downloads/#", ALL_DOWNLOADS_ID);
-        sURIMatcher.addURI(AUTHORITY, "my_downloads/#/" + Downloads.Impl.RequestHeaders.URI_SEGMENT, REQUEST_HEADERS_URI);
-        sURIMatcher.addURI(AUTHORITY, "all_downloads/#/" + Downloads.Impl.RequestHeaders.URI_SEGMENT, REQUEST_HEADERS_URI);
-        // temporary, for backwards compatibility
-        sURIMatcher.addURI(AUTHORITY, "download", MY_DOWNLOADS);
-        sURIMatcher.addURI(AUTHORITY, "download/#", MY_DOWNLOADS_ID);
-        sURIMatcher.addURI(AUTHORITY, "download/#/" + Downloads.Impl.RequestHeaders.URI_SEGMENT, REQUEST_HEADERS_URI);
-        sURIMatcher.addURI(AUTHORITY, Downloads.Impl.PUBLICLY_ACCESSIBLE_DOWNLOADS_URI_SEGMENT + "/#", PUBLIC_DOWNLOAD_ID);
+    public static void initURIMatcher(Context context) {
```
blundell: `private`

library/src/main/java/com/novoda/downloadmanager/lib/DownloadProvider.java
```diff
@@ -118,27 +119,32 @@
...
+                    Downloads.Impl.ALL_DOWNLOADS_CONTENT_URI(AUTHORITY),
+            };
+
+            sURIMatcher.addURI(AUTHORITY, "my_downloads", MY_DOWNLOADS);
+            sURIMatcher.addURI(AUTHORITY, "my_downloads/#", MY_DOWNLOADS_ID);
+            sURIMatcher.addURI(AUTHORITY, "all_downloads", ALL_DOWNLOADS);
+            sURIMatcher.addURI(AUTHORITY, "all_downloads/#", ALL_DOWNLOADS_ID);
+            sURIMatcher.addURI(AUTHORITY, "my_downloads/#/" + Downloads.Impl.RequestHeaders.URI_SEGMENT, REQUEST_HEADERS_URI);
+            sURIMatcher.addURI(AUTHORITY, "all_downloads/#/" + Downloads.Impl.RequestHeaders.URI_SEGMENT, REQUEST_HEADERS_URI);
+            // temporary, for backwards compatibility
```
blundell: lets delete this stuff? @amlcurran ?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadScanner.java
```diff
@@ -39,6 +39,7 @@
 
     private final Context mContext;
     private final MediaScannerConnection mConnection;
+    private final String mAuthority;
```
blundell: lets use Novoda style naming so we know what we've changed/added

library/src/main/java/com/novoda/downloadmanager/lib/DownloadProvider.java
amlcurran: sure, so long as those URIs aren't declared anywhere

demo/build.gradle
```diff
@@ -19,7 +19,7 @@ android {
         versionCode 1
         versionName "1.0"
 
-        manifestPlaceholders = [downloadAuthority: "${applicationId}"]
+        manifestPlaceholders = [downloadAuthority: "${applicationId}.downloads"]
```
blundell: this and https://github.com/novoda/download-manager/pull/13/files#diff-d18483ff140c15dd49ed3d88822df403R205 , 1 is the client and 1 is the library therefore the connection is not apparent. Here is lies a problem

library/src/main/java/com/novoda/downloadmanager/lib/DownloadProvider.java
blundell: yeah delete all that it uses I mean - simplify (as we do not have / need backwards compat)

demo/build.gradle
friedger: how the authority is determined can be more sophisticated. I think convention is still the best way to go.

blundell: We want the client to either completely control authority creation, or be completely agnostic of it (and we just use context). Not in the middle

blundell:
unmergable, but considered

PR #16, opened 2015-05-13
Reducing scope of DownloadManager API

amlcurran:
This PR reduces the scope of the DownloadManger API, as a lot of classes were publicly accessible which they probably shouldn't have been. This reduces their scope to package-scope. Furthermore, methods which were annotated with `@hide` have also had their scope reduced. 

Some code which is accessible only to system apps, or hidden because the original DownloadManager API wasn't finished (!) has been removed.

| Before | After |
| --- | --- |
| screen shot 2015-05-13 at 10 50 19 | screen shot 2015-05-13 at 11 41 14 |


blundell:
:+1:

PR #17, opened 2015-05-13
Scuba cleanup

amlcurran:
Removes references to "scuba" in the code. This was used as a unique identifier to avoid collisions with `android.app.DownloadManager` in MUBI.


blundell:
:+1:

PR #20, opened 2015-06-02
Remove Logging and add extensible public api

### LINKED ISSUE #19: Library still creates verbose logs even if logs are turned off
Won't actually print the logs, but creates the string for them. Results in logs of GC churn.

https://github.com/novoda/download-manager/blob/master/library/src/main/java/com/novoda/downloadmanager/lib/DownloadProvider.java#L866

blundell:
Toggles off the verbose logging that we should not be doing.
Logs are off by default for any new versions of the library.

This is done with a static flag (it is in a content provider so I cannot pass it the flag through construction). This implies you only have one instance of DownloadManager if you have two one with logging on and one with logging off the latest created will win.

Alternative would be to check the system property for verbose logging and then only log in those situtations? This still would mean 2 instances of the manager would favor the latest flag, so no gain...

Formatted some of the code so the diff is a bit bigger.

This fixes #19 which will give performance improvements for all clients.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
@@ -9,7 +9,7 @@ public class DownloadManagerBuilder {
     private ContentResolver contentResolver;
     private boolean verboseLogging;
 
-    public static DownloadManagerBuilder newInstance() {
+    public static DownloadManagerBuilder create() {
         return new DownloadManagerBuilder();
     }
 
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
@@ -6,8 +6,11 @@ import com.novoda.downloadmanager.lib.DownloadManager;
 
 public class DownloadManagerBuilder {
 
-    private ContentResolver contentResolver;
     private boolean verboseLogging;
+
+    DownloadManagerBuilder() {
+        // use the create method, Alex's favourite
+    }
 
     public static DownloadManagerBuilder create() {
         return new DownloadManagerBuilder();
@@ -18,14 +21,9 @@ public class DownloadManagerBuilder {
         return this;
     }
 
-    public DownloadManagerBuilder with(ContentResolver contentResolver) {
-        this.contentResolver = contentResolver;
-        return this;
-    }
-
-    public DownloadManager build() {
+    public DownloadManager build(ContentResolver contentResolver) {
         if (contentResolver == null) {
-            throw new IllegalStateException("You must use a ContentResolver with the DownloadManager. (use with(ContentResolver resolver);");
+            throw new IllegalStateException("You must use a ContentResolver with the DownloadManager.");
         }
         return new DownloadManager(contentResolver, verboseLogging);
     }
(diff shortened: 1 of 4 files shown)


library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
```diff
@@ -0,0 +1,33 @@
...
+    public static DownloadManagerBuilder create() {
+        return new DownloadManagerBuilder();
+    }
+
+    public DownloadManagerBuilder withVerboseLogging() {
+        this.verboseLogging = true;
+        return this;
+    }
+
+    public DownloadManagerBuilder with(ContentResolver contentResolver) {
```
ouchadam: different content resolvers can be used? otherwise the content resolver in the `build()`

blundell: yeah it could have been in the build, but having it this way adds flexibility and simplicity for the long term api

amlcurran: out of curiosity I always enforce things through the ctor, any reason for doing it through build instead?

blundell: I could do it through constructor, I'd argue it's easier for clients to use the `create()` method. (I'll make the constructor package now). Through the build method also makes sense because it enforces the use of a content resolver, but just for fluid api I believe for a client to do `.create()` then `.` and see the auto fill say `with...` , `with...`, `with...` `.build()`.    

I think through build is preferred because it enforces compile time correctness (in the standard case), but with a builder client facing, I think the ease of seeing the whole api (backed by runtime checks) wins for me.

@ouchadam vote
@amlcurran vote

and we'll go with the consensus

amlcurran: I disagree with clients using create but have no evidence to back it up, so I vote build!

blundell: and i'll still have to check null through build anyway. @ouchadam has the swinging vote!

ouchadam: for me `Builder.create` is the same as a `new Builder()` and personally I prefer builders that don't require any params to new them up, so **my vote is for build**!

amlcurran: Josh Bloch doesn't #justsayin http://www.drdobbs.com/jvm/creating-and-destroying-java-objects-par/208403883?pgno=2

blundell: TLDR! but I know it doesn't talk about public facing API's :-) hiding the constructor allows us to inject our builder without affecting clients (who will use create())

blundell: from a clients perspective looks so much better before :-( but DONE https://github.com/novoda/download-manager/commit/eab16482999ab6f00763b9485f35b6851369ea5e

amlcurran: I still personally think Joe Bloggs would be more used to seeing constructors than factory methods... just we here aren't because we hide them with `newInstance`. But don't get me started on `newInstance` either :tongue:

blundell: yes they are used to seeing constructors but we hide it for *our* benefit

PR #23, opened 2015-06-04
Configurable max concurrent downloads

Electryc:
Adds the possibility to configure the max concurrent downloads. As the download manager is a per app basis component, in order to specify the max number the client would have to include the following `meta-data` in its `AndroidManifest.xml` rather than relying on the `DownloadManagerBuilder`:

```xml
<meta-data
      android:name="com.novoda.downloadmanager.MaxConcurrentDownloads"
      android:value="1" />
```

If the `meta-data` tag is not present, the download manager will default to a maximum of 5 concurrent downloads maintaining the old behaviour. 

The specified value basically configures the `maximumPoolSize` of the `ThreadPoolExecutor`.  

_Note: The demo project now has 2 flavours - serial and parallel - in order to demo the usage of both cases_

### DIFF
--- a/demo/src/main/AndroidManifest.xml
+++ b/demo/src/main/AndroidManifest.xml
@@ -21,7 +21,7 @@
 
     <meta-data
       android:name="com.novoda.downloadmanager.MaxConcurrentDownloads"
-      android:value="1" />
+      android:value="@integer/max_concurrent_downloads" />
 
   </application>
 
--- a/demo/src/main/AndroidManifest.xml
+++ b/demo/src/main/AndroidManifest.xml
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest
-  package="com.novoda.downloadmanager.demo"
-  xmlns:android="http://schemas.android.com/apk/res/android">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+  package="com.novoda.downloadmanager.demo">
 
   <uses-permission android:name="android.permission.INTERNET" />
 
@@ -19,6 +18,10 @@
         <category android:name="android.intent.category.LAUNCHER" />
       </intent-filter>
     </activity>
+
+    <meta-data
+      android:name="com.novoda.downloadmanager.MaxConcurrentDownloads"
+      android:value="1" />
 
   </application>
 
--- a/demo/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
+++ b/demo/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
@@ -17,8 +17,8 @@ import java.util.List;
 
 public class MainActivity extends AppCompatActivity implements QueryForDownloadsAsyncTask.Callback {
     private static final String TAG = MainActivity.class.getSimpleName();
-    private static final String BIG_FILE = "http://downloads.bbc.co.uk/podcasts/radio4/fricomedy/fricomedy_20150501-1855a.mp3";
-    private static final String BBC_COMEDY_IMAGE = "http://ichef.bbci.co.uk/podcasts/artwork/266/fricomedy.jpg";
+    private static final String BIG_FILE = "http://open.live.bbc.co.uk/mediaselector/5/redir/version/2.0/mediaset/audio-nondrm-download/proto/http/vpid/p02ss0fm.mp3";
+    private static final String BBC_COMEDY_IMAGE = "http://ichef.bbci.co.uk/images/ic/640x360/p02ss0cf.jpg";
 
     private DownloadManager downloadManager;
     private ListView listView;
@@ -43,7 +43,7 @@ public class MainActivity extends AppCompatActivity implements QueryForDownloads
         request.setDestinationInInternalFilesDir(this, Environment.DIRECTORY_MOVIES, "podcast.mp3");
         request.setNotificationVisibility(Request.VISIBILITY_VISIBLE_NOTIFY_COMPLETED);
         request.setBigPictureUrl(BBC_COMEDY_IMAGE);
-        request.setTitle("BBC Friday Night Comedy");
+        request.setTitle("BBC Innuendo Bingo");
         request.setDescription("Nothing to do with beards.");
         request.setMimeType("audio/mp3");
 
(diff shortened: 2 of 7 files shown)


demo/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -17,8 +17,8 @@
 
 public class MainActivity extends AppCompatActivity implements QueryForDownloadsAsyncTask.Callback {
     private static final String TAG = MainActivity.class.getSimpleName();
-    private static final String BIG_FILE = "http://downloads.bbc.co.uk/podcasts/radio4/fricomedy/fricomedy_20150501-1855a.mp3";
-    private static final String BBC_COMEDY_IMAGE = "http://ichef.bbci.co.uk/podcasts/artwork/266/fricomedy.jpg";
+    private static final String BIG_FILE = "http://open.live.bbc.co.uk/mediaselector/5/redir/version/2.0/mediaset/audio-nondrm-download/proto/http/vpid/p02ss0fm.mp3";
```
amlcurran: hehehe

demo/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -43,7 +43,7 @@ private void setupDownloadingExample() {
         request.setDestinationInInternalFilesDir(this, Environment.DIRECTORY_MOVIES, "podcast.mp3");
         request.setNotificationVisibility(Request.VISIBILITY_VISIBLE_NOTIFY_COMPLETED);
         request.setBigPictureUrl(BBC_COMEDY_IMAGE);
-        request.setTitle("BBC Friday Night Comedy");
+        request.setTitle("BBC Innuendo Bingo");
```
ronocod: Top bantz

demo/src/main/AndroidManifest.xml
```diff
@@ -20,6 +19,10 @@
       </intent-filter>
     </activity>
 
+    <meta-data
+      android:name="com.novoda.downloadmanager.MaxConcurrentDownloads"
+      android:value="1" />
```
blundell: you can't make the demo just do a single download. Make two demo apps? What about all the people who wanna see multiple downloads  ... or set it with flavors, I dunno just need to demo both to the user :x:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadExecutorFactory.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.concurrent.TimeUnit;
+
+class DownloadExecutorFactory {
+    private static final int KEEP_ALIVE_TIME = 10;
+    private static final int DEFAULT_MAX_CONCURRENT_DOWNLOADS = 5;
+    private static final String METADATA_MAX_CONCURRENT_DOWNLOADS = "com.novoda.downloadmanager.MaxConcurrentDownloads";
+
+    public ExecutorService createExecutor(Context context) {
+        int maxConcurrentDownloads = getMaximumConcurrentDownloads(context);
+        ThreadPoolExecutor executor = new ThreadPoolExecutor(
```
amlcurran: I would use [Executors.newFixedThreadPool](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Executors.html#newFixedThreadPool(int)) instead, but I don't really know exactly what keepAliveTime and allowCoreThreadTimeOut do together

library/src/main/java/com/novoda/downloadmanager/lib/DownloadExecutorFactory.java
```diff
@@ -0,0 +1,48 @@
...
+                maxConcurrentDownloads,
+                KEEP_ALIVE_TIME,
+                TimeUnit.SECONDS,
+                new LinkedBlockingQueue<Runnable>());
+        executor.allowCoreThreadTimeOut(true);
+        return executor;
+    }
+
+    private int getMaximumConcurrentDownloads(Context context) {
+        String packageName = context.getApplicationContext().getPackageName();
```
amlcurran: :100: neat way of getting information

demo/src/main/AndroidManifest.xml
Electryc: yeah, we were thinking about how can we demo both behaviours, I think the only way is providing 2 apps.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadExecutorFactory.java
Electryc: Looking at the latest Lollipop update, it seems that they haven't used the `newFixedThreadPool` (http://androidxref.com/5.1.0_r1/xref/packages/providers/DownloadProvider/src/com/android/providers/downloads/DownloadService.java#120), but we can take a look to see if there's any difference between the two implementations.

Electryc: It seems like `Executors.newFixedThreadPool` has a keep alive time of 0.

amlcurran: yeah, as we discussed but for the benefit of everyone else keep-alive-time does nothing if you've got a fixed pool size (like we do currently)

Electryc: yeah but we use `executor.allowCoreThreadTimeOut(true)`. From the docs: 

> By default, the keep-alive policy applies only when there are more than corePoolSize threads. But method {@link #allowCoreThreadTimeOut(boolean)} can be used to apply this time-out policy to core threads as well, so long as the keepAliveTime value is non-zero.

demo/src/main/AndroidManifest.xml
Electryc: @blundell we've added 2 flavours of the demo project, one serial and one parallel. This way the user can select which flavour to run 8a29785

library/src/main/java/com/novoda/downloadmanager/lib/DownloadExecutorFactory.java
amlcurran: ah, good spot!

amlcurran:
:+1: lgtm

blundell:
but you didn't update the readme explaining the two flavors for the demo, how would anyone know :x:  I would also explain it in the app. Why they see single downloads or multiple downloads (personally I think two demo apps is more obvious for people).

ronocod:
We'll update the readme when we do the release now, hopefully the updated readme will be enough to explain the two flavours to users.

PR #25, opened 2015-06-04
Separating serial and parallel demo apps

Electryc:
This creates two demo apps in order to show the usage of downloading in serial and in parallel. The modules are named `demo_serial` and `demo_parallel` each of them with their specific application name in order to easily distinguish them in the launcher. 

Practically they are the same app, but in order the user to easily see the difference between serial and parallel downloads we've created two separate apps rather than two flavours as discussed https://github.com/novoda/download-manager/pull/23#issuecomment-108933797 

We've also updated the Readmes for each of the apps. 

Launcher | Serial | Parallel
------------ | -------- | ----------
launcher_dwm | serial_dwm | parallel_dwm


Electryc:
@blundell as you said, better going with 2 apps rather than 2 flavours in order to gain more visibility of the different configurations.

demo_parallel/build.gradle
```diff
@@ -13,7 +13,7 @@ android {
     buildToolsVersion "22.0.1"
 
     defaultConfig {
-        applicationId "com.novoda.downloadmanager.demo"
+        applicationId "com.novoda.downloadmanager.demo_parallel"
```
blundell: package convention is dotted why not `com.novoda.downloadmanager.demo.parallel`

demo_serial/README.md
```diff
@@ -7,5 +7,5 @@ This is a copy of the http://developer.android.com/reference/android/app/Downloa
 
 Here we show two use cases
 
-   - Downloading from a url
+   - Downloading items **serially** from a url
```
blundell: `- Downloading items **serially** from a url (this means the second will not start till the first completes)`

demo_serial/build.gradle
```diff
@@ -0,0 +1,33 @@
...
+    }
+}
+apply plugin: 'com.android.application'
+
+android {
+    compileSdkVersion 22
+    buildToolsVersion "22.0.1"
+
+    defaultConfig {
+        applicationId "com.novoda.downloadmanager.demo_serial"
```
blundell: see https://github.com/novoda/download-manager/pull/25/files#r31792732

demo_serial/proguard-rules.pro
```diff
@@ -0,0 +1,17 @@
+# Add project specific ProGuard rules here.
+# By default, the flags in this file are appended to flags specified
+# in /Users/Blundell/Developer/android-sdk-macosx/tools/proguard/proguard-android.txt
```
blundell: whoops, delete this file, no need for proguard rules in a demo project

demo_serial/src/main/java/com/novoda/downloadmanager/demo/QueryForDownloadsAsyncTask.java
```diff
@@ -0,0 +1,57 @@
+package com.novoda.downloadmanager.demo;
```
blundell: are these shared? update them for consistency

blundell:
we need to record somewhere to ask for a design task for some icons :) :+1:

demo_serial/src/main/java/com/novoda/downloadmanager/demo/QueryForDownloadsAsyncTask.java
Electryc: you mean `com.novoda.downloadmanager.demo.serial` ?

blundell: yes

blundell: or do both demos have the same java package? (and it's onlt the application ID you change?) because that is fine as well

demo_parallel/build.gradle
Electryc: :+1: 0cae060

demo_serial/README.md
Electryc: :+1: a008441

demo_serial/src/main/java/com/novoda/downloadmanager/demo/QueryForDownloadsAsyncTask.java
Electryc: well they don't have anything shared. In order to avoid any confusion, I've just duplicated the old demo project. I've updated the package names here 0cae060

demo_serial/proguard-rules.pro
Electryc: :dancers:

blundell:
now the two demo's can be changed and go separate paths of their own :+1:  one could be beards one moustaches ;-)

PR #26, opened 2015-06-05
Add tests around configuring concurrent downloads limit

ronocod:
* Tests creation of the `ThreadPoolExecutor` to make sure that the thread pool is the correct size as this controls the number of concurrent downloads allowed.
* Updates build tools and Gradle versions.
* Adds JUnit, Mockito and Fest for testing.

screen shot 2015-06-05 at 13 58 31

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadExecutorFactory.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadExecutorFactory.java
@@ -1,24 +1,27 @@
 package com.novoda.downloadmanager.lib;
 
 import android.content.Context;
-import android.content.pm.ApplicationInfo;
-import android.content.pm.PackageManager;
-import android.os.Bundle;
 
-import com.novoda.notils.logger.simple.Log;
-
-import java.util.concurrent.ExecutorService;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 
 class DownloadExecutorFactory {
-    private static final int KEEP_ALIVE_TIME = 10;
-    private static final int DEFAULT_MAX_CONCURRENT_DOWNLOADS = 5;
-    private static final String METADATA_MAX_CONCURRENT_DOWNLOADS = "com.novoda.downloadmanager.MaxConcurrentDownloads";
 
-    public ExecutorService createExecutor(Context context) {
-        int maxConcurrentDownloads = getMaximumConcurrentDownloads(context);
+    private static final int KEEP_ALIVE_TIME = 10;
+
+    private final ConcurrentDownloadsLimitProvider concurrentDownloadsLimitProvider;
+
+    static DownloadExecutorFactory newInstance(Context context) {
+        return new DownloadExecutorFactory(ConcurrentDownloadsLimitProvider.newInstance(context));
+    }
+
+    DownloadExecutorFactory(ConcurrentDownloadsLimitProvider concurrentDownloadsLimitProvider) {
+        this.concurrentDownloadsLimitProvider = concurrentDownloadsLimitProvider;
+    }
+
+    public ThreadPoolExecutor createExecutor() {
+        int maxConcurrentDownloads = concurrentDownloadsLimitProvider.getConcurrentDownloadsLimit();
         ThreadPoolExecutor executor = new ThreadPoolExecutor(
                 maxConcurrentDownloads,
                 maxConcurrentDownloads,
@@ -29,20 +32,4 @@ class DownloadExecutorFactory {
         return executor;
     }
 
-    private int getMaximumConcurrentDownloads(Context context) {
-        String packageName = context.getApplicationContext().getPackageName();
-        PackageManager packageManager = context.getPackageManager();
-        try {
-            ApplicationInfo applicationInfo = packageManager.getApplicationInfo(packageName, PackageManager.GET_META_DATA);
-            Bundle bundle = applicationInfo.metaData;
-            if (bundle == null) {
-                return DEFAULT_MAX_CONCURRENT_DOWNLOADS;
-            }
-
-            return bundle.getInt(METADATA_MAX_CONCURRENT_DOWNLOADS, DEFAULT_MAX_CONCURRENT_DOWNLOADS);
-        } catch (PackageManager.NameNotFoundException e) {
-            Log.e("Application info not found for: " + packageName + " " + e.getMessage());
-        }
-        return DEFAULT_MAX_CONCURRENT_DOWNLOADS;
-    }
 }
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadExecutorFactory.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadExecutorFactory.java
@@ -1,7 +1,5 @@
 package com.novoda.downloadmanager.lib;
 
-import android.content.Context;
-
 import java.util.concurrent.LinkedBlockingQueue;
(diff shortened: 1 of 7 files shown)


library/src/main/java/com/novoda/downloadmanager/lib/DownloadExecutorFactory.java
```diff
@@ -1,24 +1,27 @@
...
+    static DownloadExecutorFactory newInstance(Context context) {
+        return new DownloadExecutorFactory(ConcurrentDownloadsLimitProvider.newInstance(context));
+    }
+
+    DownloadExecutorFactory(ConcurrentDownloadsLimitProvider concurrentDownloadsLimitProvider) {
+        this.concurrentDownloadsLimitProvider = concurrentDownloadsLimitProvider;
+    }
+
+    public ThreadPoolExecutor createExecutor() {
+        int maxConcurrentDownloads = concurrentDownloadsLimitProvider.getConcurrentDownloadsLimit();
```
xrigau: this number can't change at runtime can it? so why not passing this number directly in the constructor instead?

xrigau:
:+1: this is cool

library/src/main/java/com/novoda/downloadmanager/lib/DownloadExecutorFactory.java
ronocod: That's the idea behind the factory, it hides away the retrieval of that value so that the user of the executor doesn't have to know how to get the value.

library/src/test/java/com/novoda/downloadmanager/lib/DownloadExecutorFactoryTest.java
```diff
@@ -0,0 +1,39 @@
...
+    DownloadExecutorFactory factory;
+
+    @Before
+    public void setUp() throws Exception {
+        initMocks(this);
+        factory = new DownloadExecutorFactory(metadataReader);
+    }
+
+    @Test
+    public void givenALimitProviderWhenTheExecutorIsCreatedThenTheCorrectLimitIsUsed() throws Exception {
```
amlcurran: :+1: nice one

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
```diff
@@ -149,7 +149,7 @@ public void onCreate() {
                 Downloads.Impl.ALL_DOWNLOADS_CONTENT_URI,
                 true, mObserver);
 
-        mExecutor = new DownloadExecutorFactory().createExecutor(this);
+        mExecutor = DownloadExecutorFactory.newInstance(this).createExecutor();
```
blundell: just inline this newInstance method, can't see it adding anything (lets not refactor prematurely)

ronocod: :+1: https://github.com/novoda/download-manager/commit/9eda679105b70b70390ebbdc1ed4bc852bec95aa

library/src/main/java/com/novoda/downloadmanager/lib/DownloadExecutorFactory.java
xrigau: do it in a `newInstance` method waat

ronocod: Done in https://github.com/novoda/download-manager/commit/f5b24f1fb27db739e5b5e84d54d64c9bbe16384f

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
ronocod: Went back to a different newInstance method: https://github.com/novoda/download-manager/commit/f5b24f1fb27db739e5b5e84d54d64c9bbe16384f

library/src/main/java/com/novoda/downloadmanager/lib/DownloadExecutorFactory.java
```diff
@@ -1,24 +1,28 @@
...
-    private static final int DEFAULT_MAX_CONCURRENT_DOWNLOADS = 5;
-    private static final String METADATA_MAX_CONCURRENT_DOWNLOADS = "com.novoda.downloadmanager.MaxConcurrentDownloads";
 
-    public ExecutorService createExecutor(Context context) {
-        int maxConcurrentDownloads = getMaximumConcurrentDownloads(context);
+    private final int maxConcurrentDownloads;
+
+    static DownloadExecutorFactory newInstance(Context context) {
+        ConcurrentDownloadsLimitProvider concurrentDownloadsLimitProvider = ConcurrentDownloadsLimitProvider.newInstance(context);
+        int maxConcurrentDownloads = concurrentDownloadsLimitProvider.getConcurrentDownloadsLimit();
```
blundell: why did @xrigau say do this? :-) this is "doing work in construction" sometimes it's ok because it is enforced by the Android system. However this here is hidden in the `newInstance` but you should just inline this newInstance - we should not mix construction with behaviour

xrigau: but `DownloadExecutorFactory` only cares about the number of concurrent downloads not how to get that number

xrigau: as in the object itself

xrigau: I just thought there's no point on asking the `ConcurrentDownloadsLimitProvider` every time since that number won't change at runtime

blundell: I'm just saying inline this, so it'll still do the same number of method calls.

ronocod: I don't really understand this, how is this different from any of our other `newInstance()` methods?

The only issue I could see is that if we change `ConcurrentDownloadsLimitProvider` to an implementation where the limit can change at runtime then this might cause problems.

ronocod: That question was to @blundell by the way

xrigau: I'm confused as well

blundell: i just think a `newInstance` used once is mostly pointless. Espeically when it does more than construction.

@xrigau why do you want the newInstance I cannot see the benefit shall we tri-pair discuss?

ronocod: I'm happy to do either one, I think the `newInstance()` is a bit cleaner but it's not super-important to me

blundell: i feel the "cleanliness" needs to look horrible so that next person can see what we are doing, hiding the unnecceessary abstraction (and keeping more new's in one place)

ronocod: Can we agree on this setup then? https://github.com/novoda/download-manager/commit/70f9cfa06409041dcac52d5e660e562b75619ab2

I think it's better that we don't assume that the limit will always be the same at runtime, I think it's an implementation detail

xrigau: ok I'm okay with reverting to how it looked before as discussed with @blundell since the method is called the same number of times (silly me for not reading all the code :tongue: )

PR #27, opened 2015-06-05
Cache requested bitmap for notifications

ronocod:
Currently the image for a download notification is fetched from the network every time the notification is updated. This causes memory usage to grow rapidly which causes GC to happen more often. It also wastes network bandwidth.

The retriever will now cache the latest Bitmap and return it without a network request when the same URL is requested.

## Memory Allocations

Before | After
---|---
screen shot 2015-06-05 at 14 59 05|screen shot 2015-06-05 at 14 59 13

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/OkHttpNotificationImageRetriever.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/OkHttpNotificationImageRetriever.java
@@ -14,21 +14,34 @@ class OkHttpNotificationImageRetriever implements NotificationImageRetriever {
 
     private final OkHttpClient client;
 
+    private String imageUrl;
+    private Bitmap bitmap;
+
     public OkHttpNotificationImageRetriever() {
         client = new OkHttpClient();
     }
 
     @Override
     public Bitmap retrieveImage(String imageUrl) {
+        if (imageUrl.equals(this.imageUrl)) {
+            return bitmap;
+        }
+
+        this.imageUrl = imageUrl;
+        if (bitmap != null && !bitmap.isRecycled()) {
+            bitmap.recycle();
+            bitmap = null;
+        }
         Request request = new Request.Builder()
                 .get()
-                .url(imageUrl)
+                .url(this.imageUrl)
                 .build();
         try {
             Response response = client.newCall(request).execute();
             InputStream inputStream = response.body().byteStream();
             try {
-                return BitmapFactory.decodeStream(inputStream);
+                bitmap = BitmapFactory.decodeStream(inputStream);
+                return bitmap;
             } finally {
                 inputStream.close();
             }
--- a/library/src/main/java/com/novoda/downloadmanager/lib/OkHttpNotificationImageRetriever.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/OkHttpNotificationImageRetriever.java
@@ -38,8 +38,8 @@ class OkHttpNotificationImageRetriever implements NotificationImageRetriever {
             Response response = client.newCall(request).execute();
             InputStream inputStream = response.body().byteStream();
             try {
-                this.imageUrl = imageUrl;
                 bitmap = BitmapFactory.decodeStream(inputStream);
+                this.imageUrl = imageUrl;
                 return bitmap;
             } finally {
                 inputStream.close();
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/novoda/downloadmanager/lib/OkHttpNotificationImageRetriever.java
```diff
@@ -28,7 +38,10 @@ public Bitmap retrieveImage(String imageUrl) {
             Response response = client.newCall(request).execute();
             InputStream inputStream = response.body().byteStream();
             try {
-                return BitmapFactory.decodeStream(inputStream);
+                this.imageUrl = imageUrl;
```
xrigau: how come this is done so late?

ronocod: So that we only set the new image URL on successful requests. If we set it at the start then the cached bitmap could be returned even after a failed request

library/src/main/java/com/novoda/downloadmanager/lib/OkHttpNotificationImageRetriever.java
```diff
@@ -36,4 +49,14 @@ public Bitmap retrieveImage(String imageUrl) {
...
     }
+
+    private void cleanupOldBitmap() {
+        if (bitmap == null) {
+            return;
+        }
+        if (!bitmap.isRecycled()) {
+            bitmap.recycle();
+        }
+        bitmap = null;
```
xrigau: this is unneeded since you're reassigning the reference to a new bitmap anyway

amlcurran:
Before and after memory profiles?? :angel:

library/src/main/java/com/novoda/downloadmanager/lib/OkHttpNotificationImageRetriever.java
```diff
@@ -36,4 +49,14 @@ public Bitmap retrieveImage(String imageUrl) {
             return null;
         }
     }
+
+    private void cleanupOldBitmap() {
+        if (bitmap == null) {
+            return;
+        }
+        if (!bitmap.isRecycled()) {
+            bitmap.recycle();
```
xrigau: as much as I like this, the [docs say](https://developer.android.com/reference/android/graphics/Bitmap.html#recycle())

> This is an advanced call, and normally need not be called, since the normal GC process will free up this memory when there are no more references to this bitmap.

blundell:
ah, we did have this in mubi but it was done at the imageloader level.

library/src/main/java/com/novoda/downloadmanager/lib/OkHttpNotificationImageRetriever.java
xrigau: plus the problem is you return that bitmap in `retrieveImage` which means someone else could be using that bitmap so if you recycle it here and that bitmap is still in use an exception will be thrown

library/src/main/java/com/novoda/downloadmanager/lib/OkHttpNotificationImageRetriever.java
xrigau: could still be done between L30 and L40

library/src/main/java/com/novoda/downloadmanager/lib/OkHttpNotificationImageRetriever.java
ronocod: So we can take out the whole `cleanupOldBitmap()` method then?

xrigau: yes - let the client deal with it (i.e. when the client asks for a new image then they can recycle the old bitmap if they really want)

library/src/main/java/com/novoda/downloadmanager/lib/OkHttpNotificationImageRetriever.java
ronocod: Moved it to after the bitmap is decoded. The earlier we do it, the more chance of caching an invalid bitmap.

xrigau: yeah it is more explicit now IMO

amlcurran:
:100: points for outlining the improvements in a graph!

Out of interest, why does it still do more than one?

ronocod:
Each update to the notification clones the Bitmap, can't find a way to stop it :(

amlcurran:
@ronocod nice one. I'm happy :+1:

ronocod:
:ship:?

PR #28, opened 2015-06-09
Enforces passing a Context via the DM Builder

xrigau:
Enforces passing a Context via the DM Builder

This means we don't need to pass in a Context to create a Request
 
Note this is a BREAKING CHANGE! since we've changed the API but it is an improvement since it didn't make sense having to pass in the Context for each Request

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
@@ -1,19 +1,22 @@
 package com.novoda.downloadmanager;
 
 import android.content.ContentResolver;
+import android.content.Context;
 
 import com.novoda.downloadmanager.lib.DownloadManager;
 
 public class DownloadManagerBuilder {
 
+    private final Context context;
+
     private boolean verboseLogging;
 
-    DownloadManagerBuilder() {
-        // use the create method, Alex's favourite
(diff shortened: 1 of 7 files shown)


library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
```diff
@@ -1,19 +1,22 @@
...
 import com.novoda.downloadmanager.lib.DownloadManager;
 
 public class DownloadManagerBuilder {
 
+    private final Context context;
+
     private boolean verboseLogging;
 
-    DownloadManagerBuilder() {
-        // use the create method, Alex's favourite
```
amlcurran: *sighs wistfully*

library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
```diff
@@ -1,19 +1,22 @@
...
+    DownloadManagerBuilder(Context context) {
+        this.context = context;
     }
 
-    public static DownloadManagerBuilder create() {
-        return new DownloadManagerBuilder();
+    public static DownloadManagerBuilder from(Context context) {
+        return new DownloadManagerBuilder(context.getApplicationContext());
```
amlcurran: :+1: good enforce

Issue #22, opened 2015-06-04
Pull the latest changes from AOSP


xrigau:
It'd be good to pull changes/improvements/bugfixes (if any) from the AOSP from Lollipop or even Android M

ronocod:
2 issues we've found with this:

* There's already been so many changes from the AOSP code that merging in newer versions will be very time-consuming.
* AOSP has moved to using JobScheduler for downloads cleanup since last June, so we'll have to remove any AOSP code that's not compatible with API 14.

ronocod:
Also if we move away from keeping the code AOSP-based then we can start refactoring a lot more as we wouldn't care about staying close to the original code.

blundell:
yeah we don't want to pull changes. Only any perf improvements or bug fixes that we notice in the aosp issue list.

PR #29, opened 2015-06-18
Injects a client is ready to proceed pre-check

zegnus:
### Problem description

The current DownloadManager has a set of internal rules that specifies if a download can proceed or not, but there is no any mechanism in place to enhance the check following a set of rules defined by a client.

As for example, don't download if the client is not logged in.

### Solution implemented

A client can now specify if the downloads are ready to proceed or not by implementing the interface `DownloadClientReadyChecker` in the Application Activity.

As for example:

```java
public class DemoParallelApplication extends Application implements DownloadClientReadyChecker {
    @Override
    public boolean isReadyToDownload() {
        // rule 1
        // rule 2
        // return rule1IsTrue AND rule2IsTrue
    }
}
```

For now, `isReadyToDownload` don't take any parameters. If that is needed another Pull Request will be issued. Meaning that this method is a global status query, not for a particular download.


library/build.gradle
```diff
@@ -37,6 +37,10 @@ android {
     lintOptions {
         abortOnError false
     }
+
+    testOptions {
+        unitTests.returnDefaultValues = true
```
zegnus: Otherwise the test that executes

```
values.put(Downloads.Impl.COLUMN_STATUS, mStatus);
```

will fail

library/src/main/java/com/novoda/downloadmanager/lib/DownloadClientReadyChecker.java
```diff
@@ -0,0 +1,18 @@
...
+
+public interface DownloadClientReadyChecker {
+    Ready READY = new Ready();
+
+    boolean isReadyToDownload();
+
+    /**
+     * Ready specifies that it is always ready to download
+     */
+    class Ready implements DownloadClientReadyChecker {
```
amlcurran: nice

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
```diff
@@ -40,10 +40,27 @@ public Reader(ContentResolver resolver, Cursor cursor) {
...
-        public DownloadInfo newDownloadInfo(Context context, SystemFacade systemFacade, StorageManager storageManager, DownloadNotifier notifier) {
-            final DownloadInfo info = new DownloadInfo(context, systemFacade, storageManager, notifier);
+        public DownloadInfo newDownloadInfo(
+                Context context,
+                SystemFacade systemFacade,
+                StorageManager storageManager,
+                DownloadNotifier notifier,
+                DownloadClientReadyChecker downloadClientReadyChecker) {
+
+            Fuzz fuzz = new Fuzz();
```
amlcurran: ??

amlcurran: as in, what is this ha

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
```diff
@@ -154,6 +157,13 @@ public void onCreate() {
         mExecutor = factory.createExecutor();
     }
 
+    private DownloadClientReadyChecker getDownloadClientReadyChecker() {
+        if (!(getApplication() instanceof DownloadClientReadyChecker)) {
```
amlcurran: Can't you pass this into the constructor of the DownloadManager? Having a hard dependency on the Application class isn't very extensible

library/build.gradle
blundell: :x: this option should never be turned on.  It means you're testing things not under our control.

Can discuss later :x:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
```diff
@@ -217,7 +239,10 @@ private Long getLong(String column) {
     public int mBypassRecommendedSizeLimit;
     public String bigPictureResourceUrl;
 
-    public int mFuzz;
+    public Uri mAllDownloadsUri;
```
blundell: if you're creating new fields, use Novoda coding standards - we are moving away from the Google code and this helps highlight what we have added or changed.

xrigau:
The PR description doesn't explain what happens in case that method returns `false` - does it retry later? does this happen for every download? maybe this is like a `GlobalDownloadClientReadyChecker` and then if you wanted to do it for each download then it could be a `DownloadClientReadyChecker` ?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
```diff
@@ -439,14 +473,14 @@ private NetworkState checkSizeAllowedForNetwork(int networkType) {
...
-            final boolean isReady = isReadyToDownload();
+            final boolean isReady = isDownloadManagerReadyToDownload() && isClientIsReadyToDownload();
             final boolean isActive = mSubmittedTask != null && !mSubmittedTask.isDone();
             if (isReady && !isActive) {
                 if (mStatus != Downloads.Impl.STATUS_RUNNING) {
                     mStatus = Downloads.Impl.STATUS_RUNNING;
                     ContentValues values = new ContentValues();
                     values.put(Downloads.Impl.COLUMN_STATUS, mStatus);
-                    mContext.getContentResolver().update(getAllDownloadsUri(), values, null, null);
+                    mContext.getContentResolver().update(mAllDownloadsUri, values, null, null);
```
blundell: why this change?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
zegnus: Yeah I had the same question.

If you look down the code at https://github.com/novoda/download-manager/pull/29/files#diff-68db38a7048316539cb0931793be1808L242 you will see that is a random value used in order to avoid clashing for a restart time queries.

It is needed to be injected because otherwise when the test calls https://github.com/novoda/download-manager/blob/master/library/src/main/java/com/novoda/downloadmanager/lib/Helpers.java#L39, the code fails

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
```diff
@@ -480,11 +518,12 @@ public boolean isOnCache() {
     }
 
     public Uri getMyDownloadsUri() {
-        return ContentUris.withAppendedId(Downloads.Impl.CONTENT_URI, mId);
+        return mAllDownloadsUri;
```
blundell: why this change?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
```diff
@@ -115,6 +132,11 @@ private void addHeader(DownloadInfo info, String header, String value) {
             info.mRequestHeaders.add(Pair.create(header, value));
         }
 
+        private void setDownloadUris(DownloadInfo info) {
+            info.mAllDownloadsUri = ContentUris.withAppendedId(Downloads.Impl.ALL_DOWNLOADS_CONTENT_URI, info.mId);
```
amlcurran: why do we need to do this out of curiosity/what's changed?

library/src/test/java/com/novoda/downloadmanager/lib/DownloadInfoTest.java
```diff
@@ -0,0 +1,89 @@
...
+    public void setUp() {
+        initMocks(this);
+
+        when(mockFuzz.getFuzz()).thenReturn(1);
+        when(mockContext.getContentResolver()).thenReturn(mockContentResolver);
+        when(mockContentResolver.update(any(Uri.class), any(ContentValues.class), any(String.class), any(String[].class))).thenReturn(0);
+    }
+
+    @Test
+    public void givenClientIsReadyToDownloadThenStartDownload() {
```
amlcurran: :+1: nice tests

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
amlcurran: OK, perhaps rename or javadoc to make it more obvious?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
ronocod: :+1: for not forcing it on the Application

library/build.gradle
zegnus: Yup I know, but I didn't want to change too much the code.

The only way is to inject that ContentValues then... or a better idea?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
zegnus: :+1:

ronocod: Now might be a good time to rename it to something like `RandomNumberGenerator`?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
ronocod: But it is very difficult to pass dependencies into the `DownloadService` :disappointed:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
ronocod: Should this be `mMyDownloadsUri`?

zegnus:
@xrigau What happens is that the download don't proceed. As no any logic has been added or modified regarding what to do when a download cannot proceed, I haven't documented it.

There is already an internal *can the download proceed* mechanism, the logic of this new check is added on top of the previous check, and the result is the same.

As a reference, the `activeDownload` will be false here https://github.com/novoda/download-manager/blob/master/library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java#L344 from https://github.com/novoda/download-manager/blob/master/library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java#L442 where `isReadyToDownload` becoming false

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
zegnus: Can use the method, reverting

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
zegnus: Access to `ContentUris` will fail at test. 

As well there is no need to calculate the value every time as it can be precomputed

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
zegnus: https://github.com/novoda/download-manager/pull/29#discussion_r32744701

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
zegnus: Yes I thought of that but because that comes from Google I wanted to maintain the consistency of the name just in case we have to merge back again from them, so we know what is what.

Can rename if we don't really care about merging again in the future?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
zegnus: Yes exactly, ideas welcome :)

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
amlcurran: Think we said we weren't going to take any more of their code ( #22 )

zegnus: :+1:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadClientReadyChecker.java
Electryc: :+1:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
```diff
@@ -456,6 +495,10 @@ public boolean startDownloadIfReady(ExecutorService executor) {
         }
     }
 
+    private boolean isClientIsReadyToDownload() {
```
Electryc: This reads a little bit odd. Shouldn't it be `isClientReadyToDownload`?

library/src/main/java/com/novoda/downloadmanager/lib/RandomNumberGenerator.java
```diff
@@ -0,0 +1,11 @@
+package com.novoda.downloadmanager.lib;
+
+/**
+ * This class generates a random value to be used as a randomized restart time
+ * for the server queries
+ */
+class RandomNumberGenerator {
+    int getFuzz() {
```
Electryc: `generate()` maybe?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
zegnus: yes!

demo_parallel/src/main/java/com/novoda/downloadmanager/demo/parallel/DemoParallelApplication.java
```diff
@@ -0,0 +1,12 @@
+package com.novoda.downloadmanager.demo.parallel;
+
+import android.app.Application;
+
+import com.novoda.downloadmanager.lib.DownloadClientReadyChecker;
+
+public class DemoParallelApplication extends Application implements DownloadClientReadyChecker {
```
blundell: In the readme we explicitly wrote that the example for custom rules was in the serial demo not the parralel. No need to add this

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
```diff
@@ -40,10 +40,29 @@ public Reader(ContentResolver resolver, Cursor cursor) {
...
+                Context context,
+                SystemFacade systemFacade,
+                StorageManager storageManager,
+                DownloadNotifier notifier,
+                DownloadClientReadyChecker downloadClientReadyChecker) {
+
+            RandomNumberGenerator randomNumberGenerator = new RandomNumberGenerator();
+            ContentValues contentValues = new ContentValues();
+
+            final DownloadInfo info = new DownloadInfo(
```
blundell: no need for final

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
blundell: yeah @amlcurran we did look. We don't want to do it reflectively and this service can be started from a broadcast receiver when the only thing that is available is the Application.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
```diff
@@ -115,6 +134,11 @@ private void addHeader(DownloadInfo info, String header, String value) {
             info.mRequestHeaders.add(Pair.create(header, value));
         }
 
+        private void setDownloadUris(DownloadInfo info) {
+            info.allDownloadsUri = ContentUris.withAppendedId(Downloads.Impl.ALL_DOWNLOADS_CONTENT_URI, info.mId);
```
blundell: :x: this change breaks the current behaviour.

Let's try and do minimal changes to the DownloadManager or be extra careful when doing them.

This breaks the behaviour as now you "pre-compute the allDownloadsUri" however if the `info.mId` changes which it can! see method `updateFromDatabase()`

blundell:
LGTM :+1: :shipit:

blundell:
ah yes you're right, my local build broke but I ran out of laptop battery, I'm on it :+1:

blundell:
@zegnus the problem was you where testing something that wasn't isolated, the downloadmanager code is not architectured with tests in mind and sometimes to change this is close to breaking behaviour or introducing bugs that we should not be doing, it's a tight rope to walk and we have to be careful. Minimal changes recommended at all times.

stefanhoth:
`DownloadClientReadyChecker.isReadyToDownload` dat naming :eyes:

zegnus:
@blundell Yup, I had the same feeling. I wanted to start to write tests on this module because there are almost none but as you said this is not encoded in a way that helps testing, at all and I wanted to the the minimal changes for the tests.

I thought about the same as you did here https://github.com/novoda/download-manager/commit/f3b4b6aad0ee816a83c0fff067c33fb05c9a17c7 didn't want to change much. But considering how the dependencies of the download method mixes up different concepts and makes the tests a bit of a hell, this is much much better :+1:

xrigau:
@zegnus :+1:

xrigau:
no outstanding issues so I say :shipit: and iterate :+1:

PR #30, opened 2015-06-19
Adds documentation for the new DownloadClientReadyChecker functionality

zegnus:
## Description of the task

We have a new feature that allows a client to tell the DownloadManager if a download can proceed or not.

This PR adds that information to the public DownloadManagerBuilder

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
@@ -5,6 +5,11 @@ import android.content.Context;
 
 import com.novoda.downloadmanager.lib.DownloadManager;
 
+/**
+ * A client can specify whether the downloads are allowed to proceed by implementing
+ * {@link com.novoda.downloadmanager.lib.DownloadClientReadyChecker} on your Activity class
+ *
+ */
 public class DownloadManagerBuilder {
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
```diff
@@ -5,6 +5,11 @@
 
 import com.novoda.downloadmanager.lib.DownloadManager;
 
+/**
+ * A client can specify whether the downloads are allowed to proceed by implementing
+ * {@link com.novoda.downloadmanager.lib.DownloadClientReadyChecker} on your Activity class
```
blundell: `n your Application class`

amlcurran: can we stick this in the readme too?

blundell: maybe, i just think we should be careful to not clutter the readme (even the parralel thing prob shouldn't be in the readme)

amlcurran: wiki?

stefanhoth:
Sorry, had to be done.

PR #32, opened 2015-06-19
Adds javadoc for documenting the method isAllowedToDownload

zegnus:
## Description of the task

It is of relevance for the client to be aware of that the method `isAllowedToDownload` will be called from a background thread.

This task adds a Javadoc in order to document this fact to the public API


library/src/main/java/com/novoda/downloadmanager/lib/DownloadClientReadyChecker.java
```diff
@@ -3,6 +3,9 @@
 public interface DownloadClientReadyChecker {
     Ready READY = new Ready();
 
+    /**
+     * Warning: This method is executed on a background thread
```
blundell: no need to say "warning" it's just executed on a background thread, people will probably we happy rather than warned :)

Electryc:
EZ :+1:

PR #33, opened 2015-06-19
31 bug/serialised downloads dont show as queued

ouchadam:
Downloads were being set to `RUNNING` before they were actually ran by the Executor service, this PR now applies the RUNNING status in the `runnable run()` instead of when the task is submitted.

Submitted tasks are set to `PENDING` no gif since the demo project isn't loading for me (or @Electryc even on master) but works fine in our external project, gotta have faith!

**SIDEEFFECT -** 
this causes the serialised notification to no longer be stacked, no affect on parallel (this behaviour matches play music)

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
@@ -476,11 +476,19 @@ class DownloadInfo {
         synchronized (this) {
             final boolean isActive = mSubmittedTask != null && !mSubmittedTask.isDone();
             if (!isActive) {
+                updateStatus(DownloadManager.STATUS_PENDING);
                 mTask = new DownloadThread(mContext, mSystemFacade, this, mStorageManager, mNotifier);
                 mSubmittedTask = executor.submit(mTask);
             }
             return isActive;
         }
+    }
+
+    public void updateStatus(int status) {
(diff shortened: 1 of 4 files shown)


ouchadam:
fixes #31

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
```diff
@@ -486,6 +480,13 @@ public boolean startDownloadIfNotActive(ExecutorService executor) {
         }
     }
 
+    public void updateStatus(int status) {
+        mStatus = status;
+        downloadStatusContentValues.clear();
+        downloadStatusContentValues.put(Downloads.Impl.COLUMN_STATUS, mStatus);
+        mContext.getContentResolver().update(getAllDownloadsUri(), downloadStatusContentValues, null, null);
```
blundell: so it puts PENDING into the DB just before the thread starts. When does RUNNING go into the DB?

ouchadam: here https://github.com/novoda/download-manager/pull/33/files#diff-9b10a080ea5fa55b1f8247c09b432b07R168

blundell: ah thats wwhat that means LGTM then :

PR #36, opened 2015-06-23
Improve API for enqueuing downloads and fix notification UI

ronocod:
* Ensure that the existing `enqueue(Request request)` method works correctly with batching
* Add new `enqueue(RequestBatch batch)` method which allows a batch of requests to be added and represented as one download in the UI:

```java
RequestBatch batch = RequestBatch.newInstance("Premium Video", "A video with an advert", VIDEO_IMAGE_URL);

Request videoRequest = new Request(videoUri);
videoRequest.setExtra("video");
batch.addRequest(videoRequest);

Request advertRequest = new Request(advertUri);
advertRequest.setExtra("advert");
batch.addRequest(advertRequest);

long batchId = downloadManager.enqueue(batch);
```

* Notifications now use a stacking style when multiple batches are being downloaded. Previously a stacking style was used when multiple files were being downloaded, even if they were part of the same batch. Example (1 batch is downloading, it contains 2 requests):
download


blundell:
:100:

demo_serial/src/main/java/com/novoda/downloadmanager/demo/serial/MainActivity.java
```diff
@@ -41,21 +41,22 @@ protected void onCreate(Bundle savedInstanceState) {
 
     private void setupDownloadingExample() {
         Uri uri = Uri.parse(BIG_FILE);
-        final DownloadBatch batch = new DownloadBatch("Large Beard Shipment", "Goatees galore", BEARD_IMAGE);
+        final RequestBatch batch = RequestBatch.newInstance("Large Beard Shipment", "Goatees galore", BEARD_IMAGE);
```
blundell: so you should probably create a `BatchExampleActivity` and reset this demo back to the standard case of a download.  (or make this a menu to other activities)

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
```diff
@@ -702,6 +697,28 @@ static String getWhereClauseForIds(long[] ids) {
...
+     * Enqueue a new download batch.
+     *
+     * @param batch the parameters specifying this batch
+     * @return an ID for the batch, unique across the system.  This ID is used to make future
+     * calls related to this batch.
+     */
+    public long enqueue(RequestBatch batch) {
+        long batchId = insert(batch);
+        for (Request request : batch.getRequests()) {
+            request.setBatchId(batchId);
```
blundell: `batch.setRequestId(batchId);` and do the loop internally? (hiding your innards)

blundell:
:+1: more :eyes:

library/src/main/java/com/novoda/downloadmanager/lib/RequestBatch.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class RequestBatch {
+
+    private final BatchInfo batchInfo;
+    private final List<Request> requests;
+
+    public static RequestBatch newInstance(String title, String description, String bigPictureUrl) {
```
amlcurran: *Personally* I would not go for `newInstance` because I think that is a very Novoda-y thing and not a general pattern - external users wouldn't be very familiar with it. Perhaps a builder would be nicer? I feel like the params should be optional too

amlcurran: (but this is personal preference and I'll go with da flow)

zegnus:
:+1: PR description with sample code

library/src/main/java/com/novoda/downloadmanager/lib/RequestBatch.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public String getDescription() {
+        return batchInfo.getDescription();
+    }
+
+    public String getBigPictureUrl() {
+        return batchInfo.getBigPictureUrl();
+    }
+
+    public List<Request> getRequests() {
```
zegnus: It would be awesome if this could be avoided so we don't expose the internal representation of the data that this class holds, that would problably mean to push logic into this class :)

library/src/main/java/com/novoda/downloadmanager/lib/RequestBatch.java
blundell: no that's true they should be optional, not everyone wants a notification. Therefore when optional the builder pattern works (which is what the other `Request` object adhere's to. Also allows us to add more params more extensibly than this does.)

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
Electryc: Even if we set the `requestid` for each request inside the batch, we will anyway have to do a loop here in order to `insert` each request into the db unless we want to do that on the batch as well, but that seems a little bit too much responsibility for the `RequestBatch`.

blundell: mhm true

ronocod:
All comments addressed, and we added an example of enqueuing a single request to the serial demo

PR #39, opened 2015-06-29
Allow custom database filename

StylingAndroid:
### The problem

A project upon which I'm working is already using a database with the filename `downloads.db`. For legacy reasons it is not possible to change this so it would be useful if `download-manager` could be configured to use an alternate database filename.

### The solution
Adds the ability to override the database filename via an AndroidManifest metadata value in a similar fashion to the existing `MaxConcurrentDownloads` implementation. This provides me with a simple way to overcome my specific problem while offering additional functionality for power users. All this without actually changing the default, out-of-the-box behaviour.

### Bonus points
I also fixed 3 (albeit simple) lint warnings.

### DIFF
--- a/library/src/test/java/com/novoda/downloadmanager/lib/DatabaseFilenameProviderTest.java
+++ b/library/src/test/java/com/novoda/downloadmanager/lib/DatabaseFilenameProviderTest.java
@@ -1,0 +1,71 @@
+package com.novoda.downloadmanager.lib;
+
+import android.content.pm.ApplicationInfo;
+import android.content.pm.PackageManager;
+import android.os.Bundle;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mock;
+
+import static org.fest.assertions.api.Assertions.assertThat;
+import static org.mockito.Matchers.anyString;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.when;
+import static org.mockito.MockitoAnnotations.initMocks;
+
+public class DatabaseFilenameProviderTest {
+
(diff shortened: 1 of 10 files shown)


library/src/test/java/com/novoda/downloadmanager/lib/DatabaseFilenameProviderTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.junit.Test;
+import org.mockito.Mock;
+
+import static org.fest.assertions.api.Assertions.assertThat;
+import static org.mockito.Matchers.anyString;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.when;
+import static org.mockito.MockitoAnnotations.initMocks;
+
+public class DatabaseFilenameProviderTest {
```
blundell: :+1:

blundell:
We should add it's use to the demo project. At the moment we have a serial demo and a parrallel demo. The usual pattern is to have a simple demo and an advanced demo (this would go into advanced, along with parralel execution I guess).

StylingAndroid:
I've added an example to the parallel demo

blundell:
makes sense :+1:

xrigau:
:+1:

Electryc:
@blundell at the moment the serial and the parallel demos do the same thing but as the name stands one in serial and one in parallel. If we are going to split into simple and advanced I agree that the parallel stuff plus the custom database would go into advanced, and moreover this is better as we add more advanced features, in order not to clutter with multiple apps :+1:

PR #42, opened 2015-06-29
Demo simplification

blundell:
Moves to having our normal NOS modules for downloads, we should now add all new behaviour and demo changes to the p

Removes
```
demo_serial
demo_parallel
```
For
```
demo-simple
demo-extended
```

simple demo does the least code possible to do a download and show downloads in progress
extended does everything else


Electryc:
Nice :+1:

PR #37, opened 2015-06-25
Handle batch failures, cancellation and notifications

ronocod:
* Rework the `DownloadService`, `DownloadThread`, `DownloadReceiver` and `DownloadNotifier` to operate in terms of batches, instead of individual downloads
* Adds status and notification visibility columns to the Batch table
* Update the status of the batch when the status of a corresponding download changes
* Fix downloads not being marked as active when they begin downloading: https://github.com/novoda/download-manager/commit/8af7cc38d088581fa401fd527d978298a45b0d94
* Handles cancellation correctly for an entire batch of downloads
* Stops retrying on 404 errors

Next steps for a new PR:
* Removed unused columns on the `Request` table
* Allow querying for batch status from the API
* Fix `OpenHelper` trying to use system DownloadManager as our DownloadManager
* Test with `DownloadClientReadyChecker` functionality from https://github.com/novoda/download-manager/pull/30
* Handle clicking on the notification to open a Downloads screen

PR #35, opened 2015-06-22
Batch requests

Electryc:
**This is currently work in progress towards enabling the download manager to handle batched downloads and to progress with the implementation of the feature hence this is opened against a feature branch.**

For now, each `Request` needs a `batchId` in order to be mapped to a specific batch. *This will be changed in a following PR where the download manager will handle only batches, but not individual requests*. 

This particular PR focuses more on the notification side of the download manager, which now displays only one notification per batch. In order to display the correct progress of the entire batch we had to fetch the total size of the batch by doing a HEAD call for each request in the batch to get the `Content-Length` header value.

On the SQL side, a `batch` table is created alongside with a new `batch_id` column for the `request` table.

PR #43, opened 2015-06-29
Batch downloads feature

### LINKED ISSUE #18: Remove all upgrade code
we only have one version of the database table at the moment lets simplify the code we have to maintain

> [DownloadProvider.onUpgrade:270] Upgrading downloads database from version 0 to version 109, which will destroy all old data

Electryc:
**This contains breaking changes in both database and API**

This PR is composed out of:
* #41 - Remove unused columns on Request and expose Batch API
* #37 - Handle batch failures, cancellation and notifications
* #36 - Improve API for enqueuing downloads and fix notification UI
* #35 - Batch requests

Allows the client to download in batches, meaning that if a download in a batch is cancelled/failed the whole batch is cancelled/failed. The batch details (title, description, notification visibility) are removed from the `Request` and added to the batch itself. 

Example of usage:
```java
RequestBatch batch = new RequestBatch.Builder()
                .withTitle("Premium Video")
                .withDescription("A video with an advert")
                .withBigPictureUrl(VIDEO_IMAGE_URL)
                .withVisibility(NotificationVisibility.ACTIVE_OR_COMPLETE)
                .build();

Request videoRequest = new Request(videoUri);
videoRequest.setExtra("video");
batch.addRequest(videoRequest);

Request advertRequest = new Request(advertUri);
advertRequest.setExtra("advert");
batch.addRequest(advertRequest);

long batchId = downloadManager.enqueue(batch);
```
If a single request is enqueued, a batch will be created and the request will be added to it. 

_In the below example we enqueue a batch of 2 requests and once the batch is cancelled all the downloads are cancelled_:

batch

Fixes #18


friedger:
:+1:

ouchadam:
:100:

ouchadam:
lets do this!

library/src/main/java/com/novoda/downloadmanager/lib/BatchStatusUpdater.java
```diff
@@ -0,0 +1,81 @@
...
+        // check for error statuses first
+        for (int i = 0; i < statusCounts.size(); i++) {
+            int statusCode = statusCounts.keyAt(i);
+            if (Downloads.Impl.isStatusError(statusCode) && statusCounts.get(statusCode) > 0) {
+                return statusCode;
+            }
+        }
+
+        for (Integer status : PRIORITISED_STATUSES) {
+            if (statusCounts.get(status) > 0) {
```
zegnus: This causes a change of the object type for every single value on your list, in order to prevent it I suggest:

```java
private static final int[] PRIORITISED_STATUSES = new int[] {
            STATUS_CANCELED,
            STATUS_RUNNING,

            // Paused statuses
            STATUS_PAUSED_BY_APP,
            STATUS_WAITING_TO_RETRY,
            STATUS_WAITING_FOR_NETWORK,
            STATUS_QUEUED_FOR_WIFI,

            STATUS_PENDING,
            STATUS_SUCCESS
    };
```

As this is a predefined set of values that will never change (you don't add or remove elements from that array)

```java
private static final int PRIORITISED_STATUSES_SIZE = PRIORITISED_STATUSES.length;
```
with
```java
for (int i = 0; i < PRIORITISED_STATUSES_SIZE; i++) {
  if (statusCounts.get(PRIORITISED_STATUSES[i]) > 0) {
    return PRIORITISED_STATUSES[i];
  }
}
```

And you avoid the preallocation of the iterator

Reference https://medium.com/google-developers/developing-for-android-ii-bb9a51f8c8b9

zegnus: Actually there is no iterator on Arrays
> with the exception of arrays, which can be used with the foreach syntax without causing an Iterator allocation

Good to know :)

zegnus:
No tests?

PR #44, opened 2015-06-30
Add extras column to DownloadsByBatch view so we can filter on it

ronocod:
Filtering on the extras column allows us to use `setExtra()` on a `Request` to tag a request and make it easier to retrieve it from a batch later on.

**Example**

```java
Request request = new Request(thumbnailUri);
request.setExtra("thumbnail");
downloadManager.enqueue(request);

// Later

Query query = new Query().setFilterByBatchId(batchId)
        .setFilterByExtras("thumbnail");
downloadManager.query(query);

```


ouchadam:
:+1:

xrigau:
wouldn't it be better to have a specific column for the batch? the extras column is supposed to be used by the client, not the library itself

ronocod:
Yeah we could have a `batch_tag` column or something like that, then the extras would be available for clients to use

PR #45, opened 2015-06-30
Fix bug wrong download status when is not ready to download

zegnus:
## Description of the bug ##

When there are several queued items to be downloaded, the following bug occurs

Force the `DownloadClientReadyChecker` to return false under some circumstances, for example when there is no Wi-Fi

- 2 queued items
- 1 item is downloading, 1 item is queued
- Switch off Wi-Fi

Expected:
- 2 items moved to queued
- No notification

Result:
- Previous downloading item moves to queued
- Previous queued item moves to downloading
- Notification stays in place

## Solution implemented ##

- The submitted thread for a particular download now does a check against the `DownloadClientReadyChecker`
- The `DownloadService` is stopped only if there are no any active download threads

Paired with @ronocod

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
@@ -462,6 +462,10 @@ class DownloadInfo {
         }
     }
 
+    public boolean isActive() {
+        return mSubmittedTask != null && !mSubmittedTask.isDone();
+    }
+
     public void updateStatus(int status) {
         mStatus = status;
         downloadStatusContentValues.clear();
@@ -470,14 +474,7 @@ class DownloadInfo {
     }
 
     private boolean isClientReadyToDownload() {
-        ConnectivityManager connectivityManager = (ConnectivityManager) mContext.getSystemService(Context.CONNECTIVITY_SERVICE);
-        NetworkInfo networkInfo = connectivityManager.getNetworkInfo(ConnectivityManager.TYPE_WIFI);
-        if (networkInfo != null && networkInfo.isConnected()) {
-            return true;
-        } else {
-            return false;
-        }
-        //return downloadClientReadyChecker.isAllowedToDownload();
+        return downloadClientReadyChecker.isAllowedToDownload();
     }
 
     /**
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
@@ -444,7 +444,7 @@ class DownloadInfo {
      */
     public boolean isReadyToDownload() {
         synchronized (this) {
-            return isDownloadManagerReadyToDownload() && isClientReadyToDownload();
+            return isClientReadyToDownload() && isDownloadManagerReadyToDownload();
         }
     }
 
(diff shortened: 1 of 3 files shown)


library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
```diff
@@ -444,7 +444,7 @@ private NetworkState checkSizeAllowedForNetwork(int networkType) {
      */
     public boolean isReadyToDownload() {
         synchronized (this) {
-            return isDownloadManagerReadyToDownload() && isClientReadyToDownload();
+            return isClientReadyToDownload() && isDownloadManagerReadyToDownload();
```
blundell: does the order matter?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
```diff
@@ -419,9 +419,12 @@ private void deleteFileAndMediaReference(DownloadInfo info) {
 
     private boolean kickOffDownloadTaskIfReady(boolean isActive, DownloadInfo info) {
```
blundell: it looks like this method always passes false for isActive, is that correct?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
Electryc: I think this is a bit of perf improvement. If the first one is `false` then the second won't be checked as it is an `&&`. If `isClientReadyToDownload()` takes less to execute than the other one and it is more likely to return `false` then I think we have a little bit of improvement.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
ronocod: No, `isActive` lives outside of the loop so it'll be set to true after it iterates over an active download

Electryc: The confusion and strangeness comes from the fact that the `isActive` is "accumulated" by passing it, applying an `|` and then returning it, and at the next pass in the while loop (https://github.com/novoda/download-manager/blob/Fix-bug-wrong-download-status-when-is-not-ready-to-download/library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java#L330) the method will be executed with the accumulated value up until now.

blundell: ah i missed that it was a loop and the input var changed lolz :+1: nice expl

Electryc:
LGTM :+1: letting @blundell to :eyes:

blundell:
yeah not quite sure, but if you are convinved :)

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
blundell: :x: ah ma gurd

Electryc: So it did matter in the end, never give up on your instincts @blundell !

zegnus: @blundell can you elaborate?

The idea is that the client might have business logic that has nothing to do with the downloading by itself and so should fail/checked first. For example rooted device or user not logged in.

blundell: https://github.com/novoda/download-manager/pull/52 `isDownloadManagerReadyToDownload()` checks if a download has been successfully completed and returns false, therefore `isClientReadyToDownload()` is not invoked. When the other away around the client is asked if they are allowed to download an already downloaded file...

zegnus: :+1:

PR #47, opened 2015-06-30
Broadcast batch completion events

ronocod:
Adds broadcasting of events for batch completion. 

**Example Receiver**
```xml
<receiver
  android:name="com.novoda.downloadmanager.demo.extended.DownloadBatchCompletionReceiver"
  android:exported="false">
  <intent-filter>
    <action android:name="com.novoda.downloadmanager.BATCH_COMPLETE" />
  </intent-filter>
</receiver>
```

batch_complete

### DIFF
--- a/demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DownloadBatchCompletionReceiver.java
+++ b/demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DownloadBatchCompletionReceiver.java
@@ -16,11 +16,9 @@ public class DownloadBatchCompletionReceiver extends BroadcastReceiver {
 
     @Override
     public void onReceive(@NonNull Context context, @NonNull Intent intent) {
-        if (DownloadManager.ACTION_BATCH_COMPLETE.equals(intent.getAction())) {
-            long batchId = intent.getLongExtra(DownloadManager.EXTRA_BATCH_ID, UNKNOWN_BATCH_ID);
-            Toast.makeText(context, "Batch completed with id: " + batchId, Toast.LENGTH_SHORT).show();
-            Log.d(TAG, "Batch completed: " + batchId);
-        }
+        long batchId = intent.getLongExtra(DownloadManager.EXTRA_BATCH_ID, UNKNOWN_BATCH_ID);
+        Toast.makeText(context, "Batch completed with id: " + batchId, Toast.LENGTH_SHORT).show();
+        Log.d(TAG, "Batch completed: " + batchId);
     }
 
 }
--- a/demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DownloadBatchCompletionReceiver.java
+++ b/demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DownloadBatchCompletionReceiver.java
@@ -1,0 +1,26 @@
+package com.novoda.downloadmanager.demo.extended;
+
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.Intent;
+import android.support.annotation.NonNull;
+import android.util.Log;
+import android.widget.Toast;
+
+import com.novoda.downloadmanager.lib.DownloadManager;
+
+public class DownloadBatchCompletionReceiver extends BroadcastReceiver {
+
+    private static final int UNKNOWN_BATCH_ID = -1;
+    private static final String TAG = DownloadBatchCompletionReceiver.class.getSimpleName();
+
+    @Override
+    public void onReceive(@NonNull Context context, @NonNull Intent intent) {
+        if (DownloadManager.ACTION_BATCH_COMPLETE.equals(intent.getAction())) {
+            long batchId = intent.getLongExtra(DownloadManager.EXTRA_BATCH_ID, UNKNOWN_BATCH_ID);
+            Toast.makeText(context, "Batch completed with id: " + batchId, Toast.LENGTH_SHORT).show();
+            Log.d(TAG, "Batch completed: " + batchId);
+        }
+    }
+
+}
--- a/demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/MainActivity.java
+++ b/demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/MainActivity.java
@@ -35,7 +35,7 @@ public class MainActivity extends AppCompatActivity implements QueryForDownloads
         listView = (ListView) findViewById(R.id.main_downloads_list);
         downloadManager = DownloadManagerBuilder.from(this)
                 .withVerboseLogging()
-                .build(getContentResolver());
+                .build();
 
         findViewById(R.id.single_download_button).setOnClickListener(
                 new View.OnClickListener() {
--- a/demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/MainActivity.java
+++ b/demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/MainActivity.java
@@ -1,5 +1,9 @@
 package com.novoda.downloadmanager.demo.extended;
 
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.Intent;
+import android.content.IntentFilter;
 import android.net.Uri;
 import android.os.Bundle;
 import android.os.Environment;
@@ -8,6 +12,7 @@ import android.support.v7.app.AppCompatActivity;
 import android.util.Log;
 import android.view.View;
 import android.widget.ListView;
+import android.widget.Toast;
 
 import com.novoda.downloadmanager.DownloadManagerBuilder;
 import com.novoda.downloadmanager.demo.R;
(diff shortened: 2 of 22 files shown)


library/src/main/java/com/novoda/downloadmanager/lib/BatchCompletionBroadcaster.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    public void notifyBatchCompletedFor(long batchId) {
+        Intent intent = new Intent(ACTION_BATCH_COMPLETE);
+        intent.setPackage(packageName);
+        intent.putExtra(EXTRA_BATCH_ID, batchId);
+        localBroadcastManager.sendBroadcast(intent);
+    }
+
+    public void registerReceiver(BatchCompletionListener batchCompletionListener) {
```
xrigau: this should be a setter since you can only register one :x:

xrigau: since this is a library it'd be better if you hold a list of `BatchCompletionListener`, then register would add to the list and unregister would unregister one in particular

take this as example [register](http://developer.android.com/reference/android/content/Context.html#registerReceiver(android.content.BroadcastReceiver, android.content.IntentFilter)) and [unregister](http://developer.android.com/reference/android/content/Context.html#unregisterReceiver(android.content.BroadcastReceiver))

library/src/main/java/com/novoda/downloadmanager/lib/BatchCompletionBroadcaster.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public void unregisterReceiver() {
+        this.batchCompletionListener = null;
+        localBroadcastManager.unregisterReceiver(batchCompletedReceiver);
+    }
+
+    private final BroadcastReceiver batchCompletedReceiver = new BroadcastReceiver() {
+        @Override
+        public void onReceive(@NonNull Context context, @NonNull Intent intent) {
+            long batchId = intent.getLongExtra(DownloadManager.EXTRA_BATCH_ID, -1);
```
xrigau: what is `-1`?

blundell:
Why doesn't a batch completing just act like a normal download completing? i.e. you use the same mechanism for a download completing but the id is the batch?

Electryc:
@blundell do you mean as in this case (https://github.com/novoda/download-manager/blob/master/library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java#L268)?This is just for when one download completes and it fires up an intent. We wanted to abstract out that receiver, but maybe we'll end up exposing it in the end, do you think it would be a better idea?

Electryc:
Closing while exploring another solution.

Electryc:
Moved to having just a broadcast rather than a listener so that client apps can listen for it.

blundell:
nice, yea better to have a receiver and follow the pattern we already have :+1: 

However as you say the original broadcast:
> This is just for when one download completes and it fires up an intent.

Why is there a difference, a batch of downloads is a group, I would say each individual download should not fire a broadcast and only the final download complete (all the batch) should be fired. This means the user knows less about the batching internals and just when it's finished. Do you know what I mean?

HOWEVER what you have done adds more flexibility in that users know about the whole downloading completing or individual parts so thats better I guess :) just I don't like exposing batches everywhere

Electryc:
I see what you mean now :+1: I meant that the mechanism is a little bit different where an individual completion download broadcast is fired as opposed to where we hook in the broadcast of the completion of a batch, but you're right maybe we're exposing a bit too much.

blundell:
reminder release `0.1.0` when this is merged

xrigau:
:+1: @blundell :shipit: ?

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DownloadBatchCompletionReceiver.java
```diff
@@ -0,0 +1,26 @@
...
+import com.novoda.downloadmanager.lib.DownloadManager;
+
+public class DownloadBatchCompletionReceiver extends BroadcastReceiver {
+
+    private static final int UNKNOWN_BATCH_ID = -1;
+    private static final String TAG = DownloadBatchCompletionReceiver.class.getSimpleName();
+
+    @Override
+    public void onReceive(@NonNull Context context, @NonNull Intent intent) {
+        if (DownloadManager.ACTION_BATCH_COMPLETE.equals(intent.getAction())) {
```
blundell: devils advocate, do you need this check since it's declared in the manifest to only receive these events?

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/MainActivity.java
```diff
@@ -21,7 +21,7 @@
 
 public class MainActivity extends AppCompatActivity implements QueryForDownloadsAsyncTask.Callback {
     private static final String TAG = MainActivity.class.getSimpleName();
-    private static final String BIG_FILE = "http://download.thinkbroadband.com/50MB.zip";
+    private static final String BIG_FILE = "http://download.thinkbroadband.com/20MB.zip";
```
blundell: keeps getting smaller lol

library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
```diff
@@ -30,6 +30,7 @@ public DownloadManagerBuilder withVerboseLogging() {
     }
 
     public DownloadManager build(ContentResolver contentResolver) {
+        // TODO: is the parameter need here any more? Could use context.getContentResolver()
```
blundell: delete it then, do the todo

library/src/main/java/com/novoda/downloadmanager/lib/BatchCompletionBroadcaster.java
```diff
@@ -0,0 +1,30 @@
...
+class BatchCompletionBroadcaster {
+
+    static final String ACTION_BATCH_COMPLETE = "com.novoda.downloadmanager.BATCH_COMPLETE";
+    static final String EXTRA_BATCH_ID = "com.novoda.downloadmanager.EXTRA.batch_id";
+
+    private final Context context;
+    private final String packageName;
+
+    public static BatchCompletionBroadcaster newInstance(Context context) {
+        return new BatchCompletionBroadcaster(context, context.getApplicationContext().getPackageName());
```
blundell: extract explantory var

library/src/main/java/com/novoda/downloadmanager/lib/BatchCompletionBroadcaster.java
```diff
@@ -0,0 +1,30 @@
+package com.novoda.downloadmanager.lib;
+
+import android.content.Context;
+import android.content.Intent;
+
+class BatchCompletionBroadcaster {
+
+    static final String ACTION_BATCH_COMPLETE = "com.novoda.downloadmanager.BATCH_COMPLETE";
```
blundell: `"com.novoda.downloadmanager.action.BATCH_COMPLETE";`

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DownloadBatchCompletionReceiver.java
ronocod: double devils advocate, users of the library can declare it with whatever IntentFilter they want

library/src/main/java/com/novoda/downloadmanager/lib/BatchCompletionBroadcaster.java
```diff
@@ -0,0 +1,30 @@
...
+    static final String EXTRA_BATCH_ID = "com.novoda.downloadmanager.EXTRA.batch_id";
```
blundell: nearly!
`"com.novoda.downloadmanager.extract.BATCH_ID";`

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/MainActivity.java
ronocod: It gets smaller as my patience gets thinner

library/src/main/java/com/novoda/downloadmanager/lib/BatchCompletionListener.java
```diff
@@ -0,0 +1,5 @@
+package com.novoda.downloadmanager.lib;
+
+public interface BatchCompletionListener {
```
blundell: public yes?

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DownloadBatchCompletionReceiver.java
blundell: ? but then they wouldn't get the callbacks lol  , i.e. you are repeating yourself with this if check

ronocod: True, it would be broken anyway then. :+1: for removing this `if`

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/MainActivity.java
amlcurran: "BIG"_FILE

library/src/main/java/com/novoda/downloadmanager/lib/BatchCompletionListener.java
ronocod: Not used any more https://github.com/novoda/download-manager/commit/8ae64aacecf611f3dc15c294fd5a58958e63157b

ronocod:
@blundell comments addressed

PR #49, opened 2015-07-01
Optimisation Fix - Get full batch size in one query

zegnus:
This PR replaces a full loop over all the rows retrieved by a query by one single query that does all the sums in the query itself

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/BatchStatusRepository.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/BatchStatusRepository.java
@@ -84,18 +84,15 @@ class BatchStatusRepository {
         long totalSize = 0;
         try {
             String[] selectionArgs = {String.valueOf(batchId)};
-            cursor = resolver.query(ALL_DOWNLOADS_CONTENT_URI,
-                    null,
+            cursor = resolver.query(
+                    ALL_DOWNLOADS_CONTENT_URI,
+                    new String[]{"sum(" + COLUMN_TOTAL_BYTES + ")"},
                     COLUMN_BATCH_ID + " = ?",
                     selectionArgs,
                     null);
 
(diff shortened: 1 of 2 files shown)


friedger:
:+1:

friedger:
I assume that all tests still pass :-)

library/src/main/java/com/novoda/downloadmanager/lib/BatchStatusRepository.java
```diff
@@ -84,18 +84,15 @@ long getBatchSizeInBytes(long batchId) {
...
                     null);
 
-            int totalBytesColumnIndex = cursor.getColumnIndexOrThrow(COLUMN_TOTAL_BYTES);
-
-            while (cursor.moveToNext()) {
-                int individualDownloadSize = cursor.getInt(totalBytesColumnIndex);
-                totalSize += individualDownloadSize;
-            }
+            cursor.moveToFirst();
+            totalSize = cursor.getLong(0);
```
ouchadam: what is this column id? I thought you needed to define another column variable in the raw query and then read from that?

friedger: It is the column index of the projection.


  [image: --]
Friedger Müffke
[image: http://]about.me/friedger
<http://about.me/friedger>


2015-07-01 12:00 GMT+02:00 Adam Brown <notifications@github.com>:

blundell:
if you say *Optimisation* you beg the question where are your performance figures to show the before and after gain?

xrigau:
lol @friedger don't think there are many tests here

xrigau:
:+1: @blundell

amlcurran:
:+1: It'd be really good if someone could do a H&T on advanced SQL stuff, I don't know any of this :)

ronocod:
I don't know enough to do a H&T, some useful ones here though: https://www.sqlite.org/lang_aggfunc.html
and some less useful ones: https://www.sqlite.org/lang_corefunc.html

PR #52, opened 2015-07-02
Fixes too many callbacks about the rules for already downloaded files

blundell:
Because the logical order was changed this meant clients had access to the already successfully downloaded files and could check rules against them. https://github.com/novoda/download-manager/pull/45/files#diff-68db38a7048316539cb0931793be1808R447 Now we reverted back to the download manager doing the first check to stop this crazyness.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
@@ -440,12 +440,13 @@ class DownloadInfo {
      * create a {DownloadThread} and enqueue it into given
      * {@link Executor}.
      *
-     * @param collatedDownloadInfo
      * @return If actively downloading.
      */
     public boolean isReadyToDownload(CollatedDownloadInfo collatedDownloadInfo) {
         synchronized (this) {
-            return isClientReadyToDownload(collatedDownloadInfo) && isDownloadManagerReadyToDownload();
+            // This order MATTERS
+            // it means completed downloads will not be accounted for in later downloadInfo queries
+            return isDownloadManagerReadyToDownload() && isClientReadyToDownload(collatedDownloadInfo);
         }
     }
 
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadThread.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadThread.java
@@ -196,7 +196,7 @@ class DownloadThread implements Runnable {
         long batchSizeInBytes = batchStatusRepository.getBatchSizeInBytes(mInfo.batchId);
 
         if (!mInfo.isReadyToDownload(new CollatedDownloadInfo(batchSizeInBytes))) {
-            Log.d("Download " + mInfo.mId + " is not ready to download: skipping");
+            Log.d("Download " + mInfo.mId + " is not ready to download; skipping");
             return;
         }
 


library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
```diff
@@ -440,12 +440,13 @@ private NetworkState checkSizeAllowedForNetwork(int networkType) {
      * create a {DownloadThread} and enqueue it into given
      * {@link Executor}.
      *
-     * @param collatedDownloadInfo
```
Electryc: This should be kept :tongue:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
```diff
@@ -440,12 +440,13 @@ private NetworkState checkSizeAllowedForNetwork(int networkType) {
...
      * @return If actively downloading.
      */
     public boolean isReadyToDownload(CollatedDownloadInfo collatedDownloadInfo) {
         synchronized (this) {
-            return isClientReadyToDownload(collatedDownloadInfo) && isDownloadManagerReadyToDownload();
+            // This order MATTERS
+            // it means completed downloads will not be accounted for in later downloadInfo queries
+            return isDownloadManagerReadyToDownload() && isClientReadyToDownload(collatedDownloadInfo);
```
Electryc: So the `isClientReadyToDownload(collatedDownloadInfo)` has a side effect?

ouchadam: it's `isDownloadManagerReadyToDownload` which checks if the download has already completed

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
blundell: I delete them when nobody has written javadoc about the parameter, if it's undocumented whats the point in it?

ronocod: :+1: there's no point having them if there's no documentation IMO

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
zegnus: @Electryc The thing is that in the client checking mechanism, if it fails it will trigger the UI for displaying the errors. 

This fix prevents triggering them for already completed downloads so the client check is not invoked.

@blundell correct me if I'm wrong

zegnus:
:+1: LGTM

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
zegnus: @blundell I would add that info to the comment because I think that is still not clear

blundell: what the client does is up to the client, the comment points out what happens inside of `isDownloadManagerReadyToDownload`

blundell:
@Electryc :eyes:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
Electryc: :+1: yes you're right

library/src/main/java/com/novoda/downloadmanager/lib/DownloadInfo.java
Electryc: :+1:

Electryc:
LGTM :ship:

PR #53, opened 2015-07-02
Add ability to delete batches

ronocod:
Adds marking batches for deletion to the public API.

The `DownloadService` will find batches that have been marked for deletion and:
* delete the files for their downloads
* delete the DB entries for those downloads
* delete the DB entries for the batches themselves

Before Deletion | After Deletion
---|---
before | after

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
@@ -468,7 +468,7 @@ public class DownloadManager {
     }
 
     /**
-     * Cancel downloads and remove them from the download manager.  Each download will be stopped if
+     * Cancel batch downloads and remove them from the download manager.  Each download will be stopped if
      * it was running, and it will no longer be accessible through the download manager.
      * If there is a downloaded file, partial or complete, it is deleted.
      *
@@ -476,7 +476,7 @@ public class DownloadManager {
      * @return the number of downloads actually removed
      */
     public int removeBatch(long... ids) {
-        return markDownloadRowForDeletion(ids);
+        return markBatchRowForDeletion(ids);
     }
 
     /**
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
@@ -402,7 +402,7 @@ public class DownloadManager {
             cursor = mResolver.query(Downloads.Impl.CONTENT_URI, new String[]{"_id"}, Downloads.Impl.COLUMN_FILE_NAME_HINT + "=?", new String[]{uri.toString()}, null);
             if (cursor.moveToFirst()) {
                 long id = cursor.getLong(cursor.getColumnIndexOrThrow("_id"));
-                markDownloadRowForDeletion(id);
+                markDownloadRowsForDeletion(id);
                 return;
             }
             Log.e("Didn't delete anything for uri: " + uri);
@@ -421,7 +421,7 @@ public class DownloadManager {
      * @param ids the IDs of the downloads to be marked 'deleted'
      * @return the number of downloads actually updated
      */
-    public int markDownloadRowForDeletion(long... ids) {
+    public int markDownloadRowsForDeletion(long... ids) {
         if (ids == null || ids.length == 0) {
             throw new IllegalArgumentException("called with nothing to remove. input param 'ids' can't be null");
         }
@@ -443,17 +443,17 @@ public class DownloadManager {
      * @param ids the IDs of the downloads to remove
      * @return the number of downloads actually removed
      */
-    public int removeDownload(long... ids) {
-        return markDownloadRowForDeletion(ids);
+    public int removeDownloads(long... ids) {
+        return markDownloadRowsForDeletion(ids);
     }
 
     /**
      * Marks the specified batch as 'to be deleted'. Actual cleanup of this row is done in DownloadService.
      *
-     * @param ids the IDs of the downloads to be marked 'deleted'
-     * @return the number of downloads actually updated
+     * @param ids the IDs of the batches to be marked 'deleted'
+     * @return the number of batches actually updated
      */
-    public int markBatchRowForDeletion(long... ids) {
+    public int markBatchRowsForDeletion(long... ids) {
         if (ids == null || ids.length == 0) {
             throw new IllegalArgumentException("called with nothing to remove. input param 'ids' can't be null");
         }
@@ -470,13 +470,13 @@ public class DownloadManager {
     /**
      * Cancel batch downloads and remove them from the download manager.  Each download will be stopped if
      * it was running, and it will no longer be accessible through the download manager.
-     * If there is a downloaded file, partial or complete, it is deleted.
+     * If there are any downloaded files, partial or complete, they will be deleted.
      *
-     * @param ids the IDs of the downloads to remove
-     * @return the number of downloads actually removed
+     * @param ids the IDs of the batches to remove
+     * @return the number of batches actually removed
      */
-    public int removeBatch(long... ids) {
-        return markBatchRowForDeletion(ids);
+    public int removeBatches(long... ids) {
+        return markBatchRowsForDeletion(ids);
     }
 
     /**
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
@@ -396,13 +396,13 @@ public class DownloadManager {
         return ContentUris.parseId(downloadUri);
     }
 
-    public void markDeleted(URI uri) {
+    public void markDownloadForDeletion(URI uri) {
         Cursor cursor = null;
         try {
             cursor = mResolver.query(Downloads.Impl.CONTENT_URI, new String[]{"_id"}, Downloads.Impl.COLUMN_FILE_NAME_HINT + "=?", new String[]{uri.toString()}, null);
             if (cursor.moveToFirst()) {
                 long id = cursor.getLong(cursor.getColumnIndexOrThrow("_id"));
-                markRowDeleted(id);
+                markDownloadRowForDeletion(id);
                 return;
             }
             Log.e("Didn't delete anything for uri: " + uri);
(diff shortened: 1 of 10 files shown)


library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
```diff
@@ -443,8 +443,40 @@ public int markRowDeleted(long... ids) {
      * @param ids the IDs of the downloads to remove
      * @return the number of downloads actually removed
      */
-    public int remove(long... ids) {
-        return markRowDeleted(ids);
+    public int removeDownloads(long... ids) {
```
Electryc: Wondering why did you created this method, isn't it the same as `markDownloadRowsForDeletion()`? Or maybe the later should be private rather than exposed.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
```diff
@@ -443,8 +443,40 @@ public int markRowDeleted(long... ids) {
...
+
+    /**
+     * Cancel batch downloads and remove them from the download manager.  Each download will be stopped if
+     * it was running, and it will no longer be accessible through the download manager.
+     * If there are any downloaded files, partial or complete, they will be deleted.
+     *
+     * @param ids the IDs of the batches to remove
+     * @return the number of batches actually removed
+     */
+    public int removeBatches(long... ids) {
```
Electryc: Same here, should `markBatchRowsForDeletion()` be private?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
```diff
@@ -443,8 +443,40 @@ public int markRowDeleted(long... ids) {
...
+        return markDownloadRowsForDeletion(ids);
+    }
+
+    /**
+     * Marks the specified batch as 'to be deleted'. Actual cleanup of this row is done in DownloadService.
+     *
+     * @param ids the IDs of the batches to be marked 'deleted'
+     * @return the number of batches actually updated
+     */
+    public int markBatchRowsForDeletion(long... ids) {
```
Electryc: Same here, should this be private?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
```diff
@@ -357,24 +357,34 @@ private boolean updateLocked() {
...
+                String title = batchesCursor.getString(titleIndex);
+                String description = batchesCursor.getString(descriptionIndex);
+                String bigPictureUrl = batchesCursor.getString(bigPictureUrlIndex);
+                int status = batchesCursor.getInt(statusIndex);
                 @NotificationVisibility.Value int visibility = batchesCursor.getInt(visibilityColumn);
 
                 BatchInfo batchInfo = new BatchInfo(title, description, bigPictureUrl, visibility);
 
-                List<DownloadInfo> batchDownloads = new ArrayList<>();
+                List<DownloadInfo> batchDownloads = new ArrayList<>(1);
```
Electryc: Why initial capacity of `1` hence you have to resize later on? Unless we expect here to have just one batch.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
```diff
@@ -357,24 +357,34 @@ private boolean updateLocked() {
...
 
     private List<DownloadBatch> fetchBatches(Collection<DownloadInfo> downloads) {
-        List<DownloadBatch> batches = new ArrayList<>();
         Cursor batchesCursor = resolver.query(Downloads.Impl.BATCH_CONTENT_URI, null, null, null, null);
-        batches.clear();
+        List<DownloadBatch> batches = new ArrayList<>(batchesCursor.getCount());
+        List<Long> forDeletion = new ArrayList<>();
         try {
             int idColumn = batchesCursor.getColumnIndexOrThrow(Downloads.Impl.Batches._ID);
+            int deleteIndex = batchesCursor.getColumnIndex(Downloads.Impl.Batches.COLUMN_DELETED);
```
ouchadam: soooo who's gona port to SQLAnalyser?

Electryc: :100: for SQLiteAnalyser

blundell: careful, will SQLiteAnalyser expose all of these column names as public? Where as at the moment they cannot be accessed outside of the Library?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
zegnus: Yeah this is weird right? We had that conversation as well and there are two irrelevant public methods there that just redirects.

If noone sees a reason for those methods, we'll more than happy to delete them.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
zegnus: https://github.com/novoda/download-manager/pull/53#discussion_r33847506

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
blundell: in this situation I would refer back to original source to see if there is something we have deleted that gives us more understanding.

Aaaand it doesn't help... http://androidxref.com/4.4.4_r1/xref/frameworks/base/core/java/android/app/DownloadManager.java#markRowDeleted

blundell: 4.3 actually http://androidxref.com/4.3_r2.1/xref/frameworks/base/core/java/android/app/DownloadManager.java#markRowDeleted

zegnus: Both public with different javadoc and one redirects to the other. Should be removed.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
ronocod: A batch will always have at least one download, and `ArrayList` defaults to a capacity of 0 on Android: http://androidxref.com/5.1.0_r1/xref/libcore/luni/src/main/java/java/util/ArrayList.java#81

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
ronocod: :+1: for removing

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
ouchadam: The columns are kind of public already since they're just strings/indexes but there's a queried column validation that happens in the update/insert, it's pretty cray cray

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
amlcurran: The More You Know :+1:

blundell: http://androidxref.com/4.2_r1/xref/libcore/luni/src/main/java/libcore/util/EmptyArray.java#29

xrigau: noice :+1:

ouchadam: SO MANY CHANCES FOR OPTIMISATIONS

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
ouchadam: +1 for removing, we don't need to copy the original (not so great) api

amlcurran:
Sorry for the merge conflicts @ronocod :(

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
Electryc: :+1: got it!

ronocod:
@amlcurran I live for the challenge of merge conflicts

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
ronocod: More *The More You Know*
Adding to an `ArrayList` that's full will double its capacity, unless the capacity is less than 6 in which case it will increase the capacity by 12: http://androidxref.com/5.1.0_r1/xref/libcore/luni/src/main/java/java/util/ArrayList.java#114

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
ronocod: Removed: https://github.com/novoda/download-manager/commit/ffc5000e9d890093dd7b2ded3586feb4edf10966

zegnus:
:+1:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
```diff
@@ -449,15 +449,25 @@ public int markRowDeleted(long... ids) {
...
+    public int removeBatches(long... ids) {
+        if (ids == null || ids.length == 0) {
+            throw new IllegalArgumentException("called with nothing to remove. input param 'ids' can't be null");
+        }
+        ContentValues values = new ContentValues();
+        values.put(Downloads.Impl.Batches.COLUMN_DELETED, 1);
+        // if only one id is passed in, then include it in the uri itself.
+        // this will eliminate a full database scan in the download service.
+        if (ids.length == 1) {
+            return mResolver.update(ContentUris.withAppendedId(Downloads.Impl.BATCH_CONTENT_URI, ids[0]), values, null, null);
```
amlcurran: :100: great idea

ronocod: It is, not our great idea though: http://androidxref.com/4.3_r2.1/xref/frameworks/base/core/java/android/app/DownloadManager.java#955

ouchadam:
LGTM :+1:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
blundell: yeah but if someone uses a public constant they you have a contract to maintain that (rather than them copying it by using a raw string - non public)

blundell: FYI feature on SQliteAnalyzer, make the constants package protected for libs

PR #54, opened 2015-07-02
Total and current batch size

ouchadam:
Adds the total and current size of the batch to the `DownloadBatch` 

- removes the `CollatedDownloadInfo` in favour of reusing the `DownloadBatch`


library/src/main/java/com/novoda/downloadmanager/lib/DownloadNotifier.java
```diff
@@ -167,27 +168,27 @@ private boolean shouldShowCompletedItem(int visibility) {
...
     }
 
-    private void addBatchToCluster(String tag, Map<String, List<DownloadBatch>> cluster, DownloadBatch batch) {
+    private void addBatchToCluster(String tag, Map<String, Collection<DownloadBatch>> cluster, DownloadBatch batch) {
         if (tag == null) {
             return;
         }
 
-        List<DownloadBatch> batches;
+        Collection<DownloadBatch> batches;
```
ronocod: :+1:

ronocod:
LGTM :+1:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadNotifier.java
```diff
@@ -167,27 +168,27 @@ private boolean shouldShowCompletedItem(int visibility) {
...
 
         if (cluster.containsKey(tag)) {
             batches = cluster.get(tag);
         } else {
-            batches = new ArrayList<>();
+            batches = new HashSet<>();
```
zegnus: Why of this change? Do you need random access or guarantee free of duplication? HashSet will have a slower insertion than an ArrayList

library/src/main/java/com/novoda/downloadmanager/lib/DownloadNotifier.java
```diff
@@ -167,27 +168,27 @@ private boolean shouldShowCompletedItem(int visibility) {
...
             cluster.put(tag, batches); // TODO not sure if this is right compared to ArrayListMultiMap
```
zegnus: If you really want to nail it, remove the comment and replace the type of your cluster to `SimpleArrayMap`

Reference here https://medium.com/google-developers/developing-for-android-ii-bb9a51f8c8b9

ronocod: Might need `ArrayMap` instead so that you can use the Iterator API: https://developer.android.com/reference/android/support/v4/util/ArrayMap.html

zegnus: There is no need to use the iterator, you can always get the size and iterate manually

zegnus: For example here

From
```java
for (String notificationId : clusters.keySet()) {
  int type = getNotificationTagType(notificationId);
```

to
```java
for (int i = 0, size = clusters.size(); i < size; i++) {
  String notificationId = clusters.keyAt(i);
  ...
}
```

ronocod: We could do that yeah, it'd just make this a bit messier: https://github.com/novoda/download-manager/blob/total-and-current-batch-size/library/src/main/java/com/novoda/downloadmanager/lib/DownloadNotifier.java#L189

ronocod: Commented at the same time :smile:

zegnus: If we are going to do this for performance, I would go for the SimpleArrayMap and the manual iteration :), is not daaat messy :P

ronocod: Yeah it doesn't look as bad as I thought actually

library/src/main/java/com/novoda/downloadmanager/lib/DownloadNotifier.java
ouchadam: doesn't need to be any more, it's been reverted

library/src/main/java/com/novoda/downloadmanager/lib/DownloadNotifier.java
ouchadam: it's been removed, sorry guys!

zegnus: @ouchadam uh? This comments are about this line `cluster.put(tag, batches); // TODO not sure if this is right compared to ArrayListMultiMap` :P

ouchadam: ahhhh I thought you were talking about the array list/hashset

ouchadam: would prefer to leave the TODO for another pretty pleaseeeee

zegnus: I'm OK moving this change into another PR so we can release another DM version right now

ronocod: :+1: agreed

zegnus:
:+1: LGTM with https://github.com/novoda/download-manager/pull/54#discussion_r33858809 on the queue

ronocod:
I hear the release train approaching!

library/src/main/java/com/novoda/downloadmanager/lib/DownloadBatch.java
```diff
@@ -2,18 +2,24 @@
 
 import java.util.List;
 
-class DownloadBatch {
+public class DownloadBatch {
```
blundell: in terms of library writing, we have now made this class available to clients, this means it's implementation is tied in to the version of the library.

I would have done a conversion from this internal represenation to something else `Collated..` and then THAT class can be responsible for keeping public API compatibility and this class can carry on being ruthlessly refactored.

I'd X this if it hadn't already been merged

library/src/main/java/com/novoda/downloadmanager/lib/DownloadClientReadyChecker.java
```diff
@@ -5,9 +5,10 @@
 
     /**
      * This method is executed on a background thread
-     * @param collatedDownloadInfo
+     *
+     * @param downloadBatch
```
blundell: don't add jdoc params unless you are going to fill them in with useful information

library/src/main/java/com/novoda/downloadmanager/lib/DownloadDeleter.java
```diff
@@ -0,0 +1,72 @@
...
+    public DownloadDeleter(ContentResolver resolver) {
+        this.resolver = resolver;
+    }
+
+    public void deleteFileAndDatabaseRow(DownloadInfo info) {
+        deleteFileAndMediaReference(info);
+        resolver.delete(info.getAllDownloadsUri(), null, null);
+    }
+
+    public void deleteFileAndMediaReference(DownloadInfo info) {
```
blundell: see the and in this method. That means internally it can look like this
```java
public void deleteFileAndMediaReference(DownloadInfo info) {
          deleteFile(info);
          deleteMediaReference(info);
}
```

FYI your method name should actually be `deleteMediaReferenceAndFile()`

If you had those two smaller methods, they could be easier to read with a *guard clause* & *prefer the positive*:

```java
public void deleteFileAndMediaReference(DownloadInfo info) {
     String mediaUri = info.mMediaProviderUri;
     if (TextUtils.isEmpty(mediaUri)) {
          return;
     }
     resolver.delete(Uri.parse(mediaUri), null, null);
}
```

library/src/main/java/com/novoda/downloadmanager/lib/DownloadDeleter.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+        if (info.mDestination != Downloads.Impl.DESTINATION_EXTERNAL && info.mFileName != null) {
+            Log.d("deleteDownloadLocked() deleting " + info.mFileName);
+            deleteFileIfExists(info.mFileName);
+        }
+        downloads.remove(info.mId);
+    }
+
+    private void deleteFileIfExists(String path) {
+        if (!TextUtils.isEmpty(path)) {
```
blundell: if isEmpty return

library/src/main/java/com/novoda/downloadmanager/lib/DownloadClientReadyChecker.java
ouchadam: automerged

library/src/main/java/com/novoda/downloadmanager/lib/DownloadDeleter.java
ouchadam: this code was all moved but I see your point

blundell: ah I see :+1:

PR #56, opened 2015-07-03
Auto-refresh list

amlcurran:
- List now refreshes automatically on a change
- Made list items smaller so you can see more

auto-refresh


ouchadam:
:100: definitely needed this.

Are we happy to keep it only for the extended demo?

xrigau:
:+1:

PR #60, opened 2015-07-03
Replace data structures for more appropriate ones

zegnus:
This PR replaces the following data structures:

- Batches from `HashSet` to `ArrayList` because the insertion on an ArrayList is faster and there is no usage of any of the HashSet specific implementation advantatges
- mActiveNotifs and cluster from `HashMap` to `SimpleArrayMap` because the length of the items added to the maps are not too extensive, and so the optimised SimpleArrayMap is recommended (further detail at https://medium.com/google-developers/developing-for-android-ii-bb9a51f8c8b9)

Comment at https://github.com/novoda/download-manager/compare/Replace-data-structures?expand=1#diff-ee31e2ad08f5842e25b4b0175c130855L181 solved

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadNotifier.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadNotifier.java
@@ -64,7 +64,7 @@ class DownloadNotifier {
      * @see #buildNotificationTag(DownloadBatch)
      */
 //    @GuardedBy("mActiveNotifs")
-    private final SimpleArrayMap<String, Long> mActiveNotifs = new SimpleArrayMap<>();
+    private final SimpleArrayMap<String, Long> activeNotifs = new SimpleArrayMap<>();
 
     /**
      * Current speed of active downloads, mapped from {@link DownloadInfo#batchId}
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/novoda/downloadmanager/lib/DownloadNotifier.java
```diff
@@ -36,11 +37,7 @@
 
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.HashMap;
```
Electryc: I like this red on the imports :+1:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadNotifier.java
```diff
@@ -423,13 +421,12 @@ private void setSecondaryNotificationText(NotificationCompat.Builder builder, No
         style.setSummaryText(description);
     }
 
-    private void removeStaleTagsThatWereNotRenewed(Map<String, Collection<DownloadBatch>> clustered) {
-        final Iterator<String> tags = mActiveNotifs.keySet().iterator();
-        while (tags.hasNext()) {
-            final String tag = tags.next();
+    private void removeStaleTagsThatWereNotRenewed(SimpleArrayMap<String, Collection<DownloadBatch>> clustered) {
+        for(int i = 0, size = activeNotifs.size(); i < size; i++) {
```
xrigau: missing space between for and (

amlcurran: we need prb

zegnus: @xrigau https://github.com/novoda/download-manager/commit/6d01a8551b6fc4e8d17f01a0f6cfc8abadfa5907

PR #62, opened 2015-07-06
Fix batch deletion bug

zegnus:
## Problem description ##

When deleting a particular batch, only the first download from the batch is actually deleted

## Solution implemented ##

We have spotted the bug on which the query for getting all the downloads was actually getting only the first download
- We query all the downloads for the deletion process
- We reuse query methods in order to minimise duplication and reuse code
- Split query from deletion in order to make the code simpler to understand

Paired with @ronocod

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
@@ -31,6 +31,7 @@ class BatchRepository {
     );
 
     private static final int PRIORITISED_STATUSES_SIZE = PRIORITISED_STATUSES.size();
+    private static final String SELECT_MARKED_FOR_DELETION = Batches.COLUMN_DELETED + " = 1";
 
     private final ContentResolver resolver;
     private final DownloadDeleter downloadDeleter;
@@ -166,10 +167,7 @@ class BatchRepository {
     }
 
     public void deleteMarkedBatchesFor(Collection<DownloadInfo> downloads) {
-        String selection = Downloads.Impl.Batches.COLUMN_DELETED + " = ?";
-        String[] selectionArgs = {"1"};
-
-        Cursor batchesCursor = resolver.query(Downloads.Impl.BATCH_CONTENT_URI, null, selection, selectionArgs, null);
+        Cursor batchesCursor = resolver.query(Downloads.Impl.BATCH_CONTENT_URI, null, SELECT_MARKED_FOR_DELETION, null, null);
         List<Long> batchIdsToDelete = new ArrayList<>();
         try {
             int idColumn = batchesCursor.getColumnIndexOrThrow(Downloads.Impl.Batches._ID);
--- a/library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
@@ -86,7 +86,8 @@ class BatchRepository {
         SparseIntArray statusCounts = new SparseIntArray(PRIORITISED_STATUSES_SIZE);
         try {
             String[] selectionArgs = {String.valueOf(batchId)};
-            cursor = resolver.query(ALL_DOWNLOADS_CONTENT_URI,
+            cursor = resolver.query(
+                    ALL_DOWNLOADS_CONTENT_URI,
                     null,
                     COLUMN_BATCH_ID + " = ?",
                     selectionArgs,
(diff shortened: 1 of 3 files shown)


library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
```diff
@@ -168,23 +162,42 @@ public DownloadBatch retrieveBatchBy(DownloadInfo downloadInfo) {
...
         }
 
-        if (!forDeletion.isEmpty()) {
-            deleteBatchesForIds(forDeletion, downloads);
+        return batches;
+    }
+
+    public void deleteMarkedBatchesFor(Collection<DownloadInfo> downloads) {
+        String selection = Downloads.Impl.Batches.COLUMN_DELETED + " = ?";
+        String[] selectionArgs = {"1"};
```
Electryc: Maybe this could be extracted as `MARKED_FOR_DELETION = new String[]{"1"}`

friedger: Could be even `Downloads.Impl.Batches.COLUMN_DELETED + " = 1"; `

zegnus: Will do

zegnus: @Electryc @friedger https://github.com/novoda/download-manager/commit/5d320807a18fe1ffd9c710e51911675921da7962

Electryc:
Separating query from modifier :+1:

Electryc:
LGTM :+1:

blundell:
wait

blundell:
dahm :)

blundell:
ah only 9 mins late!

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
blundell: I don't know why you combined these?

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
```diff
@@ -168,23 +163,39 @@ public DownloadBatch retrieveBatchBy(DownloadInfo downloadInfo) {
             batchesCursor.close();
         }
 
-        if (!forDeletion.isEmpty()) {
-            deleteBatchesForIds(forDeletion, downloads);
+        return batches;
+    }
+
+    public void deleteMarkedBatchesFor(Collection<DownloadInfo> downloads) {
+        Cursor batchesCursor = resolver.query(Downloads.Impl.BATCH_CONTENT_URI, null, SELECT_MARKED_FOR_DELETION, null, null);
```
blundell: http://developer.android.com/reference/android/content/ContentResolver.html#query(android.net.Uri, java.lang.String[], java.lang.String, java.lang.String[], java.lang.String)

Why aren't you using `query` like it should be used? With `selection` and `selectionArgs` like you did originally. What do you gain by combining them? Apart from not following the api :-(

ronocod: Is there any benefit in using `selectionArgs` when it can never change? There's only 1 SQL statement to cache in either scenario

zegnus: Yeah that's what I've implemented originally. I guess that @Electryc and @friedger saw that as a constant that is fixed for all calls.

I'm ok with both for this particular case, happy to revert it back if you prefer to follow the api

blundell: what Andrei was saying was right extract the constant, but not combine them - thats confusing.

@ronocod doing it this way works but is not actually in the api contract so is not guaranteed to always work like this

ronocod: :+1: we better stick to the interface then

Electryc: I was more inclined to extract the constant as I said earlier, but then taking a look at the resolver's `query` method in regards with the caching I went for merging it as it is. But now I see @blundell 's point about respecting the api contract so I think I might be better to comply to using the wildcard and `selectionArgs`.

friedger: The api contract does not say to use `?` for the right hand side, but if perfmatters: 

Use question mark parameter markers such as 'phone=?' instead of explicit values in the selection parameter, so that queries that differ only by those values will be recognized as the same for caching purposes.

Didn't know about this.
Does it make any change in performance though?

ronocod: In this case it won't affect performance because the value we're passing into the `?` is always the same

ronocod: We should use a projection for performance though, we only need `_ID` from the rows.

zegnus: @blundell @ronocod @friedger @Electryc Comments amended at https://github.com/novoda/download-manager/pull/66

PR #65, opened 2015-07-06
Fix wrong item deletion looping through a list of items

zegnus:
## Problem description ##

There is a `NullPointerException` at executing `tag.hashCode()` when `removeStaleTagsThatWereNotRenewed(...)`

## Solution implemented ##

- The deletion of an item through a manual iteration must be done in reverse order
- We have extracted the code related to displaying the notifications
- We have added tests around removing the staled notifications

Paired with @amlcurran

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadNotifier.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadNotifier.java
@@ -23,7 +23,6 @@ import android.content.Context;
 import android.content.res.Resources;
 import android.support.annotation.NonNull;
 import android.support.v4.util.SimpleArrayMap;
-import android.text.TextUtils;
 
 import com.novoda.notils.logger.simple.Log;
 
@@ -188,15 +187,6 @@ class DownloadNotifier {
         }
 
         return staleTags;
-    }
-
-    static CharSequence getDownloadTitle(BatchInfo batch) {
-        String title = batch.getTitle();
-        if (TextUtils.isEmpty(title)) {
-            return "unknown";
-        } else {
-            return title;
-        }
     }
 
     public void dumpSpeeds() {
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadNotifier.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadNotifier.java
@@ -20,7 +20,6 @@ import android.app.Notification;
 import android.app.NotificationManager;
 import android.app.PendingIntent;
 import android.content.Context;
-import android.content.res.Resources;
 import android.support.annotation.NonNull;
 import android.support.v4.util.SimpleArrayMap;
 
@@ -55,14 +54,7 @@ class DownloadNotifier {
 
     private final NotificationDisplayer notificationDisplayer;
 
-    public static DownloadNotifier newInstance(Context context, NotificationImageRetriever imageRetriever, Resources resources) {
-        NotificationManager notificationManager = (NotificationManager) context.getSystemService(Context.NOTIFICATION_SERVICE);
-        NotificationDisplayer notificationDisplayer = new NotificationDisplayer(context, notificationManager, imageRetriever, resources);
-
-        return new DownloadNotifier(context, notificationDisplayer);
-    }
-
-    DownloadNotifier(Context context, NotificationDisplayer notificationDisplayer) {
+    public DownloadNotifier(Context context, NotificationDisplayer notificationDisplayer) {
         this.context = context;
         this.notificationDisplayer = notificationDisplayer;
     }
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadNotifier.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadNotifier.java
@@ -422,7 +422,7 @@ class DownloadNotifier {
     }
 
     private void removeStaleTagsThatWereNotRenewed(SimpleArrayMap<String, Collection<DownloadBatch>> clustered) {
-        for (int i = 0, size = activeNotifs.size(); i < size; i++) {
+        for (int i = activeNotifs.size() - 1; i >= 0; i--) {
             String tag = activeNotifs.keyAt(i);
             if (!clustered.containsKey(tag)) {
                 mNotifManager.cancel(tag.hashCode());
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadNotifier.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadNotifier.java
@@ -19,26 +19,17 @@ package com.novoda.downloadmanager.lib;
 import android.app.Notification;
 import android.app.NotificationManager;
 import android.app.PendingIntent;
-import android.content.ContentUris;
 import android.content.Context;
-import android.content.Intent;
 import android.content.res.Resources;
-import android.graphics.Bitmap;
-import android.net.Uri;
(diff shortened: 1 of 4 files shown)


ronocod:
:+1: LGTM

amlcurran:
Any reason why your test is in your PR description but not in the code :smile:

zegnus:
Yes, because those are private methods :S

amlcurran:
[updateWith](https://github.com/novoda/download-manager/blob/Fix-NPE-removing-stale-notification-tags/library/src/main/java/com/novoda/downloadmanager/lib/DownloadNotifier.java#L108) is public and uses that method - did you try testing from there? It would be a good opportunity to split out some functionality into different classes too

library/src/main/java/com/novoda/downloadmanager/lib/DownloadNotifier.java
```diff
@@ -422,7 +422,7 @@ private void setSecondaryNotificationText(NotificationCompat.Builder builder, No
     }
 
     private void removeStaleTagsThatWereNotRenewed(SimpleArrayMap<String, Collection<DownloadBatch>> clustered) {
-        for (int i = 0, size = activeNotifs.size(); i < size; i++) {
+        for (int i = activeNotifs.size() - 1; i >= 0; i--) {
```
blundell: why did it need reversing? Could you have just said `i <= size` to fix it also ?

zegnus: It is because if you delete going forward, the next item is two items beyond because you just deleted the current item so the next one is at your current position.

But going backwards you always have a valid item when you delete the current one because you move to the left and that is the correct item at that position.

zegnus:
@amlcurran Yup exactly, but `updateWith` does three different things but all three methods call all instance variables of the class. It would be a bit artificial to separate the logic into an isolated class `StaleTagsRemover` maybe? :D

library/src/main/java/com/novoda/downloadmanager/lib/DownloadNotifier.java
blundell: oh I see, so you could go forwards and when you delete do `i--`

library/src/main/java/com/novoda/downloadmanager/lib/NotificationDisplayer.java
```diff
@@ -0,0 +1,314 @@
...
+import android.text.format.DateUtils;
+
+import com.novoda.downloadmanager.R;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+class NotificationDisplayer {
```
zegnus: No internal refactoring has done in this class, that could be done in future PRs

amlcurran:
:100: for the amazing test

amlcurran:
(and now we can start making people test the logic here :+1: )

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
```diff
@@ -150,7 +150,7 @@ public void onCreate() {
 
         downloadClientReadyChecker = getDownloadClientReadyChecker();
 
-        mNotifier = new DownloadNotifier(this, getNotificationImageRetriever(), getResources());
+        mNotifier = DownloadNotifier.newInstance(this, getNotificationImageRetriever(), getResources());
```
blundell: I don't think we're using this pattern in this project.

Can you just new everything up here to make it explicit what is happening

library/src/main/java/com/novoda/downloadmanager/lib/DownloadNotifier.java
```diff
@@ -181,257 +176,21 @@ private void addBatchToCluster(String tag, SimpleArrayMap<String, Collection<Dow
...
+                staleTags.add(tag.hashCode());
+                activeNotifications.removeAt(i);
             }
         }
+
+        return staleTags;
     }
 
-    private static CharSequence getDownloadTitle(BatchInfo batch) {
+    static CharSequence getDownloadTitle(BatchInfo batch) {
```
blundell: this can be moved to the other class right?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
amlcurran: :+1:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadNotifier.java
zegnus: yup!

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
zegnus: https://github.com/novoda/download-manager/commit/795a804f7217a001d53b45ec77c6d4b6cd8e24eb

library/src/main/java/com/novoda/downloadmanager/lib/DownloadNotifier.java
zegnus: https://github.com/novoda/download-manager/commit/094290abd5bd02937c3f7db4e88111cbe1890f68

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
```diff
@@ -150,7 +151,14 @@ public void onCreate() {
...
         downloadClientReadyChecker = getDownloadClientReadyChecker();
 
-        mNotifier = new DownloadNotifier(this, getNotificationImageRetriever(), getResources());
+        NotificationManager notificationManager = (NotificationManager) getSystemService(Context.NOTIFICATION_SERVICE);
+        NotificationDisplayer notificationDisplayer = new NotificationDisplayer(
+                this,
+                notificationManager,
+                getNotificationImageRetriever(),
+                getResources());
+
```
blundell: no ws needed :p

blundell:
:+1:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
zegnus: This is done here https://github.com/novoda/download-manager/blob/Fix-NPE-removing-stale-notification-tags/library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java#L165 as well and it looks better as well.

blundell: the creation of the `notificationDisplayer` is linked to the `new DownloadNotifier` so why is there whitespace in the middle?

zegnus: Because is a block of text like an `if` block. It looks cleaner with the new line in order to visually see the unity of the block and to whom it belongs

blundell: I disagree but yolo keep it if you like it

xrigau:
let's :shipit:

hal9002:
https://ci.novoda.com/job/download-manager-prb/3/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
zegnus: Oh I was about to change it now for you now, already merged. When I see it again I'll remove the line ;)

PR #66, opened 2015-07-06
Revert constant for query selection and project only the column needed

zegnus:
## Task description ##

This PR amends the comments at https://github.com/novoda/download-manager/pull/62#discussion_r33934601

- Use `selection` and `selectionArgs` instead of a constant value
- Project only the column that we need

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
@@ -32,9 +32,9 @@ class BatchRepository {
 
     private static final int PRIORITISED_STATUSES_SIZE = PRIORITISED_STATUSES.size();
 
-    private static final String[] PROJECT_BATCH_ID = {Downloads.Impl.Batches._ID};
-    private static final String SELECT_DELETED = Downloads.Impl.Batches.COLUMN_DELETED;
-    private static final String[] WHERE_MARKED_FOR_DELETION = {"1"};
+    private static final String[] PROJECT_BATCH_ID = {Batches._ID};
+    private static final String WHERE_DELETED_VALUE_IS = Batches.COLUMN_DELETED + " = ?";
+    private static final String[] MARKED_FOR_DELETION = {"1"};
 
     private final ContentResolver resolver;
     private final DownloadDeleter downloadDeleter;
@@ -170,7 +170,7 @@ class BatchRepository {
     }
 
     public void deleteMarkedBatchesFor(Collection<DownloadInfo> downloads) {
-        Cursor batchesCursor = resolver.query(Downloads.Impl.BATCH_CONTENT_URI, PROJECT_BATCH_ID, SELECT_DELETED, WHERE_MARKED_FOR_DELETION, null);
+        Cursor batchesCursor = resolver.query(Downloads.Impl.BATCH_CONTENT_URI, PROJECT_BATCH_ID, WHERE_DELETED_VALUE_IS, MARKED_FOR_DELETION, null);
         List<Long> batchIdsToDelete = new ArrayList<>();
         try {
             while (batchesCursor.moveToNext()) {
--- a/library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
@@ -32,7 +32,7 @@ class BatchRepository {
 
     private static final int PRIORITISED_STATUSES_SIZE = PRIORITISED_STATUSES.size();
 
-    private static final String[] PROJECTION_BATCH_ID = {Downloads.Impl.Batches._ID};
+    private static final String[] PROJECT_BATCH_ID = {Downloads.Impl.Batches._ID};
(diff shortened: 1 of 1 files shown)


friedger:
I still disagree that this is good use of the app because this give the impression of a query that can change, which we don't have.

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
```diff
@@ -167,13 +167,15 @@ public DownloadBatch retrieveBatchFor(DownloadInfo download) {
     }
 
     public void deleteMarkedBatchesFor(Collection<DownloadInfo> downloads) {
-        Cursor batchesCursor = resolver.query(Downloads.Impl.BATCH_CONTENT_URI, null, SELECT_MARKED_FOR_DELETION, null, null);
+        String[] projection = {Downloads.Impl.Batches._ID};
+        String selection = Batches.COLUMN_DELETED + " = ?";
+        String[] selectionArgs = {MARKED_FOR_DELETION};
```
blundell: so Andrei originally said do this
`private static final String[] SELECT_DELETED = {Downloads.Impl.Batches._ID};`
`private static final String[] WHERE_MARKED_FOR_DELETION = {"1"};`

blundell: `Cursor batchesCursor = resolver.query(Downloads.Impl.BATCH_CONTENT_URI, projection, SELECT_DELETED, WHERE_MARKED_FOR_DELETION, null);`

zegnus: https://github.com/novoda/download-manager/commit/72018f648087e820cbfd558675ef521fb25910e1

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
```diff
@@ -31,7 +31,10 @@
     );
 
     private static final int PRIORITISED_STATUSES_SIZE = PRIORITISED_STATUSES.size();
-    private static final String SELECT_MARKED_FOR_DELETION = Batches.COLUMN_DELETED + " = 1";
+
+    private static final String[] PROJECTION_BATCH_ID = {Downloads.Impl.Batches._ID};
```
blundell: `PROJECT_BATCH_ID` to match `SELECT_`  (vs projection & selection)

zegnus: https://github.com/novoda/download-manager/commit/9a8d06c7ed6b8c83d78b58dbafee2baef4d34be1

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
```diff
@@ -167,13 +170,11 @@ public DownloadBatch retrieveBatchFor(DownloadInfo download) {
     }
 
     public void deleteMarkedBatchesFor(Collection<DownloadInfo> downloads) {
-        Cursor batchesCursor = resolver.query(Downloads.Impl.BATCH_CONTENT_URI, null, SELECT_MARKED_FOR_DELETION, null, null);
+        Cursor batchesCursor = resolver.query(Downloads.Impl.BATCH_CONTENT_URI, PROJECT_BATCH_ID, SELECT_DELETED, WHERE_MARKED_FOR_DELETION, null);
```
friedger: :x: You tests should fail. The `?` is missing.

Furthermore, the third argument is the where clause not the fourth.
WHERE_MARKED_FOR_DELETION should be DELETION_VALUE or DELETION_FLAG in contrast to KEEP_VALUE or KEEP_FLAG

If you want to go with the "selection argument" then maybe something like:
`Cursor batchesCursor = resolver.query(Downloads.Impl.BATCH_CONTENT_URI, PROJECT_BATCH_ID, WHERE_DELETE_VALUE_IS, DELETION_FLAG, null);`

friedger: Or use 
`Cursor batchesCursor = resolver.query(Downloads.Impl.BATCH_CONTENT_URI, PROJECT_BATCH_ID, WHERE_MARKED_FOR_DELETION, null, null);`

hal9002:
https://ci.novoda.com/job/download-manager-prb/7/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
zegnus: @friedger You are completely right, fixed all values and names https://github.com/novoda/download-manager/commit/86d0ef542b1e06684579a8e4201ba3428b942c5b

I prefer to use `MARKED_FOR_DELETION` instead of `DELETION_FLAG` as it might be more expressive to use the former and we don't have anywhere in the code right now for a `keep` value :)

Thanks for your eye!

friedger: :+1: I'd prefer the second option with the args, but LGTM.

PR #67, opened 2015-07-06
Fix error when no batch ID is provided for the Query filter

ronocod:
Fixes the error:
```
E/SQLiteLog﹕ (1) near ")": syntax error
```
when no batch IDs have been provided to a `Query`.

Also simplifies the SQL generation and adds test around this functionality.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/Query.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/Query.java
@@ -5,6 +5,7 @@ import android.database.Cursor;
 import android.net.Uri;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 /**
@@ -141,7 +142,7 @@ public class Query {
         if (ids == null) {
             return null;
         }
-        return DownloadManager.longArrayToStringArray(ids);
+        return longArrayToStringArray(ids);
     }
 
     private void filterByDownloadIds(List<String> selectionParts) {
@@ -152,7 +153,7 @@ public class Query {
     }
 
     private void filterByBatchIds(List<String> selectionParts) {
-        if (batchIds == null) {
+        if (batchIds == null || batchIds.length == 0) {
             return;
         }
         selectionParts.add(getWhereClauseForBatchIds(batchIds));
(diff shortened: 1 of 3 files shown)


library/src/main/java/com/novoda/downloadmanager/lib/Query.java
```diff
@@ -228,6 +219,14 @@ private String joinStrings(String joiner, Iterable<String> parts) {
         return builder.toString();
     }
 
+    private static String[] longArrayToStringArray(long[] longs) {
```
blundell: moving from `DownloadManager.longArrayToStringArray` makes sense :+1:  can the old one be deleted? :dart:

library/src/main/java/com/novoda/downloadmanager/lib/Query.java
```diff
@@ -152,7 +153,7 @@ private void filterByDownloadIds(List<String> selectionParts) {
     }
 
     private void filterByBatchIds(List<String> selectionParts) {
-        if (batchIds == null) {
+        if (batchIds == null || batchIds.length == 0) {
```
blundell: so this is the fix?

library/src/main/java/com/novoda/downloadmanager/lib/Query.java
ronocod: The old one is used in `DownloadManager` itself

library/src/main/java/com/novoda/downloadmanager/lib/Query.java
ronocod: Yeah, it won't append any clause if there are no batch IDs

library/src/main/java/com/novoda/downloadmanager/lib/Query.java
blundell: make it private if you can :+1:

ronocod: :+1: boyscouted another one too https://github.com/novoda/download-manager/commit/099b33ae405e0dacb6f790aca57c51cef169bed5

xrigau:
thanks @ronocod I've been wondering about this error for a few days now haha!

PR #71, opened 2015-07-07
Incorrect batch id passed to the rule checker

ouchadam:
We weren't checking that the first batch in the retrieved list matched the dowload info batch id, this caused the `isAllowedToDownload` to always be called with the first downloaded batch!

Fixed by enforcing the batch ids to be matched instead of just using the first list entry

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
@@ -127,7 +127,14 @@ class BatchRepository {
     public DownloadBatch retrieveBatchFor(DownloadInfo download) {
         Collection<DownloadInfo> downloads = Collections.singletonList(download);
         List<DownloadBatch> batches = retrieveBatchesFor(downloads);
-        return batches.isEmpty() ? DownloadBatch.DELETED : batches.get(0);
+
+        for (DownloadBatch batch : batches) {
+            if (batch.getBatchId() == download.batchId) {
+                return batch;
+            }
+        }
+
+        return DownloadBatch.DELETED;
     }
 
     public List<DownloadBatch> retrieveBatchesFor(Collection<DownloadInfo> downloads) {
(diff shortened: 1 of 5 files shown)


hal9002:
https://ci.novoda.com/job/download-manager-prb/18/
Test PASSed.

blundell:
public method, looks easy to test :+1:

stefanhoth:
LGTM!
image

ouchadam:
can do the test here or in another PR, don't mind!

xrigau:
:+1:

amlcurran:
test here I think, then you can verify you fixed it

xrigau:
what @amlcurran said :+1:

hal9002:
https://ci.novoda.com/job/download-manager-prb/20/
Test PASSed.

hal9002:
https://ci.novoda.com/job/download-manager-prb/21/
Test PASSed.

hal9002:
https://ci.novoda.com/job/download-manager-prb/22/
Test PASSed.

hal9002:
https://ci.novoda.com/job/download-manager-prb/23/
Test PASSed.

amlcurran:
LGTM!

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
```diff
@@ -38,22 +39,30 @@
 
     private final ContentResolver resolver;
     private final DownloadDeleter downloadDeleter;
+    private final Uri batchContentUri;
+    private final Uri allDownloadsContentUri;
 
-    public BatchRepository(ContentResolver resolver, DownloadDeleter downloadDeleter) {
+    public static BatchRepository newInstance(ContentResolver resolver, DownloadDeleter downloadDeleter) {
```
blundell: we aren't using this pattern in this project. Granted I can see how here it stops the duplication of using the constants in multiple places. So i'm happy either way

ouchadam: yep it's purely for the constants which have only been made fields because of the android testing framework and Uris.

Seemed a little unfair to enforce the full constructor, the other option is two provide two constructors.....

amlcurran: if this is conveniently hiding duplication then it is conveniently hiding a code smell

I think we should keep the constructor so it is big and ugly and obvious we need to refactor

blundell: duplication in creation, the code smell being it should only be created in one place?

amlcurran: I guess the smell is that Downloads.Impl does work in generating its static finals and this makes anything that touches that class difficult to unit test :/

PR #72, opened 2015-07-07
Add checkstyle, PMD, findbugs

amlcurran:
Add our standard set of code analysis to the project. We've also replaced the previous checkstyle implementation with the official plug in, rather than our own custom implementation


hal9002:
https://ci.novoda.com/job/download-manager-prb/24/
Test PASSed.

xrigau:
don't forget we'll have to add the reports file paths to jenkins!

amlcurran:
:+1: good point

code-quality/android-code-quality.gradle
```diff
@@ -0,0 +1,80 @@
...
+def sourceSetMain = android.sourceSets.main.java.srcDirs
+def rulesDir = new File("$project.rootDir/code-quality", 'rules')
+def ignoreFail = true // sad panda
+
+def includeSrcPatternApplication = 'com/novoda/downloadmanager/**/*.java'
+def excludeSrcPatternRClass = 'com/novoda/downloadmanager/**/R.java'
+def excludeSrcPatternGeneratedFiles = '**/gen/**'
+
+apply plugin: 'checkstyle'
+task checkstyleMain(type: Checkstyle) {
```
blundell: can we still have these as 3 separate files so we can easily distinguish them and change them separately?

amlcurran: we can indeed

hal9002:
https://ci.novoda.com/job/download-manager-prb/26/
Test PASSed.

code-quality/android-code-quality.gradle
amlcurran: https://github.com/novoda/download-manager/commit/b33971516c937050b090c8a5fdfe5f446969b019

blundell:
LGTM, so we can use this because we don't have a /core module is that correct?or is it because of some other change in gradle versions

amlcurran:
not sure, this configuration was taken from another project and this from their mobile module, but they have a similar config in core as well

xrigau:
@blundell this is how cool kids do it now, use the standard gradle plugins for both java and android modules (but android need some more setup) :space_invader:

xrigau:
this is ready to :ship: #justsayin

ouchadam:
let's do it

PR #78, opened 2015-07-08
Less isAllowToDownload callbacks

ouchadam:
We found that `isAllowedToDownload` was being called for all queued and running downloads, on **EVERY** download table change. It kind of made sense for running downloads but wasn't really necessary...

This PR forces the check to only occur if a download is ready and hasn't been submitted to the downloads exectutor


hal9002:
https://ci.novoda.com/job/download-manager-prb/45/
Test PASSed.

xrigau:
:+1: seems legit!

blundell:
this changes the amount of times `startDownloadIfNotActive` so it affects more than the `isAllowedToDownload()` call - is it ok to not start the download if not active in these situations?

PR #59, opened 2015-07-03
Limit client internals knowledge

ouchadam:
Only the `Download` is given to clients via the `isAllowedToDownload` method which contains a limited amount of information (instead of the collection of `DownloadInfo`s)

PR #79, opened 2015-07-08
Query for batches

Electryc:
Introduces the ability to query for batches. The querying is done through `downloadManager.queryBatch(BatchQuery query)`. 

`BatchQuery` is the class with which you can sort/filter the query. 

The following constants were exposed:
- `DownloadManager.BATCH_CONTENT_URI` has been exposed in order clients to be able to watch the batches table. 
- `DownloadManager.COLUMN_BATCH_TOTAL_SIZE_BYTES`
- `DownloadManager.COLUMN_BATCH_CURRENT_SIZE_BYTES`

Example of usage (for querying batches that have failed or pending):
```java
BatchQuery query = new BatchQuery.Builder()
        .withStatusFilter(
                DownloadManager.STATUS_FAILED
                        | DownloadManager.STATUS_PENDING)
        .build();
Cursor batchesCursor = downloadManager.queryBatch(query);
```

The extended demo was updated with an additional activity `ShowBatchesActivity` that reflects the new functionality. 
query_batches


hal9002:
https://ci.novoda.com/job/download-manager-prb/46/
Test PASSed.

hal9002:
https://ci.novoda.com/job/download-manager-prb/47/
Test PASSed.

friedger:
Isn't Download the new public class?
Am 08.07.2015 16:34 schrieb "hal9002" <notifications@github.com>:

Electryc:
@friedger I don't really get what you mean.

library/src/test/java/com/novoda/downloadmanager/lib/CriteriaTest.java
```diff
@@ -0,0 +1,180 @@
...
+
+        CriteriaAssert.assertThat(criteria)
+                .hasSelection(SELECTION_QUERY + OR + "(" + ANOTHER_SELECTION_COLUMN_QUERY + AND + SELECTION_QUERY + ")")
+                .hasArguments(new String[]{ARGUMENT, ARGUMENT, ARGUMENT});
+    }
+
+    @Test
+    public void givenMultipleSelectionsWithMultipleInnerCriteriaWhenBuildingThenTheSelectionQueryIsCorrect() {
+        Criteria criteria = new Criteria.Builder()
+                .withInnerCriteria(
```
amlcurran: what's an inner criteria? what do we use it for?

library/src/main/java/com/novoda/downloadmanager/lib/BatchQuery.java
```diff
@@ -0,0 +1,244 @@
...
+public class BatchQuery {
+    public static final BatchQuery ALL = new BatchQuery(null, null, null);
+    private static final int LOW_END_FAILED_STATUS_CODE = 400;
+    private static final int HIGH_END_FAILED_STATUS_CODE = 600;
+
+    private final String selection;
+    private final String sortOrder;
+    private final String[] selectionArguments;
+
+    BatchQuery(String selection, String sortOrder, String[] selectionArguments) {
```
friedger: this is an unusual ordering of parameters, it should be `selection, selectionArguments, sortOrder` as in ContentProviders

library/src/main/java/com/novoda/downloadmanager/lib/BatchQuery.java
```diff
@@ -0,0 +1,244 @@
...
+         *                    {@link DownloadManager#STATUS_PENDING},
+         *                    {@link DownloadManager#STATUS_SUCCESSFUL},
+         *                    {@link DownloadManager#STATUS_RUNNING}
+         *                    <p/>
+         *                    e.g. withStatusFilter(DownloadManager.STATUS_FAILED | DownloadManager.STATUS_PENDING)
+         * @return {@link BatchQuery.Builder}
+         */
+        public Builder withStatusFilter(@Status int statusFlags) {
+            this.criteriaStatusBuilder = new Criteria.Builder();
+            List<Integer> statuses = buildStatusesFrom(statusFlags);
```
friedger: This looks overcomplicated. Instead of looping of a switch statement you could do the if and bit-wise check in here.

friedger: and then do something like builder.or(list of criteria)

library/src/main/java/com/novoda/downloadmanager/lib/BatchQuery.java
friedger: oh, I see, similar to Criteria, then swap there as well and have `sortOrder` first at both places.

library/src/test/java/com/novoda/downloadmanager/lib/BatchQueryTest.java
```diff
@@ -0,0 +1,334 @@
...
+                Downloads.Impl.STATUS_QUEUED_FOR_WIFI,
+                Downloads.Impl.STATUS_SUCCESS,
+                400,
+                600
+        };
+        assertThatSelectionArgumentAreEqualTo(query.getSelectionArguments(), expectedArguments);
+    }
+
+    @Test
+    public void givenMultipleSettingStatusFilterMultipleTimesWhenTheQueryIsBuiltThenOnlyTheLastFilterIsTakenIntoConsideration() {
```
friedger: one `Multiple` too often?

library/src/test/java/com/novoda/downloadmanager/lib/BatchQueryTest.java
```diff
@@ -0,0 +1,334 @@
...
+        Integer[] expectedArguments = {
+                400,
+                600
+        };
+        assertThatSelectionArgumentAreEqualTo(query.getSelectionArguments(), expectedArguments);
+
+    }
+
+    @Test
+    public void givenMultipleSettingIdFilterMultipleTimesWhenTheQueryIsBuiltThenOnlyTheLastFilterIsTakenIntoConsideration() {
```
friedger: same here?

library/src/test/java/com/novoda/downloadmanager/lib/CriteriaTest.java
friedger: a selection with brackets around it
https://github.com/novoda/download-manager/pull/79/files#diff-9dee9901928c738b64cad1093dca5f25R55

friedger: could be renamed to `inBrackets()`

library/src/main/java/com/novoda/downloadmanager/lib/Criteria.java
```diff
@@ -0,0 +1,177 @@
+package com.novoda.downloadmanager.lib;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+public class Criteria {
```
friedger: Could be package local

library/src/main/java/com/novoda/downloadmanager/lib/Criteria.java
```diff
@@ -0,0 +1,177 @@
...
+
+    public static final Criteria ALL = new Criteria(null, null);
+
+    private final String selection;
+    private final String sort;
+    private final String[] selectionArguments;
+
+    Criteria(String selection, String sort, String... selectionArguments) {
```
friedger: and then this private

library/src/main/java/com/novoda/downloadmanager/lib/Criteria.java
```diff
@@ -0,0 +1,177 @@
...
+
+    public String[] getSelectionArguments() {
+        return Arrays.copyOf(selectionArguments, selectionArguments.length);
+    }
+
+    public String getSort() {
+        return sort;
+    }
+
+    public static class Builder {
```
friedger: Here you are using the `with...()` methods for appending values while in the `BatchQuery` the values are set/overwritten.

`new BatchQuery.Builder().withId(1).withId(2)` has a different meaning than `new Criteria.Builder().withSelection(A, wildcard).withArgument(1).withSelection(B, wildcard).withArgument(2)` 

better just use `selection()` and `argument()` similar to `or`, ..

Electryc: The idea was to use the `Criteria` as a helper to construct the basic forms of a query in terms of selection, arguments, sort order. 

The `BatchQuery` is our public API where we define different filtering options. I don't know if it should be in sync with `Criteria`. 

Maybe in order to keep some consistency, the `Query` could use a similar approach, that of using the `Criteria` to construct the actual selections, and use a `repository` to do the actual querying as, at the moment, the `query` does too much (constructs the query and also runs it).

friedger: I didn't know about `Query`.

>The BatchQuery is our public API where we define different filtering options. I don't know if it should be in sync with Criteria.

The easier to learn our APIs the better to maintain for us. This is already a lot to learn for a few sql queries. I prefer the approach without `Criteria` to build queries because it is less API.

library/src/main/java/com/novoda/downloadmanager/lib/BatchQuery.java
Electryc: Well the bit-wise check is done in `buildStatusesFrom(statusFlags)`. Initially I was doing only the bit-wise check without the for loop, but then there was no way to know when to put and not to put `or` between them (ending with an additional `OR` at the end). Then I moved to not using the flags but an `int[]` for the filtering, but then we loose the nicer looks of the flags + plus the restriction that we can apply with the `@IntDef` annotation (see https://github.com/novoda/download-manager/commit/72265ca176b95473cc3e7fd48612f60c46405151#diff-5fabc9dc335ace9af378b7c1c777a861R43)

Here I'm just getting the list of statuses out of the flags in order to construct the query and to correctly place the `OR` between them. Do you think https://github.com/novoda/download-manager/commit/c7b58607bc6950c146b6d4a0322889762a5543b3 looks better?

hal9002:
https://ci.novoda.com/job/download-manager-prb/63/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/BatchQuery.java
Electryc: :+1: 92b797b

hal9002:
https://ci.novoda.com/job/download-manager-prb/64/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/Criteria.java
Electryc: :+1: ca5b7ba

library/src/main/java/com/novoda/downloadmanager/lib/Criteria.java
Electryc: And the class `final` then?

library/src/test/java/com/novoda/downloadmanager/lib/BatchQueryTest.java
Electryc: :+1: lost in `multiples` :smile_cat:

hal9002:
https://ci.novoda.com/job/download-manager-prb/65/
Test PASSed.

library/src/test/java/com/novoda/downloadmanager/lib/CriteriaTest.java
Electryc: @amlcurran as @friedger mentions the `innerCritera` is just an inner selection as in: 
`SELECT * from TABLE WHERE column1=1 OR (colum2=2 AND column3=3)`

here `(colum2=2 AND column3=3)` is an _inner criteria_ as it is contained within another criteria. This enables us to favour the priority of the criteria evaluation as the inner one will be evaluated first. 

I don't know what to say about `inBrackets()`, I wanted to emphasise the composable nature of the criterias (being able to construct one within another).

hal9002:
https://ci.novoda.com/job/download-manager-prb/66/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/BatchQuery.java
```diff
@@ -0,0 +1,223 @@
...
+         *                    {@link DownloadManager#STATUS_PENDING},
+         *                    {@link DownloadManager#STATUS_SUCCESSFUL},
+         *                    {@link DownloadManager#STATUS_RUNNING}
+         *                    <p/>
+         *                    e.g. withStatusFilter(DownloadManager.STATUS_FAILED | DownloadManager.STATUS_PENDING)
+         * @return {@link BatchQuery.Builder}
+         */
+        public Builder withStatusFilter(@Status int statusFlags) {
+            this.criteriaStatusBuilder = new Criteria.Builder()
+                    .or(buildCriteriaListFrom(statusFlags));
```
Electryc: Not completely sure about this though, related discussion [here](https://github.com/novoda/download-manager/pull/79#discussion_r34183529)

library/src/test/java/com/novoda/downloadmanager/lib/CriteriaTest.java
friedger: @Electryc I don't care about the naming but it should be easy to understand what it can be used for.

library/src/main/java/com/novoda/downloadmanager/lib/BatchQuery.java
friedger: maybe `oneOf(List<Criteria>)` makes is less weird.

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/batches/Batch.java
```diff
@@ -0,0 +1,51 @@
+package com.novoda.downloadmanager.demo.extended.batches;
+
+import com.novoda.downloadmanager.lib.DownloadManager;
+
+public class Batch {
```
ouchadam: :x: this should use the `Download` object

ouchadam: ah we're in the demo, ignore that!

friedger: @ouchadam I am also confused about the naming of Batches and Downloads.

library/src/main/java/com/novoda/downloadmanager/lib/BatchQuery.java
```diff
@@ -0,0 +1,223 @@
...
+            builder = new Criteria.Builder();
+        }
+
+        /**
+         * Filter by batch id
+         *
+         * @param id id of the batch
+         * @return {@link BatchQuery.Builder}
+         */
+        public Builder withId(long id) {
```
ouchadam: how come this creates a new builder?

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/batches/Batch.java
Electryc: They're just local model objects used to encapsulate the data from the db. One major difference between them is that `Download` has the path to the individual file as opposed to the `Batch` which doesn't.

library/src/main/java/com/novoda/downloadmanager/lib/BatchQuery.java
Electryc: The below `criteriaIdBuilder` is created every time in order to reset the id to the last one if multiple subsequent calls to `withId()` are made.

library/src/test/java/com/novoda/downloadmanager/lib/CriteriaTest.java
amlcurran: this seems like new functionality, did we add it because we're using it?

Electryc: https://github.com/novoda/download-manager/pull/79/files#diff-5fabc9dc335ace9af378b7c1c777a861R179 here we are using it. That `AND` needs to be inner (or in brackets) in order to take precedence in evaluation in between the `OR`s.

amlcurran: OK, was making sure this wasn't YAGNI as it seems to have added a ton of complexity but we do use it :+1:

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/batches/Batch.java
friedger: How does it go together with @ouchadam 's statement https://github.com/novoda/download-manager/pull/59#discussion_r33864682 
Could we update the README? Different PR though.

friedger: Good for me for now.

Electryc: Well this is only a model in the demo, not something that the library exposes.

blundell: just add javadoc to the classes with your comments about the difference Andrei , javadoc in demo;s is cool

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/batches/BatchAdapter.java
```diff
@@ -0,0 +1,62 @@
...
+    public long getItemId(int position) {
+        return position;
+    }
+
+    @Override
+    public View getView(int position, View convertView, ViewGroup parent) {
+        View view = View.inflate(parent.getContext(), R.layout.list_item_batch, null);
+
+        final Batch batch = getItem(position);
+        TextView idTextView = Views.findById(view, R.id.batch_id_text);
```
blundell: we shouldn't really use this lookup in demo's, use the normal view.findViewById

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
```diff
@@ -489,6 +501,16 @@ public Cursor query(Query query) {
         return new CursorTranslator(underlyingCursor, Downloads.Impl.DOWNLOADS_BY_BATCH_URI);
     }
 
+    public Cursor queryBatch(BatchQuery query) {
```
blundell: I would just call it `query` the typing is enough to see you are doing a batch query

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/batches/BatchAdapter.java
Electryc: :+1: I saw it in other places so I thought we can use it, but yeah you're right

Electryc: d2be097

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
Electryc: :+1: 980254d

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/batches/Batch.java
Electryc: :+1: javadoc'd ff977ed

hal9002:
https://ci.novoda.com/job/download-manager-prb/72/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/BatchQuery.java
Electryc: Renamed `joinWithOr(List<Criteria>)` 477c46e

Electryc:
@blundell @friedger :eyes: ?

blundell:
changes from my comments :+1:

amlcurran:
@Electryc might be worth taking a look at my updated sort #86

hal9002:
https://ci.novoda.com/job/download-manager-prb/78/
Test FAILed.

Electryc:
:+1: @amlcurran 924c88e

hal9002:
https://ci.novoda.com/job/download-manager-prb/79/
Test FAILed.

hal9002:
https://ci.novoda.com/job/download-manager-prb/82/
Test PASSed.

blundell:
@amlcurran you happy?

amlcurran:
yarp

PR #80, opened 2015-07-08
Adds back in-memory cache of objects

amlcurran:
This broke pause and resume because each FileDownloadInfo contains the thread that is running for that download :cry: So for now we don't create a new FileDownloadInfo if we already have an in-memory representation of it.

Ideally we'll move the threading out of this object, but at a later date


hal9002:
https://ci.novoda.com/job/download-manager-prb/48/
Test PASSed.

blundell:
@ouchadam this vs #82 ?

ouchadam:
This was the back up plan in case we weren't able to rely on the database. If we do want to keep the cache we'll need to prune it for old ids

PR #84, opened 2015-07-08
Less isAllowToDownload callbacks

ouchadam:
We found that isAllowedToDownload was being called for all queued and running downloads, on EVERY download table change. It kind of made sense for running downloads but wasn't really necessary...

This PR forces the check to only occur if a download hasn't been set to submitted or in progress before checking if that download is "ready to download"


hal9002:
https://ci.novoda.com/job/download-manager-prb/62/
Test PASSed.

blundell:
Are failed downloads affected by this? i.e. when it's failed does it hit this code

ouchadam:
Failed downloads will be caught by this https://github.com/novoda/download-manager/blob/master/library/src/main/java/com/novoda/downloadmanager/lib/FileDownloadInfo.java#L268

blundell:
which is before this point

ouchadam:
yep, it's part of the `info.isReadyToDownload()`

PR #86, opened 2015-07-09
Sort tweaks

amlcurran:
Sorting wasn't quite correct - failed items were mixed in with queued items, and the order was undefined inside one of the status groups. 

Now, we place unknown status codes at the bottom (we can iterate later when we come to understand the error codes more). Also, we sort within the groups by ID, which means that the first item below the currently downloading item will be the next to attempt a download.


hal9002:
https://ci.novoda.com/job/download-manager-prb/74/
Test FAILed.

amlcurran:
man who thought a prb was a good idea

hal9002:
https://ci.novoda.com/job/download-manager-prb/75/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/Query.java
```diff
@@ -31,14 +31,14 @@
      */
     public static final int ORDER_DESCENDING = 2;
 
-    private static final String ORDER_BY_LIVENESS = String.format("CASE batch_status "
+    private static final String ORDER_BY_LIVENESS = String.format("CASE status "
```
Electryc: Why the change from `batch_status` to `status` ? 
Should this be then  [`Downloads.Impl.COLUMN_STATUS`](https://github.com/novoda/download-manager/blob/master/library/src/main/java/com/novoda/downloadmanager/lib/Downloads.java#L204) ?

amlcurran: not entirely sure why I did this change, but given this is querying for downloads rather than batches, it is probably more proper to have it as download status. But yeah, will use the constant :+1:

hal9002:
https://ci.novoda.com/job/download-manager-prb/76/
Test PASSed.

amlcurran:
closing whilst deconflicting

hal9002:
https://ci.novoda.com/job/download-manager-prb/77/
Test PASSed.

zegnus:
The problem was the missing `_id ASC`?

amlcurran:
yeah and also making any uncovered statuses resolve to 6 rather than 2

zegnus:
:+1: LGTM

ronocod:
LGTM :boat:

library/src/main/java/com/novoda/downloadmanager/lib/Query.java
```diff
@@ -31,14 +31,15 @@
...
-                    + "END",
+    private static final String ORDER_BY_LIVENESS = String.format("CASE %1$s "
+                    + "WHEN %2$d THEN 1 "
+                    + "WHEN %3$d THEN 2 "
+                    + "WHEN %4$d THEN 3 "
+                    + "WHEN %5$d THEN 4 "
+                    + "WHEN %6$d THEN 5 "
+                    + "ELSE 6 "
+                    + "END, _id ASC",
+            DownloadContract.Downloads.COLUMN_STATUS,
```
Electryc: :+1:

PR #87, opened 2015-07-09
Fix current size for batches

Electryc:
We've noticed that the current size for the batches was not computed anymore. This is due to the fact that `batchRepository.updateCurrentSize(downloadInfo.getBatchId())` was executed only when there are no total bytes on the batch (notice the `if`) which was wrong - the current size should always be computed. 

Pulls updating the current size out of the check and replaces the hardcoded check with a method on `downloadInfo`.


hal9002:
https://ci.novoda.com/job/download-manager-prb/80/
Test PASSed.

hal9002:
https://ci.novoda.com/job/download-manager-prb/81/
Test PASSed.

amlcurran:
whoops! Gonna let @Electryc or @ronocod release this time though ;)

ouchadam:
:100:

xrigau:
:+1:

PR #89, opened 2015-07-09
Only run can download checks against the current download

ouchadam:
`isAllowedToDownload` was still being called on queued downloads whilst the download table changed! But it also ran the check for each item within a batch.

This PR introduced refactorings around the `canDownload` checks to be able to use them on a batch level. We also now query for the batches by their _id ordering to make sure the most recently added item is checked first.

- Adds a top level `isActive` flag from all the batches instead of iterating through, this allows us to only ask if the client is ready if a download is currently not in progress (we had this before on each iteration but the ordering meant we were out of sync).

- extracts all the `downloadManagerIsReady` logic into the DownloadReadyChecker instead of living on the `FileDownloadInfo` and separates out the networking logic.

- moves the download deletion and next retry time to the batch level

- big ol' refactoring of the FileDownloadInfo to pull out as much behaviour and dependencies as possible #92


hal9002:
https://ci.novoda.com/job/download-manager-prb/86/
Test PASSed.

hal9002:
https://ci.novoda.com/job/download-manager-prb/87/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/NetworkChecker.java
```diff
@@ -0,0 +1,60 @@
...
+        return checkIsNetworkTypeAllowed(downloadInfo, info.getType());
+    }
+
+    /**
+     * Check if this download can proceed over the given network type.
+     *
+     * @param networkType a constant from ConnectivityManager.TYPE_*.
+     * @return one of the NETWORK_* constants
+     */
+    private FileDownloadInfo.NetworkState checkIsNetworkTypeAllowed(FileDownloadInfo downloadInfo, int networkType) {
```
ouchadam: this probably doesn't need to exist any more since we have the client side checks

hal9002:
https://ci.novoda.com/job/download-manager-prb/88/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/FileDownloadInfo.java
```diff
@@ -267,90 +275,6 @@ private long restartTime(long now) {
     }
 
     /**
-     * Returns whether this download should be enqueued.
-     */
-    private boolean isDownloadManagerReadyToDownload() {
```
blundell: move this makes sense :+1:

ouchadam: unfortunately it had a lot of baggage

hal9002:
https://ci.novoda.com/job/download-manager-prb/90/
Test PASSed.

hal9002:
https://ci.novoda.com/job/download-manager-prb/93/
Test PASSed.

hal9002:
https://ci.novoda.com/job/download-manager-prb/94/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadBatch.java
```diff
@@ -50,6 +50,35 @@ public boolean isDeleted() {
...
+    /**
+     * Return time when this download will be ready for its next action, in
+     * milliseconds after given time.
+     *
+     * @return If {@code 0}, download is ready to proceed immediately. If
+     * {@link Long#MAX_VALUE}, then download has no future actions.
+     */
+    public long nextActionMillis(long now, long nextRetryTimeMillis) {
+        for (FileDownloadInfo info : downloads) {
+            long foo = getNextActionMillisFor(now, info);
```
Electryc: I see you!

ouchadam: arghhhhhhhh busted

hal9002:
https://ci.novoda.com/job/download-manager-prb/95/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
```diff
@@ -34,6 +34,7 @@
     private static final String[] PROJECT_BATCH_ID = {DownloadContract.Batches._ID};
     private static final String WHERE_DELETED_VALUE_IS = DownloadContract.Batches.COLUMN_DELETED + " = ?";
     private static final String[] MARKED_FOR_DELETION = {"1"};
+    private static final String SORT_BY_ID_DESCENDING = DownloadContract.Batches._ID + " DESC";
```
xrigau: why descending?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadBatch.java
```diff
@@ -50,6 +50,35 @@ public boolean isDeleted() {
         return this == DELETED;
     }
 
+    /**
+     * Return time when this download will be ready for its next action, in
```
xrigau: what does `action` mean? retry? what other actions exist?

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
ouchadam: because we want the newest download first

library/src/main/java/com/novoda/downloadmanager/lib/DownloadBatch.java
ouchadam: I'm not 100% sure, the main service loops sets alarm manager wake ups based on this retry time amount

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
xrigau: :+1:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadReadyChecker.java
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * Returns whether this download should be enqueued.
+     */
+    private boolean isDownloadManagerReadyToDownload(FileDownloadInfo downloadInfo) {
+        if (downloadInfo.getControl() == DownloadsControl.CONTROL_PAUSED) {
+            // the download is paused, so it's not going to start
+            return false;
+        }
+        switch (downloadInfo.getStatus()) {
+            case 0: // status hasn't been initialized yet, this is a new download
```
xrigau: could be extracted as `NOT_INITIALISED`

xrigau:
:+1: seems legit but would love another pair of :eyes: before hitting the button

library/src/main/java/com/novoda/downloadmanager/lib/DownloadBatch.java
Electryc: As far as I could see the name of `action` is a little bit misleading, this returns the time at which an alarm will restart the `DownloadService`.

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
amlcurran: do we? don't we want the earliest first?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
```diff
@@ -317,28 +324,26 @@ private boolean updateLocked() {
 
         List<DownloadBatch> downloadBatches = batchRepository.retrieveBatchesFor(allDownloads);
         for (DownloadBatch downloadBatch : downloadBatches) {
+            if (downloadBatch.isActive()) {
+                isActive = true;
```
Electryc: Do we want to `break` once we've found an active downloadBatch in order not to go through all of them?

ouchadam: :+1: good point

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
ouchadam: no and yes, until we have a proper queue order column. The problem we're going to suffer with is running checks on the queue, should we block all downloads if 1 item in the queue has been denied?

amlcurran: well current functionality is that the downloads run first-in-first-out, that shouldn't change

ronocod: :+1: for FIFO ordering, it's the way I would assume it works if I was a new user of the lib

ouchadam: take this scenario,

A download is added to the queue, it's denied
Another download is added to the queue, it's allowed to download.

FIFO causes the first download to show it's `download rule error` and then start downloading the second.

Perhaps the question is whether a denied download should then block the rest of the queue?

blundell: Downloads are denied on an individual basis right? So then one should not block the other.

amlcurran: but from a UI perspective the second item goes to queued, which are shown below running ones e.g.:

Before check:
- Download 1 (will get denied) : queued
- Download 2 (won't get denied) : queued
- Download 3 (won't get denied) : queued

After check:
- Download 2 : running
- Download 1 : queued
- Download 3 : queued

ouchadam: yep so by adding a new download to the queue, it will trigger the check on all previously queued items, and if you have a queue of 100 for example, and all of them fail but the newly added item, there will be 100 errors displayed

amlcurran: Ah I see the issue now. BUT we should not change this functionality because we need FIFO. 

Also surely if all but the first couldn't run in this new scenario, you'd still get 99 errors?

hal9002:
https://ci.novoda.com/job/download-manager-prb/98/
Test PASSed.

hal9002:
https://ci.novoda.com/job/download-manager-prb/99/
Test PASSed.

hal9002:
https://ci.novoda.com/job/download-manager-prb/100/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadBatch.java
```diff
@@ -50,6 +50,35 @@ public boolean isDeleted() {
...
+    }
+
+    private long getNextActionMillisFor(long now, FileDownloadInfo info) {
+        if (DownloadStatus.isCompleted(status)) {
+            return Long.MAX_VALUE;
+        }
+        if (status != DownloadStatus.WAITING_TO_RETRY) {
+            return 0;
+        }
+        long when = info.restartTime(now);
```
ataulm: this is just like using `from` or `what`

library/src/main/java/com/novoda/downloadmanager/lib/NetworkChecker.java
blundell: i think it still should, it's nice to have on the request if people don't need any other rules

library/src/main/java/com/novoda/downloadmanager/lib/DownloadBatch.java
blundell: what do you recommend `nextActionTime`

library/src/main/java/com/novoda/downloadmanager/lib/DownloadThread.java
```diff
@@ -842,14 +866,40 @@ private void addRequestHeaders(State state, HttpURLConnection conn) {
...
+        context.sendBroadcast(intent);
+    }
+
+    public void broadcastIntentDownloadFailedInsufficientSpace() {
+        Intent intent = new Intent(DownloadManager.ACTION_DOWNLOAD_INSUFFICIENT_SPACE);
+        intent.setPackage(getPackageName());
+        intent.putExtra(DownloadManager.EXTRA_DOWNLOAD_ID, originalDownloadInfo.getId());
+        intent.setData(originalDownloadInfo.getMyDownloadsUri());
+        if (originalDownloadInfo.getExtras() != null) {
+            intent.putExtra(DownloadManager.EXTRA_EXTRA, originalDownloadInfo.getExtras());
```
ataulm: READ ALL ABOUT IT

library/src/main/java/com/novoda/downloadmanager/lib/FileDownloadInfo.java
```diff
@@ -17,21 +12,19 @@
...
 class FileDownloadInfo {
 
-    public static final String EXTRA_EXTRA = "com.novoda.download.lib.KEY_INTENT_EXTRA";
     private static final int UNKNOWN_BYTES = -1;
 
     // TODO: move towards these in-memory objects being sources of truth, and periodically pushing to provider.
 
     /**
      * Constants used to indicate network state for a specific download, after
+     * private final CanDownload canDownload;
```
ataulm: boom

library/src/main/java/com/novoda/downloadmanager/lib/DownloadThread.java
ronocod: lol

blundell: literally I had `private static final String EXTRA_EXTRA = "read all about it";` Xavi was not amused

hal9002:
https://ci.novoda.com/job/download-manager-prb/103/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadBatch.java
ataulm: :+1: even `time` would be better

library/src/main/java/com/novoda/downloadmanager/lib/DownloadThread.java
xrigau: 0 likes

PR #94, opened 2015-07-10
Add deleting status

zegnus:
## Feature needed ##

We need to bubble up immediately when a batch is marked for deletion.

The current implementation marks a flag in a table that doesn't affect the current status of a downloads

## Solution implemented ##

As soon as a batch is marked for deletion, the batch status immediately changes to a new status `deleting`. This status happens in the same table that any other statuses and so all listeners monitoring the status of a download will be immediately notified.

Previous | Now
--- | ---
after | after-new

Notice that the demo `DeleteActivity` has changed the way that listens for updates in the database. The previous implementation failed to refresh a second event if that one is faster than the `lastQueryTimestamp`, here's an example of the previous implementation (the gif old has this fixed in order to show updates):

after

### DIFF
--- a/demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/batches/ShowBatchesActivity.java
+++ b/demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/batches/ShowBatchesActivity.java
@@ -67,6 +67,7 @@ public class ShowBatchesActivity extends AppCompatActivity implements QueryForBa
                         break;
                     case R.id.show_batches_query_deleting:
                         query = new BatchQuery.Builder().withStatusFilter(DownloadManager.STATUS_DELETING).build();
+                        break;
                     default:
                         break;
                 }
--- a/demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/batches/ShowBatchesActivity.java
+++ b/demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/batches/ShowBatchesActivity.java
@@ -64,6 +64,9 @@ public class ShowBatchesActivity extends AppCompatActivity implements QueryForBa
                         break;
                     case R.id.show_batches_query_live:
                         query = new BatchQuery.Builder().withSortByLiveness().build();
+                        break;
+                    case R.id.show_batches_query_deleting:
+                        query = new BatchQuery.Builder().withStatusFilter(DownloadManager.STATUS_DELETING).build();
                     default:
                         break;
                 }
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
@@ -499,9 +499,9 @@ public class DownloadManager {
 
         if (batchesIds.length == 1) {
             contentResolver.update(downloadsUriProvider.getContentUri(), values, COLUMN_BATCH_ID + "=?", new String[]{String.valueOf(batchesIds[0])});
+        } else {
+            contentResolver.update(downloadsUriProvider.getContentUri(), values, getWhereClauseFor(batchesIds, COLUMN_BATCH_ID), longArrayToStringArray(batchesIds));
         }
-
-        contentResolver.update(downloadsUriProvider.getContentUri(), values, getWhereClauseFor(batchesIds, COLUMN_BATCH_ID), longArrayToStringArray(batchesIds));
     }
 
     private int markBatchesBeDeleted(long[] batchesIds) {
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
@@ -473,7 +473,7 @@ public class DownloadManager {
         if (ids.length == 1) {
             return contentResolver.update(ContentUris.withAppendedId(baseUri, ids[0]), values, null, null);
         }
-        return contentResolver.update(baseUri, values, getWhereClauseForIds(ids), longArrayToStringArray(ids));
+        return contentResolver.update(baseUri, values, getWhereClauseFor(ids, DownloadContract.Downloads._ID), longArrayToStringArray(ids));
     }
 
     /**
@@ -512,7 +512,11 @@ public class DownloadManager {
             return contentResolver.update(ContentUris.withAppendedId(downloadsUriProvider.getBatchesUri(), batchesIds[0]), valuesDelete, null, null);
         }
 
-        return contentResolver.update(downloadsUriProvider.getBatchesUri(), valuesDelete, getWhereClauseForIds(batchesIds), longArrayToStringArray(batchesIds));
+        return contentResolver.update(
+                downloadsUriProvider.getBatchesUri(),
+                valuesDelete,
+                getWhereClauseFor(batchesIds, DownloadContract.Downloads._ID),
+                longArrayToStringArray(batchesIds));
     }
 
     /**
@@ -669,7 +673,7 @@ public class DownloadManager {
         values.putNull(DownloadContract.Downloads.COLUMN_DATA);
         values.put(DownloadContract.Downloads.COLUMN_STATUS, DownloadStatus.PENDING);
         values.put(DownloadContract.Downloads.COLUMN_FAILED_CONNECTIONS, 0);
-        contentResolver.update(baseUri, values, getWhereClauseForIds(ids), longArrayToStringArray(ids));
+        contentResolver.update(baseUri, values, getWhereClauseFor(ids, DownloadContract.Downloads._ID), longArrayToStringArray(ids));
     }
 
     /**
@@ -795,14 +799,7 @@ public class DownloadManager {
         return downloadsUriProvider.getBatchesUri();
     }
 
-    /**
-     * Get a parameterized SQL WHERE clause to select a bunch of IDs.
-     */
-    static String getWhereClauseForIds(long[] ids) {
-        return getWhereClauseFor(ids, DownloadContract.Downloads._ID);
-    }
-
-    private static String getWhereClauseFor(long[] ids, String column) {
+    static String getWhereClauseFor(long[] ids, String column) {
         StringBuilder whereClause = new StringBuilder();
         whereClause.append("(");
         for (int i = 0; i < ids.length; i++) {
@@ -816,9 +813,6 @@ public class DownloadManager {
         return whereClause.toString();
     }
 
-    /**
-     * Get the selection args for a clause returned by {@link #getWhereClauseForIds(long[])}.
-     */
     private static String[] longArrayToStringArray(long[] ids) {
         String[] whereArgs = new String[ids.length];
         for (int i = 0; i < ids.length; i++) {
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
@@ -198,6 +198,11 @@ public class DownloadManager {
     public static final int STATUS_FAILED = 1 << 4;
 
     /**
+     * Value of {@link #COLUMN_STATUS} when the download is marked for deletion.
+     */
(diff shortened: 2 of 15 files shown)


hal9002:
https://ci.novoda.com/job/download-manager-prb/96/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
```diff
@@ -472,21 +481,38 @@ public int removeDownloads(long... ids) {
...
      * If there are any downloaded files, partial or complete, they will be deleted.
      *
-     * @param ids the IDs of the batches to remove
+     * @param batchesIds the IDs of the batches to remove
      * @return the number of batches actually removed
      */
-    public int removeBatches(long... ids) {
-        if (ids == null || ids.length == 0) {
-            throw new IllegalArgumentException("called with nothing to remove. input param 'ids' can't be null");
+    public int removeBatches(long... batchesIds) {
```
amlcurran: batchIds

zegnus: So the shall rename the method as well? `removeBatchIds`? It looks like a batch can have several ids instead of that there are batches with ids... wasn't sure about that

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
```diff
@@ -472,21 +481,38 @@ public int removeDownloads(long... ids) {
...
+
+    private void setDeletingStatusFor(long[] batchesIds) {
+        ContentValues values = new ContentValues(1);
+        values.put(DownloadContract.Downloads.COLUMN_STATUS, DownloadStatus.DELETING);
+
+        if (batchesIds.length == 1) {
+            contentResolver.update(downloadsUriProvider.getContentUri(), values, COLUMN_BATCH_ID + "=?", new String[]{String.valueOf(batchesIds[0])});
         }
-        return contentResolver.update(downloadsUriProvider.getBatchesUri(), values, getWhereClauseForIds(ids), longArrayToStringArray(ids));
+
```
amlcurran: should have an `else` here?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
```diff
@@ -472,21 +481,38 @@ public int removeDownloads(long... ids) {
...
-        // if only one id is passed in, then include it in the uri itself.
-        // this will eliminate a full database scan in the download service.
-        if (ids.length == 1) {
-            return contentResolver.update(ContentUris.withAppendedId(downloadsUriProvider.getBatchesUri(), ids[0]), values, null, null);
+
+        setDeletingStatusFor(batchesIds);
+        return markBatchesBeDeleted(batchesIds);
+    }
+
+    private void setDeletingStatusFor(long[] batchesIds) {
```
ouchadam: what's the difference between this and `markBatchesBeDeleted`? they both instantly update the database?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
```diff
@@ -472,21 +481,38 @@ public int removeDownloads(long... ids) {
...
+            contentResolver.update(downloadsUriProvider.getContentUri(), values, COLUMN_BATCH_ID + "=?", new String[]{String.valueOf(batchesIds[0])});
         }
-        return contentResolver.update(downloadsUriProvider.getBatchesUri(), values, getWhereClauseForIds(ids), longArrayToStringArray(ids));
+
+        contentResolver.update(downloadsUriProvider.getContentUri(), values, getWhereClauseFor(batchesIds, COLUMN_BATCH_ID), longArrayToStringArray(batchesIds));
+    }
+
+    private int markBatchesBeDeleted(long[] batchesIds) {
+        ContentValues valuesDelete = new ContentValues(1);
+        valuesDelete.put(DownloadContract.Batches.COLUMN_DELETED, 1);
```
amlcurran: why not update the status at the same time? lessens disk IO

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
```diff
@@ -472,21 +481,38 @@ public int removeDownloads(long... ids) {
...
+
+        if (batchesIds.length == 1) {
+            contentResolver.update(downloadsUriProvider.getContentUri(), values, COLUMN_BATCH_ID + "=?", new String[]{String.valueOf(batchesIds[0])});
         }
-        return contentResolver.update(downloadsUriProvider.getBatchesUri(), values, getWhereClauseForIds(ids), longArrayToStringArray(ids));
+
+        contentResolver.update(downloadsUriProvider.getContentUri(), values, getWhereClauseFor(batchesIds, COLUMN_BATCH_ID), longArrayToStringArray(batchesIds));
+    }
+
+    private int markBatchesBeDeleted(long[] batchesIds) {
```
amlcurran: markBatchesToBeDeleted :smile:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadStatus.java
```diff
@@ -13,6 +13,10 @@
 final class DownloadStatus {
 
     /**
+     * This download has been marked for deletion and it will be deleted in the future
+     */
+    public static final int DELETING = 188;
```
amlcurran: I'm not sure about this one - but the DownloadManager has error codes as >400, does this count as an error or not? Discuss!

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
zegnus: yes! missed it

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
zegnus: One is to add a flag in the database that this batch will be deleted by the DownloadService thread, so it marked now, will be deleted in the future.

This one sets **now** the new status, that will trigger a content observer update and everyone listening for the status changes in the database will be notified immediately. This is the part of the code that updates the status immediately after clicking on it instead of having to wait for 3s for the thread to pick it up and do the update.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
zegnus: They modify different Uris so cannot happen at the same time :(

I've splitted it on two to make it easier to read as they do two different things

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
zegnus: **to** :+1:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
ouchadam: this highlights the problem with `batches` compared to `downloads`, we really need a single point to observer on. Not going to block this PR but it's definitely something we should fix sooner than later

xrigau:
dat speed tho

10x+ speed on deletion <sup>perceived by the user</sup>

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
amlcurran: no method is fine, just the param :smile:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
amlcurran: :+1: completely missed that

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/batches/ShowBatchesActivity.java
```diff
@@ -64,6 +64,9 @@ public void onCheckedChanged(RadioGroup group, int checkedId) {
                         break;
                     case R.id.show_batches_query_live:
                         query = new BatchQuery.Builder().withSortByLiveness().build();
+                        break;
+                    case R.id.show_batches_query_deleting:
+                        query = new BatchQuery.Builder().withStatusFilter(DownloadManager.STATUS_DELETING).build();
```
blundell: :x: missing break

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
blundell: @ouchadam i think i've got this right: their are two columns, one is it's current status - and we've added a new status "deleting" , the other column is `should delete` which gets set to `true` here and this is always how the DL manager has worked.

Still questionable?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
```diff
@@ -773,13 +799,17 @@ public Uri getBatchesUri() {
      * Get a parameterized SQL WHERE clause to select a bunch of IDs.
      */
     static String getWhereClauseForIds(long[] ids) {
```
blundell: inline this method it's confusing that one use of `getWhere` calls the single param and one calls the double param, they should both just call the double param

library/src/main/java/com/novoda/downloadmanager/lib/DownloadStatus.java
blundell: sleepy Alex, this is a status code like `public static final int SUBMITTED = 189;` i assume

blundell:
:+1:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
zegnus: https://github.com/novoda/download-manager/commit/bb290944d48f96d34de7dfcfc8d0b83f5403857d

hal9002:
https://ci.novoda.com/job/download-manager-prb/97/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
zegnus: https://github.com/novoda/download-manager/commit/cb027b13dd5fda271d14d161e533caff9015bb1f

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
zegnus: https://github.com/novoda/download-manager/commit/7b22c513969917e90067818a2252fb33630b58e6

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/batches/ShowBatchesActivity.java
zegnus: https://github.com/novoda/download-manager/commit/9b405efbdb0bcd3e9b876160bb4e80276db7de85

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
zegnus: https://github.com/novoda/download-manager/commit/04f887eed5d66bf4c5931e521ae745e53e5fa624

zegnus:
@blundell @amlcurran All comments amended :)

library/src/main/java/com/novoda/downloadmanager/lib/DownloadStatus.java
amlcurran: yeah but status codes between 400 and 500 count as an error - just wondering if this is an error but probably not

amlcurran:
LGTM!

library/src/main/java/com/novoda/downloadmanager/lib/DownloadStatus.java
ouchadam: more of a status I suppose than an error

PR #95, opened 2015-07-13
Fix log warning about no icon being set

blundell:
we should always set a notification icon

https://github.com/shorelinedev/aosp_frameworks_base/blob/master/services/java/com/android/server/NotificationManagerService.java#L1842

```kotlin
07-13 12:03:25.501     470-2496/system_process E/NotificationService﹕ WARNING: In a future release this will crash the app: com.novoda.demo
07-13 12:03:25.533      470-470/system_process E/NotificationService﹕ Not posting notification with icon==0: Notification(pri=0 contentView=com.novoda.demo/0x1090082 vibrate=null sound=null defaults=0x0 flags=0x0 kind=[null])
```

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/NotificationDisplayer.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/NotificationDisplayer.java
@@ -82,7 +82,7 @@ class NotificationDisplayer {
                 builder.setSmallIcon(android.R.drawable.stat_sys_download_done);
                 break;
             default:
-                //don't set an icon if none matches
+                builder.setSmallIcon(android.R.drawable.stat_sys_warning);
                 break;
         }
     }


hal9002:
https://ci.novoda.com/job/download-manager-prb/101/
Test PASSed.

Electryc:
:+1: nice catch

This was wrong even before the refactoring https://github.com/novoda/download-manager/commit/e206d9a4daac4cd61d9788b3ecb908edcb0bdaf6#diff-ee31e2ad08f5842e25b4b0175c130855L199 so :+1:

ronocod:
Is this gonna set a warning icon on any actual notifications? It's kinda worrying for a user to see a warning icon if there's no actual problems occuring

xrigau:
@ronocod the only unhandled case is CANCELLED, but when a notification cancelled it should be dismissed so it should be fine

ouchadam:
EZ

PR #104, opened 2015-07-14
Reduce number of db update callbacks

zegnus:
## Description of the problem ##

Downloads are taking a really long time when the wifi connection is strong and good enough

## Causes of the problem ##

After some investigation, we have find out that there was a kind of recursive calls going from the download thread to the database and back to the download thread.

Here's the cycle:
- In the DownloadService, the download thread is triggered by a `Handler` that is listening to the database. When there is a change in the database, the handler is triggered.
- In the Handler method, there is a call to `updateTotalBytesFor(allDownloads)`. This method triggers a database update calling `batchRepository.updateCurrentSize(batchId)`
- The previous database update triggers a change in the database and so a news handler is immediately triggered

This causes an overload on the system that leads to a slowdown on the downloads

## Solution implemented ##

The size update for the batch doesn't need to happen immediately as there are other mechanisms that triggers updates if a download has finished completely.

We have put in place a security for too many unnecessary calls preventing the system to call this method more than once per second.

**Time to download 100MB**

Before | Now
--- | ---
10m | 5m10s

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
@@ -384,7 +384,7 @@ public class DownloadService extends Service {
         executor.submit(downloadThread);
     }
 
-    private long lastUpdate = 0;
+    private long lastUpdate;
 
     private void updateTotalBytesFor(Collection<FileDownloadInfo> downloadInfos) {
         if (SystemClock.elapsedRealtime() - lastUpdate < TimeUnit.SECONDS.toMillis(1)) {
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
@@ -63,6 +63,8 @@ public class DownloadService extends Service {
     // DownloadReceiver to protect our entire workflow.
 
     private static final boolean DEBUG_LIFECYCLE = false;
+    private static final long ONE_SECOND = TimeUnit.SECONDS.toMillis(1);
+
     private final ContentLengthFetcher contentLengthFetcher = new ContentLengthFetcher();
 
     private SystemFacade systemFacade;
@@ -387,7 +389,7 @@ public class DownloadService extends Service {
     private long lastUpdate;
 
     private void updateTotalBytesIfNecessaryFor(Collection<FileDownloadInfo> downloadInfos) {
-        if (SystemClock.elapsedRealtime() - lastUpdate < TimeUnit.SECONDS.toMillis(1)) {
+        if (SystemClock.elapsedRealtime() - lastUpdate < ONE_SECOND) {
             return;
         }
 
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
@@ -322,7 +322,7 @@ public class DownloadService extends Service {
         long now = systemFacade.currentTimeMillis();
 
         Collection<FileDownloadInfo> allDownloads = downloadsRepository.getAllDownloads();
-        updateTotalBytesFor(allDownloads);
+        updateTotalBytesIfNecessaryFor(allDownloads);
 
         List<DownloadBatch> downloadBatches = batchRepository.retrieveBatchesFor(allDownloads);
         for (DownloadBatch downloadBatch : downloadBatches) {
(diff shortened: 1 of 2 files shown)


hal9002:
https://ci.novoda.com/job/download-manager-prb/112/
Test PASSed.

ronocod:
This isn't necessary if we go for this instead: https://github.com/novoda/download-manager/pull/105

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
```diff
@@ -382,8 +386,16 @@ private void download(FileDownloadInfo info) {
         executor.submit(downloadThread);
     }
 
-    private void updateTotalBytesFor(Collection<FileDownloadInfo> downloadInfos) {
-        ContentValues values = new ContentValues();
+    private long lastUpdate;
+
+    private void updateTotalBytesIfNecessaryFor(Collection<FileDownloadInfo> downloadInfos) {
+        if (SystemClock.elapsedRealtime() - lastUpdate < ONE_SECOND) {
```
blundell: use a `Clock` object so that time is under our control

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
```diff
@@ -320,7 +324,7 @@ private boolean updateLocked() {
         long now = systemFacade.currentTimeMillis();
 
         Collection<FileDownloadInfo> allDownloads = downloadsRepository.getAllDownloads();
-        updateTotalBytesFor(allDownloads);
+        updateTotalBytesIfNecessaryFor(allDownloads);
```
ouchadam: only the current size updating needs to be protected by the timing

ouchadam:
:heavy_multiplication_x: I really really really feel like the thread should be the one responsible for updating the current size

zegnus:
I'm with @ronocod

zegnus:
@ronocod @ouchadam let's explain you both your points of views and make a decision by tomorrow and solve this :)

ouchadam:
conors solution seems more elegant, but I have a few queries (see what I
did there)
On 14 Jul 2015 20:34, "Ferran Garriga" <notifications@github.com> wrote:

> @ronocod <https://github.com/ronocod> @ouchadam
> <https://github.com/ouchadam> let's explain you both your points of views
> and make a decision by tomorrow and solve this :)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/novoda/download-manager/pull/104#issuecomment-121350508>
> .
>

-- 
Disclaimer: The information in this e-mail and any attachments is the 
property of Novoda Ltd and is confidential and may be legally privileged. 
This e-mail is intended solely for the person or organisation to which it 
is addressed. Any disclosure, copying or other use of the information by 
any person or organisation who is not the intended recipient is strictly 
prohibited and may be unlawful. If you have received this e-mail in error, 
please inform the sender immediately and delete/destroy this e-mail and any 
copies of it. Novoda Ltd has taken reasonable precautions to minimise the 
risk of any software viruses which may damage your systems, but we advise 
that you take the necessary steps to ensure that no virus contamination is 
suffered. Novoda Ltd does not accept any liability for any loss or damage 
caused by the transmission of any virus.

Novoda Ltd, Company No: 347444, Registered in Scotland Registered 
Office: C/O Alexander Sloan, 38 Cadogan Street, Glasgow, G2 7HF, Scotland.  VAT 
Registration Number  GB 984 2525 93

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
zegnus: True, but I've done it this way in order to avoid the iteration and the check every single time when only has to happen once.

With this implementation, the first check will happen anyway immediately or at maximum after 1s from the previous check, and we save the iteration and checks for every execution of the `updateLocked`

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
zegnus: I've end up doing this https://github.com/novoda/download-manager/commit/d8ca9569957b24577dfaa8c4da5295505a4946ac so everything is encapsulated :)

amlcurran:
Have you run absolutely ever demo multiple times and checked everything? These large architecture changes have broken stuff before

hal9002:
https://ci.novoda.com/job/download-manager-prb/117/
Test PASSed.

ronocod:
@ouchadam lol

xrigau:
:+1: @ouchadam NICE JOKE BRAH

xrigau:
SO FUN FUN FUN

ouchadam:
can we close this?

PR #106, opened 2015-07-14
Add last modified to batches

xrigau:
Adds a 'last modified' column to the batches table

The column is updated with `System.currentTimeMillis()` when a new row is created and every time a row is updated

I did some boyscouting too!

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
@@ -177,6 +177,12 @@ public class DownloadManager {
     public static final String COLUMN_DOWNLOAD_LAST_MODIFIED_TIMESTAMP = "download_last_modified_timestamp";
 
     /**
+     * Timestamp when the batch was last modified, in {@link System#currentTimeMillis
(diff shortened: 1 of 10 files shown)


library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
```diff
@@ -174,7 +174,13 @@
      * Timestamp when the download was last modified, in {@link System#currentTimeMillis
      * System.currentTimeMillis()} (wall clock time in UTC).
      */
-    public static final String COLUMN_LAST_MODIFIED_TIMESTAMP = "last_modified_timestamp";
+    public static final String COLUMN_DOWNLOAD_LAST_MODIFIED_TIMESTAMP = "download_last_modified_timestamp";
```
xrigau: wasn't sure about this maybe this one shouldn't be exposed publicly

hal9002:
https://ci.novoda.com/job/download-manager-prb/115/
Test FAILed.

amlcurran:
what is it used for?

ouchadam:
what @amlcurran said

PR #109, opened 2015-07-15
Notify batch modified after all downloads

Electryc:
- We have found out that `DownloadContract.Downloads.COLUMN_NOTIFICATION_EXTRAS` was not part of the columns bundled up in the Cursor for a download query, so I have added it. 

- Also, in case of filtering by multiple notification extras, we should not attempt if the passed in array is empty as well as the `OR`s done between the statements should be inclosed in brackets. 

- After we enqueue a `RequestBatch` along with its `Requests` we have to notify that the batch has changed. This should be done because conceptually the batch is available only when its downloads are in the database as well, but because we cannot insert the batch at the end (i.e. after we insert the `Requests`) due to the fact that we need the `batch_id` for the requests, we have to manually notify after the whole process has ended. This enables us when watching the batch table, the later not to be out of sync with the downloads table. 
Consider the scenario when you want to query the downloads table when something changes on the whole batch, in this particular situation because the `insert(Batch)` is done first [here](https://github.com/novoda/download-manager/blob/master/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java#L844) we will get a notification that the batch table has changed, but the downloads have not been yet inserted so we get a inconsistent response for our query. 

- Reduced the number of PMDs by suppressing `AvoidDuplicateLiterals` because it doesn't give us any benefit.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
@@ -847,6 +847,9 @@ public class DownloadManager {
             request.setBatchId(batchId);
             insert(request);
         }
+        contentResolver.notifyChange(getBatchesUri(), null);
+        contentResolver.notifyChange(getBatchesWithoutProgressUri(), null);
+
         return batchId;
     }
 
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
@@ -847,10 +847,14 @@ public class DownloadManager {
             request.setBatchId(batchId);
             insert(request);
         }
-        contentResolver.notifyChange(getBatchesUri(), null);
-        contentResolver.notifyChange(getBatchesWithoutProgressUri(), null);
+        notifyAllBatchHasBeenEnqueued();
 
         return batchId;
+    }
+
+    private void notifyAllBatchHasBeenEnqueued() {
+        contentResolver.notifyChange(getBatchesUri(), null);
+        contentResolver.notifyChange(getBatchesWithoutProgressUri(), null);
     }
 
     private long insert(RequestBatch batch) {
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
@@ -359,6 +359,7 @@ public class DownloadManager {
             DownloadContract.Downloads.COLUMN_CURRENT_BYTES + " AS " + COLUMN_BYTES_DOWNLOADED_SO_FAR,
             DownloadContract.Downloads.COLUMN_BATCH_ID,
             DownloadContract.Downloads.COLUMN_EXTRA_DATA,
+            DownloadContract.Downloads.COLUMN_NOTIFICATION_EXTRAS,
             DownloadContract.Batches.COLUMN_TITLE,
             DownloadContract.Batches.COLUMN_DESCRIPTION,
             DownloadContract.Batches.COLUMN_BIG_PICTURE,
--- a/library/src/main/java/com/novoda/downloadmanager/lib/Query.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/Query.java
@@ -227,7 +227,7 @@ public class Query {
         for (String filterExtra : filterNotificiationExtras) {
             parts.add(notificationExtrasClause(filterExtra));
         }
-        selectionParts.add(joinStrings(" OR ", parts));
+        selectionParts.add("(" + joinStrings(" OR ", parts) + ")");
     }
 
     private void filterByExtraData(List<String> selectionParts) {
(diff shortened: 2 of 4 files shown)


hal9002:
https://ci.novoda.com/job/download-manager-prb/121/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
```diff
@@ -359,6 +359,7 @@
             DownloadContract.Downloads.COLUMN_CURRENT_BYTES + " AS " + COLUMN_BYTES_DOWNLOADED_SO_FAR,
             DownloadContract.Downloads.COLUMN_BATCH_ID,
             DownloadContract.Downloads.COLUMN_EXTRA_DATA,
+            DownloadContract.Downloads.COLUMN_NOTIFICATION_EXTRAS,
```
Electryc: This wasn't part of the _translated_ cursor, so the cursor we've been given back when querying did not contains this column.

library/src/main/java/com/novoda/downloadmanager/lib/Query.java
```diff
@@ -220,14 +220,14 @@ private String getWhereClauseForBatchIds(long[] ids) {
...
-        if (filterNotificiationExtras == null) {
+        if (filterNotificiationExtras == null || filterNotificiationExtras.length == 0) {
             return;
         }
         List<String> parts = new ArrayList<>();
         for (String filterExtra : filterNotificiationExtras) {
             parts.add(notificationExtrasClause(filterExtra));
         }
-        selectionParts.add(joinStrings(" OR ", parts));
+        selectionParts.add("(" + joinStrings(" OR ", parts) + ")");
```
Electryc: Due to the fact that at the end of all filtering in [`runQuery()`](https://github.com/novoda/download-manager/blob/master/library/src/main/java/com/novoda/downloadmanager/lib/Query.java#L186) we append an `AND deleted!=1` these `OR`s should be inclosed within brackets.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
```diff
@@ -846,9 +847,16 @@ public long enqueue(RequestBatch batch) {
             request.setBatchId(batchId);
             insert(request);
         }
+        notifyAllBatchHasBeenEnqueued();
+
         return batchId;
     }
 
+    private void notifyAllBatchHasBeenEnqueued() {
```
ouchadam: :100: for the method

zegnus:
Can we add some tests?

Electryc:
@zegnus on the Query yes! will add, but on the notification part I'm not sure if we can test that :confused:

Electryc:
closing while adding some tests on `Query`

Electryc:
@zegnus  
Tests added for `Query`. 
Attempted to test `DownloadManager.enqueue(RequestBatch)`, but it turned out in order to set up the test we would require a lot of implementation details knowledge in the test making it not maintainable

hal9002:
https://ci.novoda.com/job/download-manager-prb/127/
Test FAILed.

Electryc:
restest this please

hal9002:
https://ci.novoda.com/job/download-manager-prb/128/
Test FAILed.

hal9002:
https://ci.novoda.com/job/download-manager-prb/129/
Test FAILed.

hal9002:
https://ci.novoda.com/job/download-manager-prb/130/
Test PASSed.

ronocod:
All demos work as expected :+1:

amlcurran:
mergin' in ten :ship:

amlcurran:
11 minutes, so close

library/src/test/java/com/novoda/downloadmanager/lib/QueryTest.java
```diff
@@ -129,6 +129,25 @@ public void givenNoExtraDataWhenTheQueryIsCreatedThenTheWhereStatementContainsNo
         assertSelectionDoesNotContain(COLUMN_EXTRA_DATA + " IN ");
     }
 
+    @Test
+    public void givenMultipleNotificationExtrasWhenTheQueryIsCreatedThenTheWhereStatementIsCorrect() {
```
xrigau: :+1: noice

blundell:
> We have found out that DownloadContract.Downloads.COLUMN_NOTIFICATION_EXTRAS was not part of the columns bundled up in the Cursor for a download query, so I have added it.

I think it wasn't part of the cursor on purpose because it wasn't designed to be queried it is expected to be past with Intent's later on when download's complete

xrigau:
yeah what @blundell said :+1:

Electryc:
@blundell yes, I've talked with @xrigau yesterday about this. A solution would be to add another column, something like `COLUMN_EXTERNAL_IDENTIFICATION` ? But at the same time how would you use the `COLUMN_NOTIFICATION_EXTRAS` , as in as a user I would like the intent from the download's complete broadcast to contain some kind of identification from the client side of the download that has just completed.

blundell:
you can use `COLUMN_NOTIFICATION_EXTRAS` to pass anything thought the DLM into a broadcast receiver, like a url or the time it started or an analytics id or whatever. The identification could also be sent, but thats not specifically this columns role, does that make sense

amlcurran:
@blundell but I guess we're abusing the `COLUMN_NOTIFICATION_EXTRAS` really?

blundell:
yes we are currently abusing it

xrigau:
@amlcurran that's in @Electryc 's TODO list :tongue:

PR #111, opened 2015-07-16
Simplify query status control

zegnus:
## Investigation done ##

I've noticed an excessive number of invocations to the creation of the `FileDownloadInfo` object.

This is the analysis:
- Creation of the `FileDownloadInfo` object have a class level `synchronise`, along with an extra database query for the headers
- `FileDownloadInfo` is fully created but only the values `status` and `control` are used, the rest of the values are completely useless for this trace
- The `status` and `control` are queried at the method `checkPausedOrCreated` and invokated at `DownloadThread.transferData`.

So everytime that there is a transfer data, we call checkPausedOrCreated that creates a full FileDownloadInfo object in order to just check the status and the control.

## Solution implemented ##

I've created a slimmed down version that only creates an object with the `status` and `control` fields, removing all the unnecessary code that is not involved in the `checkPausedOrCreated` task

**Time for downloading 100MB from the Demo application**

Before | Now
--- | ---
6m12s | 5m

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/FileDownloadInfo.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/FileDownloadInfo.java
@@ -449,4 +449,48 @@ class FileDownloadInfo {
             return cursor.getLong(cursor.getColumnIndexOrThrow(column));
         }
     }
+
+    static final class ControlStatus {
+
+        private int control;
+        private int status;
+
+        public ControlStatus(int control, int status) {
+            this.control = control;
+            this.status = status;
+        }
+
+        public boolean isPaused() {
+            return control == DownloadsControl.CONTROL_PAUSED;
+        }
+
+        public boolean isCanceled() {
+            return status == DownloadStatus.CANCELED;
+        }
+
+        static final class Reader {
+
+            private final ContentResolver contentResolver;
+            private final DownloadsUriProvider downloadsUriProvider;
+
+            public Reader(ContentResolver contentResolver, DownloadsUriProvider downloadsUriProvider) {
+                this.contentResolver = contentResolver;
+                this.downloadsUriProvider = downloadsUriProvider;
+            }
+
+            public ControlStatus newControlStatus(long id) {
+                String[] projection = {DownloadContract.Downloads.COLUMN_CONTROL, DownloadContract.Downloads.COLUMN_STATUS};
+                Uri uri = ContentUris.withAppendedId(downloadsUriProvider.getAllDownloadsUri(), id);
+                Cursor downloadsCursor = contentResolver.query(uri, projection, null, null, null);
+                try {
+                    downloadsCursor.moveToFirst();
+                    int control = downloadsCursor.getInt(0);
+                    int status = downloadsCursor.getInt(1);
+                    return new FileDownloadInfo.ControlStatus(control, status);
+                } finally {
+                    downloadsCursor.close();
+                }
+            }
+        }
+    }
 }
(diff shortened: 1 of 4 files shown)


hal9002:
https://ci.novoda.com/job/download-manager-prb/132/
Test PASSed.

blundell:
general idea seems sound to me :+1: didn't delve to deep in the code

library/src/main/java/com/novoda/downloadmanager/lib/FileDownloadInfo.java
```diff
@@ -454,4 +454,48 @@ private Long getLong(String column) {
...
+                this.downloadsUriProvider = downloadsUriProvider;
+            }
+
+            public ControlStatus newControlStatus(long id) {
+                String[] projection = {DownloadContract.Downloads.COLUMN_CONTROL, DownloadContract.Downloads.COLUMN_STATUS};
+                Uri uri = ContentUris.withAppendedId(downloadsUriProvider.getAllDownloadsUri(), id);
+                Cursor downloadsCursor = contentResolver.query(uri, projection, null, null, null);
+                try {
+                    downloadsCursor.moveToFirst();
+                    int control = downloadsCursor.getInt(0);
```
xrigau: a bit flaky using raw ints here IMO

zegnus: The reason is because we do the projection so we don't need to query again to the cursor where are the columns, we have just done it at the projection line. As well is saves execution time for querying the column int when we have already specify them in the projection.

amlcurran: why?

amlcurran:
Does it add any more allocations?

zegnus:
This reduces allocations :)

amlcurran:
Excellent :D

amlcurran:
Ah, so we only create a ControlStatus object rather than a whole new FileDownloadInfo? Nice.

Just make sure it works OK with pause & resume and then LGTM :+1:

library/src/main/java/com/novoda/downloadmanager/lib/FileDownloadInfo.java
```diff
@@ -454,4 +454,48 @@ private Long getLong(String column) {
...
+
+        public boolean isPaused() {
+            return control == DownloadsControl.CONTROL_PAUSED;
+        }
+
+        public boolean isCanceled() {
+            return status == DownloadStatus.CANCELED;
+        }
+
+        static final class Reader {
```
ouchadam: would prefer if we didn't follow this pattern and had separate repositories

zegnus: could have a discussion and modify this pattern in the FileDownloadInfo as well, but I would prefer if we do it in another PR

ouchadam: sounds good :+1:

zegnus:
@amlcurran @ouchadam @xrigau all comments amended?

library/src/main/java/com/novoda/downloadmanager/lib/FileDownloadInfo.java
```diff
@@ -454,4 +454,48 @@ private Long getLong(String column) {
             return cursor.getLong(cursor.getColumnIndexOrThrow(column));
         }
     }
+
+    static final class ControlStatus {
```
ouchadam: I like it!

ouchadam:
LGTM :100:

PR #116, opened 2015-07-16
Fix download goes to queued

zegnus:
## Problem description ##

We have seen how a batch moves to the `queued` status for no reason in the middle of a download

## Solution implemented ##

[PENDING TO BE UPDATED AFTER LASTS MODIFICATIONS]

After investigating the code, this is what is happening:
- The `downloadService` has the handler that executes the downloads and processes statuses. Inside the handler there is a method called `updateLocked` that returns a boolean and is assigned to a `isActive` local variable
- When `isActive` is false, the `downloadService` shuts down
- `updateLocked` checks if a client is ready for a download or if the downloadmanager itself is ready for a download.
- The check for `isDownloadManagerReadyToDownload` returns false and shuts down the `downloadService` when the status of the `FileDownloadInfo` is `SUCCESS`

The solution has been to add the status `SUCCESS` as a valid status and so the `downloadService` can proceed with a proper clean up

I provide several screenshots in order to show that the downloads are processed properly as expected

1 Download | Multiple downloads
--- | ---
after | after

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadBatch.java
@@ -113,12 +113,7 @@ class DownloadBatch {
     }
 
     public boolean isActive() {
-        for (FileDownloadInfo info : downloads) {
-            if (info.isSubmittedOrRunning()) {
-                return true;
-            }
-        }
-        return false;
+        return status == DownloadStatus.SUBMITTED || status == DownloadStatus.RUNNING;
     }
 
     public boolean scanCompletedMediaIfReady(DownloadScanner downloadScanner) {
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
@@ -997,6 +997,7 @@ public class DownloadManager {
 
         private int translateStatus(int status) {
             switch (status) {
+                case DownloadStatus.SUBMITTED:
                 case DownloadStatus.PENDING:
                     return STATUS_PENDING;
 
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
@@ -443,7 +443,12 @@ public class DownloadManager {
         ContentValues values = new ContentValues();
         values.put(DownloadContract.Downloads.COLUMN_CONTROL, DownloadsControl.CONTROL_RUN);
         values.put(DownloadContract.Downloads.COLUMN_STATUS, DownloadStatus.PENDING);
-        contentResolver.update(downloadsUriProvider.getAllDownloadsUri(), values, COLUMN_BATCH_ID + "=?", new String[]{String.valueOf(id)});
+        String where = COLUMN_BATCH_ID + "= ? AND " + DownloadContract.Downloads.COLUMN_STATUS + " != ?";
+        String[] selectionArgs = {String.valueOf(id), String.valueOf(DownloadStatus.PENDING)};
+        contentResolver.update(downloadsUriProvider.getAllDownloadsUri(), values, where, selectionArgs);
+
+        BatchRepository batchRepository = new BatchRepository(contentResolver, new DownloadDeleter(contentResolver), downloadsUriProvider);
+        batchRepository.updateBatchStatus(id, DownloadStatus.PENDING);
     }
 
     public void removeDownload(URI uri) {
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
@@ -453,7 +453,9 @@ public class DownloadManager {
         String[] selectionArgs = {String.valueOf(id), String.valueOf(DownloadStatus.PENDING)};
         contentResolver.update(downloadsUriProvider.getAllDownloadsUri(), values, where, selectionArgs);
 
-        BatchRepository batchRepository = new BatchRepository(contentResolver, new DownloadDeleter(contentResolver), downloadsUriProvider);
+        DownloadDeleter downloadDeleter = new DownloadDeleter(contentResolver);
+        RealSystemFacade systemFacade = new RealSystemFacade(GlobalState.getContext());
+        BatchRepository batchRepository = new BatchRepository(contentResolver, downloadDeleter, downloadsUriProvider, systemFacade);
         batchRepository.updateBatchStatus(id, DownloadStatus.PENDING);
     }
 
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadReadyChecker.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadReadyChecker.java
@@ -37,6 +37,7 @@ class DownloadReadyChecker {
         }
         switch (downloadInfo.getStatus()) {
             case 0: // status hasn't been initialized yet, this is a new download
+            case DownloadStatus.SUCCESS:
             case DownloadStatus.PENDING: // download is explicit marked as ready to start
             case DownloadStatus.RUNNING: // download interrupted (process killed etc) while
                 // running, without a chance to update the database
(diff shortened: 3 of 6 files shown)


hal9002:
https://ci.novoda.com/job/download-manager-prb/141/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
```diff
@@ -997,6 +997,7 @@ private long getErrorCode(int status) {
 
         private int translateStatus(int status) {
             switch (status) {
+                case DownloadStatus.SUBMITTED:
```
ouchadam: :+1: seems legit

ouchadam:
I would have assumed the cause of this issue is the batch status being set to success when individual downloads complete (similar issue to pause resume) rather than accumulating the total status

Electryc:
@ouchadam that's not happening at the moment as a batch is SUCCESS when **all** of its downloads are SUCESS. I think what you are saying is the PENDING status when if at least **one** download is PENDING then the whole batch is PENDING.

blundell:
on a non blocking for this PR note: that screenshot for multiple downloads :/ it should be a big notification with a list of what's downloading or multiple notifications? Not this crazy flashing thing

library/src/main/java/com/novoda/downloadmanager/lib/DownloadReadyChecker.java
```diff
@@ -37,6 +37,7 @@ private boolean isDownloadManagerReadyToDownload(FileDownloadInfo downloadInfo)
         }
         switch (downloadInfo.getStatus()) {
             case 0: // status hasn't been initialized yet, this is a new download
+            case DownloadStatus.SUCCESS:
```
Electryc: So in a batch if one of the downloads has been successfully completed it caused the service to shut down? I'm just wondering why as there are clearly downloads in the queue as being pending :confused: 

Does this means that even if a download has successfully completed there is a chance to be processed again?

amlcurran: when a  download is completed, we actually attempt to restart it again anyway - because we request *all* downloads in a batch not just uncomplete ones. This is a known issue - we came across it in pause/resume

zegnus:
@amlcurran @Electryc @ouchadam Any other question, are we ok with this? :)

amlcurran:
:x: just wanna verify it works with pause and resume in our secret app :smile:

ouchadam:
I think the problem is due to this https://github.com/novoda/download-manager/blob/master/library/src/main/java/com/novoda/downloadmanager/lib/DownloadThread.java#L921

We update the batch status based on the result of the individual result status, it should take into account that its part of a batch and continue to be running until all parts are complete

amlcurran:
This seems to work fine for me :+1: nice job @zegnus !

xrigau:
EZ PR, who wants to :ship: :it: ?

ouchadam:
not yet, we've got another solution being tried out

amlcurran:
@xrigau I think ferran might be making more chances, lets confirm with him

zegnus:
:X don't merge just yet :)

xrigau:
closing meanwhile

library/src/main/java/com/novoda/downloadmanager/lib/DownloadReadyChecker.java
ouchadam: it should be denied by this method though ^ 
we need to query for all downloads in order to update the notification

hal9002:
https://ci.novoda.com/job/download-manager-prb/144/
Test FAILed.

hal9002:
https://ci.novoda.com/job/download-manager-prb/145/
Test FAILed.

hal9002:
https://ci.novoda.com/job/download-manager-prb/146/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadBatch.java
```diff
@@ -113,12 +113,7 @@ public boolean prune(DownloadDeleter downloadDeleter) {
     }
 
     public boolean isActive() {
-        for (FileDownloadInfo info : downloads) {
-            if (info.isSubmittedOrRunning()) {
-                return true;
-            }
-        }
-        return false;
+        return status == DownloadStatus.SUBMITTED || status == DownloadStatus.RUNNING;
```
ouchadam: the batch is the truth!

ouchadam:
:100:  LGTM

xrigau:
PR description says

>[PENDING TO BE UPDATED AFTER LASTS MODIFICATIONS]

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
```diff
@@ -449,7 +449,14 @@ public void resumeBatch(long id) {
         ContentValues values = new ContentValues();
         values.put(DownloadContract.Downloads.COLUMN_CONTROL, DownloadsControl.CONTROL_RUN);
         values.put(DownloadContract.Downloads.COLUMN_STATUS, DownloadStatus.PENDING);
-        contentResolver.update(downloadsUriProvider.getAllDownloadsUri(), values, COLUMN_BATCH_ID + "=?", new String[]{String.valueOf(id)});
+        String where = COLUMN_BATCH_ID + "= ? AND " + DownloadContract.Downloads.COLUMN_STATUS + " != ?";
+        String[] selectionArgs = {String.valueOf(id), String.valueOf(DownloadStatus.PENDING)};
+        contentResolver.update(downloadsUriProvider.getAllDownloadsUri(), values, where, selectionArgs);
+
+        DownloadDeleter downloadDeleter = new DownloadDeleter(contentResolver);
+        RealSystemFacade systemFacade = new RealSystemFacade(GlobalState.getContext());
```
xrigau: :x: unnecessary as this is a field already in the class https://github.com/novoda/download-manager/blob/master/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java#L380

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
```diff
@@ -449,7 +449,14 @@ public void resumeBatch(long id) {
         ContentValues values = new ContentValues();
         values.put(DownloadContract.Downloads.COLUMN_CONTROL, DownloadsControl.CONTROL_RUN);
         values.put(DownloadContract.Downloads.COLUMN_STATUS, DownloadStatus.PENDING);
-        contentResolver.update(downloadsUriProvider.getAllDownloadsUri(), values, COLUMN_BATCH_ID + "=?", new String[]{String.valueOf(id)});
+        String where = COLUMN_BATCH_ID + "= ? AND " + DownloadContract.Downloads.COLUMN_STATUS + " != ?";
+        String[] selectionArgs = {String.valueOf(id), String.valueOf(DownloadStatus.PENDING)};
+        contentResolver.update(downloadsUriProvider.getAllDownloadsUri(), values, where, selectionArgs);
+
+        DownloadDeleter downloadDeleter = new DownloadDeleter(contentResolver);
```
xrigau: should be a ctor dependency for testability

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
```diff
@@ -449,7 +449,14 @@ public void resumeBatch(long id) {
...
+        RealSystemFacade systemFacade = new RealSystemFacade(GlobalState.getContext());
+        BatchRepository batchRepository = new BatchRepository(contentResolver, downloadDeleter, downloadsUriProvider, systemFacade);
```
xrigau: probably same here

Electryc: We considered that since the beginning. The problem is DownloadManagerBuilder is out of the `lib` package so it cannot see `BatchRepository` since it's package protected. The only way would be to expose it, but I it does more harm.

xrigau: ah I see

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
xrigau: this happens in many places so it'd be nice if you fix all of them at once

PR #117, opened 2015-07-17
Batch with timestamp

xrigau:
Adds a new column that keeps the timestamp of the last modification made to a batch, that is when the status of a batch was last modified.

We also renamed the LAST_MODIFIED column from Downloads so that we can use the same name as the one in Batches.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
@@ -39,16 +39,19 @@ class BatchRepository {
     private final ContentResolver resolver;
     private final DownloadDeleter downloadDeleter;
     private final DownloadsUriProvider downloadsUriProvider;
+    private final SystemFacade systemFacade;
 
-    BatchRepository(ContentResolver resolver, DownloadDeleter downloadDeleter, DownloadsUriProvider downloadsUriProvider) {
+    BatchRepository(ContentResolver resolver, DownloadDeleter downloadDeleter, DownloadsUriProvider downloadsUriProvider, SystemFacade systemFacade) {
         this.resolver = resolver;
         this.downloadDeleter = downloadDeleter;
         this.downloadsUriProvider = downloadsUriProvider;
+        this.systemFacade = systemFacade;
     }
 
     void updateBatchStatus(long batchId, int status) {
         ContentValues values = new ContentValues();
         values.put(DownloadContract.Batches.COLUMN_STATUS, status);
+        values.put(DownloadContract.Batches.COLUMN_LAST_MODIFICATION, systemFacade.currentTimeMillis());
         resolver.update(downloadsUriProvider.getBatchesUri(), values, DownloadContract.Batches._ID + " = ?", new String[]{String.valueOf(batchId)});
     }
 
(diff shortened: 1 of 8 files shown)


hal9002:
https://ci.novoda.com/job/download-manager-prb/142/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
```diff
@@ -39,16 +39,19 @@
...
         this.resolver = resolver;
         this.downloadDeleter = downloadDeleter;
         this.downloadsUriProvider = downloadsUriProvider;
+        this.systemFacade = systemFacade;
     }
 
     void updateBatchStatus(long batchId, int status) {
         ContentValues values = new ContentValues();
         values.put(DownloadContract.Batches.COLUMN_STATUS, status);
+        values.put(DownloadContract.Batches.COLUMN_LAST_MODIFICATION, systemFacade.currentTimeMillis());
```
ouchadam: this timestamp is only set when updating the status and not during a download?

xrigau: yep, PR desr. says it
>Adds a new column that keeps the timestamp of the last modification made to a batch, that is when the status of a batch was last modified.

xrigau: as in update of the last modified column during a download is in the download table itself

amlcurran:
what do you need it for?

xrigau:
secret

ronocod:
:+1: LGTM

library/src/main/java/com/novoda/downloadmanager/lib/DownloadContract.java
```diff
@@ -79,7 +79,7 @@ private DownloadContract() {
          * <P>Type: BIGINT</P>
          * <P>Owner can Read</P>
          */
-        public static final String COLUMN_LAST_MODIFICATION = "lastmod";
+        public static final String COLUMN_LAST_MODIFICATION = "last_modified_timestamp";
```
amlcurran: API breaking change

amlcurran:
:ship: in ten

library/src/main/java/com/novoda/downloadmanager/lib/DownloadContract.java
xrigau: why?

xrigau: the public API is the constant, not the string itself

friedger: I disagree

friedger:
11 minutes are over

ronocod:
:watch: :keycap_ten:

xrigau:
for merge

danybony:
LGTM :+1:

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
ouchadam: I would have argued for a `last_status_modification` :tongue:

PR #118, opened 2015-07-17
FileDownloadInfo setter removal

ouchadam:
Removes all the setters from `FileDownloadInfo`! Next step, all final fields. 

- Removes unused `cleanUpStaleDownloadsThatDisappeared` which was part of the in memory downloads syncing which is no longer needed

- Makes use of the `state.totalSize` instead of setting and refetching the the total download size from the `FileDownloadInfo`

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadThread.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadThread.java
@@ -742,7 +742,7 @@ class DownloadThread implements Runnable {
         if (state.mimeType != null) {
             values.put(DownloadContract.Downloads.COLUMN_MIME_TYPE, state.mimeType);
         }
-        values.put(DownloadContract.Downloads.COLUMN_TOTAL_BYTES, originalDownloadInfo.getTotalBytes());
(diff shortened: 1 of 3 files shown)


library/src/main/java/com/novoda/downloadmanager/lib/DownloadThread.java
```diff
@@ -742,7 +742,7 @@ private void updateDatabaseFromHeaders(State state) {
         if (state.mimeType != null) {
             values.put(DownloadContract.Downloads.COLUMN_MIME_TYPE, state.mimeType);
         }
-        values.put(DownloadContract.Downloads.COLUMN_TOTAL_BYTES, originalDownloadInfo.getTotalBytes());
+        values.put(DownloadContract.Downloads.COLUMN_TOTAL_BYTES, state.totalBytes);
```
ouchadam: the method ordering makes this look strange, but at the is point the `state.totalSize` has already been set

hal9002:
https://ci.novoda.com/job/download-manager-prb/143/
Test PASSed.

Electryc:
LGTM less mutable state :+1:

xrigau:
:+1: seems legit

PR #124, opened 2015-07-21
Fix resumed downloads not being updated in the UI

amlcurran:
Resuming an item would not change its state from paused to resume in UI. This was because we failed to notify a change from the batches URI. 

As another nice thing, we only attempt to resume items which aren't complete, which reduces flickering and reduces the amount of threads we create

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
@@ -28,6 +28,7 @@ import android.os.Build;
 import android.os.ParcelFileDescriptor;
 import android.provider.Settings;
 import android.provider.Settings.SettingNotFoundException;
+import android.support.annotation.NonNull;
 import android.text.TextUtils;
 
 import com.novoda.notils.logger.simple.Log;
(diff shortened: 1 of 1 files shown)


hal9002:
https://ci.novoda.com/job/download-manager-prb/158/
Test PASSed.

amlcurran:
(I attempted to write a test but ContentValues tripped me over)

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
```diff
@@ -446,17 +447,24 @@ public void pauseBatch(long id) {
     }
 
     public void resumeBatch(long id) {
-        ContentValues values = new ContentValues();
-        values.put(DownloadContract.Downloads.COLUMN_CONTROL, DownloadsControl.CONTROL_RUN);
-        values.put(DownloadContract.Downloads.COLUMN_STATUS, DownloadStatus.PENDING);
+        ContentValues values = getResumeContentValues();
         String where = COLUMN_BATCH_ID + "= ? AND " + DownloadContract.Downloads.COLUMN_STATUS + " != ?";
-        String[] selectionArgs = {String.valueOf(id), String.valueOf(DownloadStatus.PENDING)};
+        String[] selectionArgs = {String.valueOf(id), String.valueOf(DownloadStatus.SUCCESS)};
```
xrigau: https://github.com/novoda/download-manager/commit/6464ebd3dbf43d0517829fcf1ac16190d448e3ef

>4 days ago

#awkward

xrigau: nah but seriously, how come this has changed?

from the PR desc. it seems like the fix was adding the `notifyBatchesHaveChanged` method?

amlcurran: > we only attempt to resume items which aren't complete, which reduces flickering

xrigau: wouldn't we need to do both? `STATUS != SUCCESS AND STATUS != PENDING`?

amlcurran: Doesn't really matter if they're pending and you set them to pending again.

Also @zegnus commit message says to update non-complete items only but it updated non-pending items - was this intentional?

zegnus: Seems ok, my fix was to ensure that we only change the minimum number of values, added the STATUS != PENDING won't change the result but will avoid to change some unnecessary rows.

I'm ok with this

zegnus:
:+1: LGTM

PR #125, opened 2015-07-21
Fix batch status flickering

amlcurran:
This PR adds fixes an issue where, when a download in a batch completes, the batch's status would briefly go to "queued", even though from the batch's point of view it is still running.

Also, it looks like the notification doesn't flicker either!

FYI the flicker happens at around 10MB

| Before | After |
| --- | --- |
| broken-wobble | fix-wobbly-state |

### DIFF
--- a/demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/batches/BatchesActivity.java
+++ b/demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/batches/BatchesActivity.java
@@ -10,7 +10,6 @@ import android.widget.RadioGroup;
 
 import com.novoda.downloadmanager.DownloadManagerBuilder;
 import com.novoda.downloadmanager.demo.R;
-import com.novoda.downloadmanager.demo.extended.QueryTimestamp;
 import com.novoda.downloadmanager.lib.BatchQuery;
 import com.novoda.downloadmanager.lib.DownloadManager;
 
@@ -20,7 +19,6 @@ import java.util.List;
 public class BatchesActivity extends AppCompatActivity implements QueryForBatchesAsyncTask.Callback {
 
     private final Handler handler = new Handler(Looper.getMainLooper());
-    private final QueryTimestamp lastQueryTimestamp = new QueryTimestamp();
 
     private DownloadManager downloadManager;
     private BatchesAdapter adapter;
@@ -92,11 +90,7 @@ public class BatchesActivity extends AppCompatActivity implements QueryForBatche
 
         @Override
         public void onChange(boolean selfChange) {
-            if (lastQueryTimestamp.updatedRecently()) {
-                return;
-            }
             queryForBatches(query);
-            lastQueryTimestamp.setJustUpdated();
         }
 
     };
--- a/library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
@@ -204,11 +204,11 @@ class BatchRepository {
         return resolver.query(downloadsUriProvider.getBatchesUri(), null, query.getSelection(), query.getSelectionArguments(), query.getSortOrder());
     }
 
-    private class StatusCountMap {
+    private static class StatusCountMap {
 
         private final SparseArrayCompat<Integer> statusCounts = new SparseArrayCompat<>(PRIORITISED_STATUSES_SIZE);
 
-        private boolean hasNoItemsWithStatuses(List<Integer> excludedStatuses) {
+        public boolean hasNoItemsWithStatuses(List<Integer> excludedStatuses) {
             boolean hasOtherItems = false;
             for (int status : excludedStatuses) {
                 boolean hasItemsForStatus = hasCountFor(status);
@@ -220,11 +220,11 @@ class BatchRepository {
             return hasOtherItems;
         }
 
-        private boolean hasCountFor(int status) {
+        public boolean hasCountFor(int status) {
             return statusCounts.get(status, 0) > 0;
         }
 
-        private void increment(int statusCode) {
+        public void increment(int statusCode) {
             int currentStatusCount = statusCounts.get(statusCode, 0);
             statusCounts.put(statusCode, currentStatusCount + 1);
         }
--- a/library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
@@ -31,7 +31,7 @@ class BatchRepository {
             DownloadStatus.SUCCESS
     );
 
-    private static final List<Integer> STATUSES_EXCEPT_PENDING_SUCCESS = Arrays.asList(
+    private static final List<Integer> STATUSES_EXCEPT_PENDING_SUCCESS_SUBMITTED = Arrays.asList(
             DownloadStatus.CANCELED,
             DownloadStatus.PAUSED_BY_APP,
             DownloadStatus.RUNNING,
@@ -40,9 +40,7 @@ class BatchRepository {
             // Paused statuses
             DownloadStatus.WAITING_TO_RETRY,
             DownloadStatus.WAITING_FOR_NETWORK,
-            DownloadStatus.QUEUED_FOR_WIFI,
-
-            DownloadStatus.SUBMITTED
+            DownloadStatus.QUEUED_FOR_WIFI
     );
 
     private static final int PRIORITISED_STATUSES_SIZE = PRIORITISED_STATUSES.size();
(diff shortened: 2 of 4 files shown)


library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
```diff
@@ -3,7 +3,7 @@
 import android.content.ContentResolver;
 import android.content.ContentValues;
 import android.database.Cursor;
-import android.util.SparseIntArray;
+import android.support.v4.util.SparseArrayCompat;
```
amlcurran: Using the compat version allows us to test the functionality, although does mean we get an autoboxing penalty.

hal9002:
https://ci.novoda.com/job/download-manager-prb/159/
Test FAILed.

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
```diff
@@ -86,8 +98,16 @@ int getBatchStatus(long batchId) {
             }
         }
 
+        boolean hasCompleteItems = statusCountMap.hasCountFor(DownloadStatus.SUCCESS);
+        boolean hasPendingItems = statusCountMap.hasCountFor(DownloadStatus.PENDING);
+        boolean hasSubmittedItems = statusCountMap.hasCountFor(DownloadStatus.SUBMITTED);
+        boolean hasOtherItems = statusCountMap.hasNoItemsWithStatuses(STATUSES_EXCEPT_PENDING_SUCCESS_SUBMITTED);
+        if (hasCompleteItems && (hasPendingItems || hasSubmittedItems) && !hasOtherItems) {
```
amlcurran: this is kinda grim, sorry.

hal9002:
https://ci.novoda.com/job/download-manager-prb/160/
Test PASSed.

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/batches/BatchesActivity.java
```diff
@@ -20,7 +19,6 @@
 public class BatchesActivity extends AppCompatActivity implements QueryForBatchesAsyncTask.Callback {
 
     private final Handler handler = new Handler(Looper.getMainLooper());
-    private final QueryTimestamp lastQueryTimestamp = new QueryTimestamp();
```
xrigau: :+1: forgot to get rid of this

library/src/test/java/com/novoda/downloadmanager/lib/BatchStatusTests.java
```diff
@@ -0,0 +1,312 @@
...
+    public void givenABatchWithSomePendingItemsAndSomeCompleteItemsAndSomeSubmittedItemsThenTheBatchStatusIsRunning() {
+        BatchRepository repository = givenABatchWithStatuses(DownloadStatus.PENDING, DownloadStatus.SUCCESS, DownloadStatus.SUBMITTED);
+
+        int batchStatus = repository.getBatchStatus(1);
+
+        assertThat(batchStatus).isEqualTo(DownloadStatus.RUNNING);
+    }
+
+    @Test
+    public void givenABatchWithAMixOfItemsThenTheBatchStatusIsNotRunning() {
```
xrigau: previous test name is really specific whereas this one isn't :-1: I had to scroll to see that 'a mix of items' means there's one failed item. why not using a more specific name here?

amlcurran: I couldn't think of one that wasn't really long :( but I'll flip the test name around

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
```diff
@@ -183,4 +203,35 @@ private void deleteBatchesForIds(List<Long> batchIdsToDelete, Collection<FileDow
...
+                if (hasItemsForStatus) {
+                    hasOtherItems = true;
+                    break;
+                }
+            }
+            return hasOtherItems;
+        }
+
+        public boolean hasCountFor(int status) {
+            return statusCounts.get(status, 0) > 0;
```
xrigau: maybe you can avoid some unboxing by using the `compareTo` method in the `Integer` class
http://docs.oracle.com/javase/7/docs/api/java/lang/Integer.html#compareTo(java.lang.Integer)

http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/lang/Integer.java#Integer.compareTo%28java.lang.Integer%29

xrigau: but don't think it's a bottleneck

library/src/test/java/com/novoda/downloadmanager/lib/BatchStatusTests.java
amlcurran: https://github.com/novoda/download-manager/commit/ef5d3721afb030a80f1b46acc8d79e9b2c81cd06

hal9002:
https://ci.novoda.com/job/download-manager-prb/161/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
amlcurran: We have the autoboxing as soon as we add something to the array. In Android OS there is a SparseIntArray which is backed by an `int[]` but there is no equivalent in the support library

amlcurran: :(

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
xrigau: I don't understand why we need to check if there's complete items?

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
```diff
@@ -183,4 +203,35 @@ private void deleteBatchesForIds(List<Long> batchIdsToDelete, Collection<FileDow
...
+    private static class StatusCountMap {
+
+        private final SparseArrayCompat<Integer> statusCounts = new SparseArrayCompat<>(PRIORITISED_STATUSES_SIZE);
+
+        public boolean hasNoItemsWithStatuses(List<Integer> excludedStatuses) {
+            boolean hasOtherItems = false;
+            for (int status : excludedStatuses) {
+                boolean hasItemsForStatus = hasCountFor(status);
+                if (hasItemsForStatus) {
+                    hasOtherItems = true;
```
xrigau: could just `return true` here and on L220 just `return false`

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
xrigau: yeah I saw that :disappointed:

library/src/test/java/com/novoda/downloadmanager/lib/BatchStatusTests.java
xrigau: :+1:

xrigau:
:+1: in general LGMT, just waiting on the Q about the completed status

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
amlcurran: If you have 1 complete and 2 pending items, the batch is technically running (you've just finished one download before starting the next), but the batch's status will actually be pending. 

But if it has no complete items you're at the start of the download and it is actually pending

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
amlcurran: :+1:

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
xrigau: It makes a bit more sense but I don't fully understand 100%, anyway I trust you and the tests so :+1: from me

xrigau:
will :ship:  if no other comments

ouchadam:
:+1: seems legit but I do worry if it will break something with some more extensive testing. let's find out!

amlcurran:
it's not like we've worried about that with our other PRs :tongue: 

But I have tested it in other places and it works OK :+1:

xrigau:
:roller_coaster:

Electryc:
> Also, it looks like the notification doesn't flicker either!

Winning.

blundell:
> it's not like we've worried about that with our other PRs

we're always worried

xrigau:
> > it's not like we've worried about that with our other PRs

> we're always worried

(⊙﹏⊙)

PR #128, opened 2015-07-22
Resume downloads after loosing signal

zegnus:
## Task description ##

The current download implementation does not support loosing the signal while processing batches

## Solution implemented ##

- When the `DownloadService` is about to be killed, we move all the submitted tasks to the batch status
- When the `DownloadThread`is about to abort due to client checks, we move that task to a new queued status
- Support for resume and TAGs has been added along with a new `Request` flag `applicationChecksFileIntegrity`. Notice that if eTags with resumable downloads is needed, the client must activate the new flag in the request.

This covers the following scenarios:
- Wifi drops while downloading, progress and submitted tasks goes into a queued status
- When wifi is recovered, all previous queued downloads continue
- When a download finishes and tries the following one, it is possible that the client check fails. Previously we were going into an unstable status. This is covered now.
- All queued downloads resumes and the batch finishes to download properly

**Flickering hasn't been reintroduced in this PR**
after


demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/batches/BatchDownloadsActivity.java
```diff
@@ -115,11 +113,7 @@ protected void onStart() {
 
         @Override
         public void onChange(boolean selfChange) {
-            if (lastQueryTimestamp.updatedRecently()) {
```
zegnus: This prevents the client to show status changes

hal9002:
https://ci.novoda.com/job/download-manager-prb/167/
Test FAILed.

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
```diff
@@ -99,10 +101,9 @@ int getBatchStatus(long batchId) {
         }
 
         boolean hasCompleteItems = statusCountMap.hasCountFor(DownloadStatus.SUCCESS);
-        boolean hasPendingItems = statusCountMap.hasCountFor(DownloadStatus.PENDING);
         boolean hasSubmittedItems = statusCountMap.hasCountFor(DownloadStatus.SUBMITTED);
         boolean hasOtherItems = statusCountMap.hasNoItemsWithStatuses(STATUSES_EXCEPT_PENDING_SUCCESS_SUBMITTED);
-        if (hasCompleteItems && (hasPendingItems || hasSubmittedItems) && !hasOtherItems) {
+        if (hasCompleteItems && hasSubmittedItems && !hasOtherItems) {
```
zegnus: Pending tasks don't imply that the batch is running. This causes the following bug:
- There are `SUCCESS` and `SUBMITTED` and `PENDING` tasks, the batch is running
- The submitted task finishes and moves to success, the `DownloadService` dies for any reason, like a phone reboot
- The application restarts again, the DownloadService starts again, reads the status of the batch and as there are success and pending tasks, the batch is active
- When the batch is active, no download is moved to the submitted status and no download is added to the executor.
- The DownloadStatus stays active forever and no tasks going to be added to the DownloadThread, nothing is downloading and all pending tasks stay pending.

zegnus:
@Dorvaryn Can you please help with this failing test?

```
InputStream resourceAsStream = getResourceAsStream("tar/testExpectedResult.tar");
// resourceAsStream is null
```

at

```
java.lang.NullPointerException
	at org.apache.commons.io.IOUtils.copyLarge(IOUtils.java:1025)
	at org.apache.commons.io.IOUtils.copy(IOUtils.java:999)
	at com.novoda.downloadmanager.lib.TarFileTruncatorTest.givenATarFileWithoutEndBlockMarker(TarFileTruncatorTest.java:61)
	at com.novoda.downloadmanager.lib.TarFileTruncatorTest.itDoeNotModifyATarFileWithoutEndBlockMarker(TarFileTruncatorTest.java:38)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runners.Suite.runChild(Suite.java:128)
	at org.junit.runners.Suite.runChild(Suite.java:27)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:78)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:212)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:68)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at com.intellij.rt.execution.application.AppMain.main(AppMain.java:140)
```

hal9002:
https://ci.novoda.com/job/download-manager-prb/168/
Test PASSed.

hal9002:
https://ci.novoda.com/job/download-manager-prb/169/
Test PASSed.

blundell:
So this functionality we should already of had. Did you find out what we broke in the first place to stop this working? 
It looks like you did a lot of changes ...

hal9002:
https://ci.novoda.com/job/download-manager-prb/170/
Test PASSed.

zegnus:
@blundell Yes, here a brief explanation:

- Downloads were not resumable unless the special extra request flag was activated `alwaysAttemptResume()`. That flag bypass all eTags. If we want to actually use eTags, the previous code would always reject the resumable status because the eTag was never set to the `FileDownloadInfo`. After setting the missing eTag value from the header reply, the download is still not resumable because it is mandatory that for a resumable download with eTag, the client checks the integrity of the file after the download has finished. This is solved with the new integrity request flag. With this fix, a download is resumable through eTags.
- When a batch is downloading, we have an active batch with some tasks as running, some tasks maybe as success, one or two tasks as submitted and the rest as pending. When there is a drop on wifi, the downloading task will have a socket exception and it will be stored with a new status as a queued for wifi, and so the batch. The DownloadService is destroyed as there are no any active batches because nothing can be downloaded. When we regain wifi again, the batch status is loaded and because we still have some submitted tasks, the status of the batch is running. As a running batch, no more tasks will be added to the new DownloadThread and so nothing will be downloaded. The fix was to move the submitted tasks status to the batch status when no more tasks can be executed.
- There was an edge case where we have downloading and a submitted task. We drop wifi but the downloading tasks gets to finish and the status are now success and submitted. No exception is thrown as when the submitted tasks tries to be executed, the DownloadThread aborts because the client check denies as there is no wifi. The thread dies BUT the download service stays active because a submitted tasks means that the batch is active. When we regain wifi, the submitted tasks are not resubmitted again and so no tasks are going to be added to the thread but the service will stay forever active.

amlcurran:
So did you confirm that this doesn't break the flickering status?

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
```diff
@@ -232,6 +235,18 @@ public void increment(int statusCode) {
         public void clear() {
             statusCounts.clear();
         }
+
+        @Override
+        public String toString() {
```
amlcurran: Is this debug code?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
```diff
@@ -370,9 +371,34 @@ private boolean updateLocked() {
...
         }
 
+        if (!isActive) {
+            moveSubmittedTasksToBatchStatusIfNecessary();
+        }
+
         return isActive;
     }
 
+    private void moveSubmittedTasksToBatchStatusIfNecessary() {
```
amlcurran: is there any way we can do this without so many queries?

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
zegnus: Nope, the string representation of that class is undefined, so I've defined it.

amlcurran: Do we need it though?

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
```diff
@@ -232,6 +235,18 @@ public void increment(int statusCode) {
...
+            StringBuilder stringBuilder = new StringBuilder("{");
+
+            int size = statusCounts.size();
+            for (int i = 0; i < size; i++) {
+                stringBuilder.append("[status: " + statusCounts.keyAt(i) + ", count: " + statusCounts.valueAt(i)).append("]");
```
blundell: if you're going to use a stringbuilder then use it ... drop all of the `+`

zegnus: Yup!!!!!

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
zegnus: No we don't but is a good practice to have it, I don't see any damage leaving it :)

hal9002:
https://ci.novoda.com/job/download-manager-prb/171/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
```diff
@@ -31,16 +32,19 @@
...
 
-    private static final List<Integer> STATUSES_EXCEPT_PENDING_SUCCESS_SUBMITTED = Arrays.asList(
+    private static final List<Integer> STATUSES_EXCEPT_SUCCESS_SUBMITTED = Arrays.asList(
             DownloadStatus.CANCELED,
             DownloadStatus.PAUSED_BY_APP,
             DownloadStatus.RUNNING,
             DownloadStatus.DELETING,
 
             // Paused statuses
+            DownloadStatus.QUEUED_DUE_CLIENT_PERMISSIONS,
```
xrigau: instead of `PERMISSIONS` wouldn't it be `RESTRICTIONS`?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadReadyChecker.java
```diff
@@ -18,29 +20,24 @@
...
-        if (downloadBatch.getStatus() == DownloadStatus.PENDING) {
-            return true;
-        }
-
-        for (FileDownloadInfo fileDownloadInfo : downloadBatch.getDownloads()) {
-            if (!isDownloadManagerReadyToDownload(fileDownloadInfo)) {
-                return false;
-            }
+        if (isDownloadManagerReadyToDownload(downloadBatch)) {
+            return downloadClientReadyChecker.isAllowedToDownload(downloadMarshaller.marshall(downloadBatch));
```
xrigau: extract var for the batch?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadReadyChecker.java
```diff
@@ -48,19 +45,50 @@ private boolean isDownloadManagerReadyToDownload(FileDownloadInfo downloadInfo)
...
-                return downloadInfo.restartTime(now) <= now;
+                long now = systemFacade.currentTimeMillis();
+                return isThereARetryDownloadThatCanRestart(downloads, now);
             case DownloadStatus.DEVICE_NOT_FOUND_ERROR:
                 // is the media mounted?
                 return Environment.getExternalStorageState().equals(Environment.MEDIA_MOUNTED);
             case DownloadStatus.INSUFFICIENT_SPACE_ERROR:
                 // avoids repetition of retrying download
                 return false;
+            default:
```
xrigau: :+1:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
```diff
@@ -370,9 +371,34 @@ private boolean updateLocked() {
...
+        List<FileDownloadInfo> allDownloads = downloadsRepository.getAllDownloads();
+        List<DownloadBatch> downloadBatches = batchRepository.retrieveBatchesFor(allDownloads);
```
xrigau: following @amlcurran 's comment, we could avoid doing this if we could `getAllBatches()` instead of getting all the downloads and then all the batches from those downloads (which I assume it's all the batches, correct me if this is wrong)

library/src/main/java/com/novoda/downloadmanager/lib/DownloadThread.java
```diff
@@ -226,30 +227,30 @@ private void runInternal() {
...
-
-//        final NetworkPolicyManager netPolicy = NetworkPolicyManager.from(context);
-        PowerManager powerManager = (PowerManager) context.getSystemService(Context.POWER_SERVICE);
+        State state = new State(originalDownloadInfo);
 
         try {
+            DownloadBatch currentBatch = batchRepository.retrieveBatchFor(originalDownloadInfo);
+
+            if (!downloadReadyChecker.canDownload(currentBatch)) {
+                throw new StopRequestException(DownloadStatus.QUEUED_DUE_CLIENT_PERMISSIONS, "Cannot proceed because client denies");
```
xrigau: I'd avoid doing this since we're using exceptions as control flow and that's not a nice solution

http://c2.com/cgi/wiki?DontUseExceptionsForFlowControl

I'd rather extract the code in the catch in a method and call that method for example, than do this

but :+1: nice catch, I assume [this](https://github.com/novoda/download-manager/pull/128/files#diff-9b10a080ea5fa55b1f8247c09b432b07L233) was what left the DLM in an inconsistent state?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
ronocod: Would this work?
```sql
UPDATE downloads
SET downloads.status = batch.status
INNER JOIN batch ON download.batchId = batch._id
WHERE status = SUBMITTED
```

library/src/main/java/com/novoda/downloadmanager/lib/DownloadThread.java
amlcurran: Using exceptions for control flow is how everything is done in the Thread atm, I think we should keep it all consistent for now and refactor it all at the same time

library/src/main/java/com/novoda/downloadmanager/lib/Request.java
```diff
@@ -246,6 +247,15 @@ public Request allowTarUpdates() {
     }
 
     /**
+     * When a ETag header is present, the application should check the integrity of the
+     * downloaded file, otherwise the current download won't be able to be resumed
```
xrigau: this says

>the application should check the integrity of the downloaded file

what does that mean? does the client need to check the md5 of the downloaded file or something like this?

xrigau: or is this more like an `unsafe resume` mode?

amlcurran:
I'm a :x: until @zegnus confirms this doesn't regress #125

amlcurran:
Saying that, I think #125 may have regressed stuff...

zegnus:
@amlcurran I've updated the description with a gif, no flickering :)

library/src/main/java/com/novoda/downloadmanager/lib/Request.java
zegnus: Correct, when resuming a download through eTags, the application has to check the integrity of the downloaded file as it could be possible that the server changes the file keeping the same eTag

library/src/main/java/com/novoda/downloadmanager/lib/DownloadThread.java
zegnus: @xrigau That is bad indeed but what @amlcurran said

ouchadam: ^ yep like the others said, that's how it works at the moment

amlcurran:
:+1: thanks!

library/src/main/java/com/novoda/downloadmanager/lib/Request.java
amlcurran: I don't think that is possible, or at least it is a violation of the spec so maybe we don't need to handle this? http://tools.ietf.org/html/rfc7232#section-2.3

zegnus: I didn't know about that check as well so this is what I found out:

```
/**
         * The name of the column containing the flags that indicates whether
         * the initiating application is capable of verifying the integrity of
         * the downloaded file. When this flag is set, the download manager
         * performs downloads and reports success even in some situations where
         * it can't guarantee that the download has completed (e.g. when doing
         * a byte-range request without an ETag, or when it can't determine
         * whether a download fully completed).
         * <P>Type: BOOLEAN</P>
         * <P>Owner can Init</P>
         */
        public static final String COLUMN_NO_INTEGRITY = "no_integrity";
```

And the Google coded this:
```
public boolean isResumable() {
        return alwaysResume || (eTag != null && isNoIntegrity());
    }
```

So I assumed that they forced that if there is an eTag, the flag `noIntegrity` is checked, leading at that if the default flag is false, with an eTag present, the download is non resumable

hal9002:
https://ci.novoda.com/job/download-manager-prb/174/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
zegnus: Yeah I guess that it will work, but I would prefer to leave my code in this PR and I encourage @ronocod to do that in another PR ;)

hal9002:
https://ci.novoda.com/job/download-manager-prb/177/
Test PASSed.

xrigau:
> This branch has conflicts that must be resolved

hal9002:
https://ci.novoda.com/job/download-manager-prb/180/
Test PASSed.

library/src/test/java/com/novoda/downloadmanager/lib/BatchStatusTests.java
```diff
@@ -9,16 +9,14 @@
 import android.os.Bundle;
 import android.support.annotation.NonNull;
 
-import org.junit.Test;
```
amlcurran: UH OH mine or @zegnus import statements are wrong again :(

amlcurran: actually it my fault - that'll teach me for not committing with Intellij (which organises imports on commit)

blundell: all those things you said

xrigau: lol

zegnus:
@amlcurran @ronocod @xrigau this is ready to go :)

library/src/main/java/com/novoda/downloadmanager/lib/DownloadsRepository.java
```diff
@@ -59,6 +63,37 @@ public FileDownloadInfo getDownloadFor(long id) {
...
+
+        ContentValues values = new ContentValues(1);
+        values.put(DownloadContract.Downloads.COLUMN_STATUS, status);
+
+        String where = DownloadContract.Downloads._ID + " IN (" + getPlaceHoldersOf(ids.size()) + ")";
+        String[] selectionArgs = getStringArrayFrom(ids.toArray());
+        contentResolver.update(downloadsUriProvider.getAllDownloadsUri(), values, where, selectionArgs);
+    }
+
+    private String getPlaceHoldersOf(int size) {
```
xrigau: use NoTils? https://github.com/novoda/notils/pull/62

library/src/main/java/com/novoda/downloadmanager/lib/DownloadThread.java
```diff
@@ -673,6 +679,9 @@ private void updateDatabaseFromHeaders(State state) {
             values.put(DownloadContract.Downloads.COLUMN_MIME_TYPE, state.mimeType);
         }
 
+        originalDownloadInfo.setETag(state.headerETag);
```
xrigau: method says `updateDatabaseFromHeaders` but you're actually updating fields, what do you think of moving these 2 somewhere else?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadThread.java
```diff
@@ -228,30 +229,30 @@ private void runInternal() {
...
+                ContentValues contentValues = new ContentValues();
+                contentValues.put(COLUMN_STATUS, DownloadStatus.RUNNING);
+                context.getContentResolver().update(originalDownloadInfo.getAllDownloadsUri(), contentValues, null, null);
+                updateBatchStatus(originalDownloadInfo.getBatchId(), originalDownloadInfo.getId());
+            }
+
+    //        final NetworkPolicyManager netPolicy = NetworkPolicyManager.from(context);
+            PowerManager powerManager = (PowerManager) context.getSystemService(Context.POWER_SERVICE);
+
+        //try {
```
xrigau: remove commented code

library/src/main/java/com/novoda/downloadmanager/lib/DownloadThread.java
```diff
@@ -228,30 +229,30 @@ private void runInternal() {
...
+            }
+
+            if (downloadStatus != DownloadStatus.RUNNING) {
+                ContentValues contentValues = new ContentValues();
+                contentValues.put(COLUMN_STATUS, DownloadStatus.RUNNING);
+                context.getContentResolver().update(originalDownloadInfo.getAllDownloadsUri(), contentValues, null, null);
+                updateBatchStatus(originalDownloadInfo.getBatchId(), originalDownloadInfo.getId());
+            }
+
+    //        final NetworkPolicyManager netPolicy = NetworkPolicyManager.from(context);
```
xrigau: remove commented code

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
xrigau: agreed :+1:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadsRepository.java
danybony: Will be a separate PR after this is merged

zegnus: Notils upgrade was created because of this and decided to modify it in another PR in order to keep this small. What @danybony said :)

zegnus:
@xrigau All comments amended :)

hal9002:
https://ci.novoda.com/job/download-manager-prb/181/
Test PASSed.

zegnus:
@xrigau @ronocod @amlcurran This PR is ready to be merged :)

xrigau:
LGMT :+1: will :shipit: if no other comments

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
ronocod: Sounds good :+1:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadsRepository.java
blundell: fyi it would be **smaller** using notils as I wouldn't have to read this code

xrigau: I know rite

xrigau: these guys owe me some brain power :battery:

PR #130, opened 2015-07-23
Fix concurrent downloads and memory leaks

amlcurran:
An issue was occurring where pausing three downloads and quickly resuming them again resulting in two downloads at the same time (if the limit was one download at a time).

It was determined that the service would shut down but the executor free to continue downloading (which it did), and then resuming another download triggered another service with a *different* executor.

The fix is to force shutting down the executor when the service stops.

Also, we now check the pause state at the beginning of the download task because it wouldn't actually pause until it had downloaded the first 4KB.

I have also made some changes that @zegnus made in his PR #128 as these were blocking me too


library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
```diff
@@ -227,6 +227,7 @@ private void shutDown() {
         Log.d("Shutting down service");
         getContentResolver().unregisterContentObserver(downloadManagerContentObserver);
         downloadScanner.shutdown();
+        executor.shutdownNow();
```
amlcurran: this what was actually fixed the leak

hal9002:
https://ci.novoda.com/job/download-manager-prb/173/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
zegnus: This is weird because the executor shouldn't have any pending threads as they should imply that there is an active batch and while an active batch the DownloadService should remain active.

When no downloads can proceed, all threads should die, the status of the batch should not be active and the service  die nicely, this shouldn't be necessary.

Maybe there is a bug somewhere else that is preventing the work that I've described to happen?

amlcurran: yeah, actually given I've added the fix you made into my branch, I'll see if that fixes it because that could have been the issue

(but we should shutdown the executor anyway, its good practise)

amlcurran:
Closing this, we're going to work branching off Ferran's branch

amlcurran:
Reopening - we've determined that this is independent of @zegnus' branch

hal9002:
https://ci.novoda.com/job/download-manager-prb/175/
Test PASSed.

hal9002:
https://ci.novoda.com/job/download-manager-prb/176/
Test PASSed.

library/src/test/java/com/novoda/downloadmanager/lib/BatchStatusTests.java
```diff
@@ -25,8 +25,8 @@
 public class BatchStatusTests {
 
     @Test
-    public void givenABatchWithSomePendingItemsAndSomeCompleteItemsThenTheBatchStatusIsRunning() {
-        BatchRepository repository = givenABatchWithStatuses(DownloadStatus.PENDING, DownloadStatus.SUCCESS, DownloadStatus.PENDING);
+    public void givenABatchWithSomeCompleteItemsAndSomeSubmittedItemsThenTheBatchStatusIsRunning() {
```
ronocod: :+1:

ronocod:
LGTM :+1:

library/src/test/java/com/novoda/downloadmanager/lib/BatchStatusTests.java
amlcurran: this one is @zegnus work :+1:

zegnus:
This is good enough to me for now, we don't want to do massive changes again in order to make sure that the thread sees the status in sync :+1:

PR #131, opened 2015-07-23
Revert "Fix concurrent downloads and memory leaks"

amlcurran:
Reverts novoda/download-manager#130

Removed all of Ben's TAR code

PR #132, opened 2015-07-23
Revert "Revert "Fix concurrent downloads and memory leaks""

amlcurran:
Reverts novoda/download-manager#131

Turns out, it didn't remove Ben's code.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadThread.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadThread.java
@@ -86,6 +86,7 @@ class DownloadThread implements Runnable {
     private final DownloadsRepository downloadsRepository;
     private final NetworkChecker networkChecker;
     private final DownloadReadyChecker downloadReadyChecker;
+    private final Clock clock;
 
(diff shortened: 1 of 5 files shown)


hal9002:
https://ci.novoda.com/job/download-manager-prb/179/
Test PASSed.

zegnus:
Oh dear :)

library/src/main/java/com/novoda/downloadmanager/lib/DownloadThread.java
```diff
@@ -343,6 +346,7 @@ private void runInternal() {
      * handle the response, and transfer the data to the destination file.
      */
     private void executeDownload(State state) throws StopRequestException {
+        checkPausedOrCanceled(originalDownloadInfo);
```
Dorvaryn: @zegnus Why is this checked here as well ? This case should never happen right ?

amlcurran: if the task is already submitted to the executor, it will start part of the download and only after check whether it should be paused

PR #133, opened 2015-07-24
Fix deleteBatchesForIds when multiple selectionArgs

danybony:
### Problem
`BatchRepository.deleteBatchesForIds` was working just because it was always invoked with a single element in the `batchIdsToDelete` parameter list.

### Solution
We updated the notils dependency to 2.2.13 and use the new QueryUtils method in order to create the right placeholder string, depending on the number of selection args.

We also replace the duplicated code in `DownloadsRepository`


hal9002:
https://ci.novoda.com/job/download-manager-prb/182/
Test FAILed.

xrigau:
oh and nao we have to review more code :sob:

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
```diff
@@ -198,9 +199,9 @@ private void deleteBatchesForIds(List<Long> batchIdsToDelete, Collection<FileDow
             }
         }
 
-        String selection = StringUtils.join(batchIdsToDelete, ", ");
-        String[] selectionArgs = {selection};
-        resolver.delete(downloadsUriProvider.getBatchesUri(), DownloadContract.Batches._ID + " IN (?)", selectionArgs);
+        String where = DownloadContract.Batches._ID + " IN (" + QueryUtils.createSelectionPlaceholdersOfSize(batchIdsToDelete.size()) + ")";
```
xrigau: would be nice to extract some vars here

danybony: :+1:

hal9002:
https://ci.novoda.com/job/download-manager-prb/183/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
```diff
@@ -410,6 +410,7 @@ public DownloadManager(Context context, ContentResolver contentResolver, boolean
         this.systemFacade = systemFacade;
         GlobalState.setContext(context);
         GlobalState.setVerboseLogging(verboseLogging);
+        Log.setShowLogs(verboseLogging);
```
danybony: This is needed singe NoTils 2.2.12 :/

zegnus:
:+1: LGTM, +10 minutes for more comments

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
xrigau: so this'll be removed in a later PR?

xrigau: @blundell ^?

danybony: Yes, this will be removed when we'll get rid of the Notils logger

friedger: Better fix notils logger.

blundell: changing this on my branch, so you can carry on

PR #138, opened 2015-07-24
Reducing control reader query method allocations

ouchadam:
More transfer data while(true) loop allocation reductions!

Forces the `ControlStatus.Reader` to be constructed with all its query dependencies rather than creating them each time a query is requested.

This means we don't have to create the uri or projection upon querying

Allocations for a single GC from a fresh install

| Before | After
| --- | --- 
|download-before|download-after

Total allocations before **81,034** After **80,375**


hal9002:
https://ci.novoda.com/job/download-manager-prb/195/
Test PASSed.

blundell:
benchmarks!

ouchadam:
not the most scientific... but added nonetheless

blundell:
:+1: let's wait for a 2nd thumb

xrigau:
:+1:

PR #146, opened 2015-08-05
Provide restart mechanism

zegnus:
### Feature requested ###

We need a way to restart the downloads workflow in case that it was not already active

A possible scenario: A client denies a download for a particular business rule and that rule does not apply any more. Calling this method will reactivate the downloads workflow, check the client rules and proceed if necessary

### Solution implemented ###

A new public method `restart` is provided

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
@@ -21,6 +21,7 @@ import android.content.ContentResolver;
 import android.content.ContentUris;
 import android.content.ContentValues;
 import android.content.Context;
+import android.content.Intent;
 import android.database.Cursor;
 import android.database.CursorWrapper;
 import android.net.Uri;
@@ -1100,5 +1101,17 @@ public class DownloadManager {
                     return STATUS_FAILED;
             }
         }
+    }
+
+    /**
+     * Restart will activate the downloads workflow in case that it was not already active
+     * <p/>
+     * A possible scenario: A client denies a download for a particular business rule and that
+     * rule does not apply any more. Calling this method will reactivate the downloads workflow,
+     * check the client rules and proceed if necessary
+     */
+    public void restart() {
+        Context context = GlobalState.getContext();
+        context.startService(new Intent(context, DownloadService.class));
     }
 }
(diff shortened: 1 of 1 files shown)


hal9002:
https://ci.novoda.com/job/download-manager-prb/203/
Test PASSed.

hal9002:
https://ci.novoda.com/job/download-manager-prb/204/
Test PASSed.

ouchadam:
this could be client side?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
```diff
@@ -1101,4 +1102,16 @@ private int translateStatus(int status) {
...
     }
+
+    /**
+     * Restart will activate the downloads workflow in case that it was not already active
+     * <p/>
+     * A possible scenario: A client denies a download for a particular business rule and that
+     * rule does not apply any more. Calling this method will reactivate the downloads workflow,
+     * check the client rules and proceed if necessary
+     */
+    public void restart() {
```
blundell: when would a client call this? Can the downloadmanager be clever and call restart itself?

zegnus:
@ouchadam Yup I guess that we could start the DownloadService client side but I would prefer to access the DownloadManager through public methods. What do you think?

ouchadam:
:+1: I understand the reasoning, maybe the name could be `forceStart`? `Restart` suggests that it is already running?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
zegnus: The downloadmanager has some mechanisms in order to restart itself automatically. For example when a new row is inserted in the database or there is a connection change, but there are cases where the client business rules don't modify the database and there is no network change either.

For example, let's say that in order to download something the user needs to be registered. If the user is not the download cannot proceed. Once the user is registered, the download can proceed.

In this scenario, after the user has registered, we can call `restart`, the download manager will proceed with all the checks and this time the paused download will be resumed as the client no longer denies.

But it could be any other random reason.

blundell: oh yes that makes sense :+1:  restart sounds so brutal lol `prod()` :=)

zegnus:
@ouchadam https://github.com/novoda/download-manager/commit/07cc7bd270ef78a847796de268d8186b3b9d6f8c

hal9002:
https://ci.novoda.com/job/download-manager-prb/205/
Test PASSed.

ronocod:
LGTM :+1:

PR #147, opened 2015-08-06
Add client checks along with pause and cancel checks and speed up calls

zegnus:
### Task requested ###

When a transfer is on its way, when we write down some transfered bytes we check if the current download has moved to paused or canceled but we do not check if there are client rules that could potentially deny the progress of the current download

### Solution implemented ###

The solution is to include the client check along with the current pause and canceled checks

In terms of performance implications:
- New originalDownloadBatch along with the originalDownloadInfo in order to avoid the creation of the former when there is a bunch of bytes transferred on write
- Exposed packaged protected the method that checks the client rules in order to avoid the overhead of checking the internals of the download manager, specially the unnecessary query for all downloads

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadReadyChecker.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadReadyChecker.java
@@ -23,13 +23,13 @@ class DownloadReadyChecker {
 
     public boolean canDownload(DownloadBatch downloadBatch) {
         if (isDownloadManagerReadyToDownload(downloadBatch)) {
-            return clientAllowToDownload(downloadBatch);
+            return clientAllowsToDownload(downloadBatch);
         }
 
         return false;
     }
 
-    boolean clientAllowToDownload(DownloadBatch downloadBatch) {
+    boolean clientAllowsToDownload(DownloadBatch downloadBatch) {
         Download download = downloadMarshaller.marshall(downloadBatch);
         return downloadClientReadyChecker.isAllowedToDownload(download);
     }
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadReadyChecker.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadReadyChecker.java
@@ -23,11 +23,15 @@ class DownloadReadyChecker {
 
     public boolean canDownload(DownloadBatch downloadBatch) {
         if (isDownloadManagerReadyToDownload(downloadBatch)) {
-            Download download = downloadMarshaller.marshall(downloadBatch);
-            return downloadClientReadyChecker.isAllowedToDownload(download);
+            clientAllowToDownload(downloadBatch);
         }
 
         return false;
+    }
+
+    boolean clientAllowToDownload(DownloadBatch downloadBatch) {
+        Download download = downloadMarshaller.marshall(downloadBatch);
+        return downloadClientReadyChecker.isAllowedToDownload(download);
     }
 
     private boolean isDownloadManagerReadyToDownload(DownloadBatch downloadBatch) {
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadReadyChecker.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadReadyChecker.java
@@ -23,7 +23,7 @@ class DownloadReadyChecker {
 
     public boolean canDownload(DownloadBatch downloadBatch) {
         if (isDownloadManagerReadyToDownload(downloadBatch)) {
-            clientAllowToDownload(downloadBatch);
+            return clientAllowToDownload(downloadBatch);
         }
 
         return false;
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
@@ -402,10 +402,12 @@ public class DownloadService extends Service {
     private void download(FileDownloadInfo info) {
         Uri downloadUri = ContentUris.withAppendedId(downloadsUriProvider.getAllDownloadsUri(), info.getId());
         FileDownloadInfo.ControlStatus.Reader controlReader = new FileDownloadInfo.ControlStatus.Reader(getContentResolver(), downloadUri);
+        DownloadBatch downloadBatch = batchRepository.retrieveBatchFor(info);
         DownloadTask downloadTask = new DownloadTask(
-                this, systemFacade, info, storageManager, downloadNotifier,
-                batchCompletionBroadcaster, batchRepository, downloadsUriProvider,
-                controlReader, networkChecker, downloadReadyChecker, new Clock()
+                this, systemFacade, info, downloadBatch, storageManager,
+                downloadNotifier, batchCompletionBroadcaster, batchRepository,
+                downloadsUriProvider, controlReader, networkChecker, downloadReadyChecker,
+                new Clock()
         );
 
         ContentValues contentValues = new ContentValues();
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadTask.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadTask.java
@@ -78,6 +78,7 @@ class DownloadTask implements Runnable {
 
     private final Context context;
     private final FileDownloadInfo originalDownloadInfo;
+    private final DownloadBatch originalDownloadBatch;
     private final SystemFacade systemFacade;
     private final StorageManager storageManager;
     private final DownloadNotifier downloadNotifier;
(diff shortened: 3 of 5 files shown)


hal9002:
https://ci.novoda.com/job/download-manager-prb/206/
Test FAILed.

hal9002:
https://ci.novoda.com/job/download-manager-prb/207/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadReadyChecker.java
```diff
@@ -23,13 +23,17 @@
 
     public boolean canDownload(DownloadBatch downloadBatch) {
         if (isDownloadManagerReadyToDownload(downloadBatch)) {
-            Download download = downloadMarshaller.marshall(downloadBatch);
-            return downloadClientReadyChecker.isAllowedToDownload(download);
+            clientAllowToDownload(downloadBatch);
```
Electryc: :x: you're missing the `return` here

library/src/main/java/com/novoda/downloadmanager/lib/DownloadReadyChecker.java
```diff
@@ -23,13 +23,17 @@
...
         }
 
         return false;
     }
 
+    boolean clientAllowToDownload(DownloadBatch downloadBatch) {
```
Electryc: Maybe `isClientAllowtoDownload` ?

zegnus: The client is always allowed :P Is the download that might not, what about clientAllowsToDownload(...)

library/src/main/java/com/novoda/downloadmanager/lib/DownloadReadyChecker.java
zegnus: https://github.com/novoda/download-manager/commit/6afd56b44d015b7f4031db6c5165713b98c2463e

library/src/main/java/com/novoda/downloadmanager/lib/DownloadReadyChecker.java
Electryc: Oh I see, maybe yes `clientAllowsToDownload` or `doesClientAllowToDownload`.

hal9002:
https://ci.novoda.com/job/download-manager-prb/208/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadReadyChecker.java
zegnus: https://github.com/novoda/download-manager/commit/965cceff9213aff186ac9dfea13452e00f65d68e

hal9002:
https://ci.novoda.com/job/download-manager-prb/209/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
```diff
@@ -402,10 +402,12 @@ private void downloadOrContinueBatch(List<FileDownloadInfo> downloads) {
     private void download(FileDownloadInfo info) {
         Uri downloadUri = ContentUris.withAppendedId(downloadsUriProvider.getAllDownloadsUri(), info.getId());
         FileDownloadInfo.ControlStatus.Reader controlReader = new FileDownloadInfo.ControlStatus.Reader(getContentResolver(), downloadUri);
+        DownloadBatch downloadBatch = batchRepository.retrieveBatchFor(info);
```
Electryc: So this is called every time the service is notified, but I think we cannot get around it (performance wise).

zegnus: This is just called once, when we add a new download :)

Electryc: Yes, you're right :+1:

Electryc:
LGTM :+1:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadTask.java
```diff
@@ -432,13 +430,19 @@ private void executeDownload(State state) throws StopRequestException {
      * Check if the download has been paused or canceled, stopping the request appropriately if it
      * has been.
      */
-    private void checkPausedOrCanceled() throws StopRequestException {
+    private void checkDownloadCanProceed() throws StopRequestException {
```
ouchadam: :+1:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadReadyChecker.java
```diff
@@ -23,13 +23,17 @@
...
         if (isDownloadManagerReadyToDownload(downloadBatch)) {
-            Download download = downloadMarshaller.marshall(downloadBatch);
-            return downloadClientReadyChecker.isAllowedToDownload(download);
+            return clientAllowsToDownload(downloadBatch);
         }
 
         return false;
     }
 
+    boolean clientAllowsToDownload(DownloadBatch downloadBatch) {
```
blundell: why is this package default protected - don't say to make it testable.

zegnus: Because it does not need to be public :)
Used in the `DownloadTask` class in the same package

ouchadam:
:heavy_multiplication_x: soooo I don't think the client should know about this, it should react to events and force a pause/cancel

zegnus:
@ouchadam what do you mean? The client doesn't change its knowledge at all.

library/src/test/java/com/novoda/downloadmanager/lib/DownloadReadyCheckerTest.java
```diff
@@ -32,7 +32,7 @@ public void givenClientIsReadyToDownloadThenStartDownload() {
                 mock(PublicFacingDownloadMarshaller.class)
         );
 
-        boolean isReady = downloadReadyChecker.canDownload(mock(DownloadBatch.class));
+        boolean isReady = downloadReadyChecker.clientAllowsToDownload(mock(DownloadBatch.class));
```
blundell: why did this change?

zegnus: Because now we can test directly what the test class says, the ready checker

ouchadam:
it triggers the `canDownload` check client side on every chunk of data. As a client of this lib I wouldn't expect the canDownload to be triggered more than once for a single download

zegnus:
@ouchadam let's have this conversation offline

ouchadam:
:+1: had a chat and for now this is the best way, eventually we'll move to have a `COLUMN_REASON` for the status and have the DLM agnostic

library/src/main/java/com/novoda/downloadmanager/lib/DownloadTask.java
```diff
@@ -449,6 +453,12 @@ private void checkPausedOrCanceled() throws StopRequestException {
         }
     }
 
+    private void checkClientRules() throws StopRequestException {
+        if (!downloadReadyChecker.clientAllowsToDownload(originalDownloadBatch)) {
```
ouchadam: I've been thinking more about this and it make less sense, and in fact it's lying to the client.

We'll be triggering rule checks on an invalidated batch. The original batch no longer represents the same batch as the one we have in progress.  

:x:

zegnus: Can you please elaborate what do you mean that the batch is an invalidated?

I really don't see any difference between this and pausing or canceling a batch that is in progress. The reason that I'm saying this is because every second we query the database for that particular batch looking for a canceled or paused status.

That paused or canceled status comes from a user action -> modify a flag in the database -> the check is done on write -> an exception is thrown -> the batch is paused or canceled

The rule check is conceptually exactly the same as pausing, the only difference is that in can be paused by several reasons, not only user action.

The only difference is that instead of querying the database and looking for a paused status, we ask a client if that batch has to be paused. The rest is exactly the same.

If you could please explain why the paused mechanism is ok but the paused by client is not?

We agreed that a possible alternative was to change how we do the check. Instead of through the client callback proceed along with the paused status, querying the database.

As you've put a blocker for this task it would be really useful if you could explain your alternative solution in order to help to unblock it :)

ouchadam: The problem is that the batch being sent to the client is incorrect, it's total size and current size are the original values instead of the current ones.

the "easy" fix is to requery for the batch before checking it but the correct fix would be to avoid doing this check like we mentioned and relying on a `REASON` column!

zegnus: Oh I see, with your little more explanation now is clearer. 

That is correct, we send the original one. 

Now that you mention it, if a client checks for free space and we don't do that while downloading. If another application is downloading something, we would potentially continue until consuming all free space. Is that correct? 

I'm not sure of querying the full object as for performance impact. 

Considering my precios question, a client could check free space before downloading. If that was the case, not querying the full batch every time wouldn't change current behaviour. So we could go as we are now and as agreed move directly to the new column implementation.

PR #149, opened 2015-08-10
Use a larger buffer to allow for less often write to disk

Dorvaryn:
When using the tar truncator the buffer used for the data transfer is really small (512  bytes) to match the tar format buffer. This allows for the tuncation but creates writes to disk too often.

This uses a double buffer in memory to allow for a buffer of the same size than the original data transfer to be used making sure there is no more writes to disk than needed.

### Before
```
Start at 1439216120380
Stop at 1439216120823
Total 443ms
```

### After
```
Start at 1439216197991
Stop at 1439216198146
Total 155ms
```

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/Constants.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/Constants.java
@@ -123,9 +123,14 @@ class Constants {
     public static final String MIMETYPE_APK = "application/vnd.android.package";
 
     /**
+     * The size of a tar block
+     */
+    public static final int BLOCK_SIZE = 512;
+
+    /**
      * The buffer size used to stream the data
      */
-    public static final int BUFFER_SIZE = 4096;
+    public static final int BUFFER_SIZE = 8 * BLOCK_SIZE;
 
     /**
      * The value representing the end of stream when, reading an InputStream
--- a/library/src/main/java/com/novoda/downloadmanager/lib/Constants.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/Constants.java
@@ -125,12 +125,12 @@ class Constants {
     /**
      * The size of a tar block
      */
-    public static final int BLOCK_SIZE = 512;
+    public static final int TAR_BLOCK_SIZE = 512;
 
     /**
      * The buffer size used to stream the data
      */
-    public static final int BUFFER_SIZE = 8 * BLOCK_SIZE;
+    public static final int BUFFER_SIZE = 8 * TAR_BLOCK_SIZE;
 
     /**
      * The value representing the end of stream when, reading an InputStream
--- a/library/src/main/java/com/novoda/downloadmanager/lib/TarTruncator.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/TarTruncator.java
@@ -6,7 +6,6 @@ import java.io.InputStream;
 class TarTruncator implements DataTransferer {
 
     private static final byte BYTE_ZERO = 0x0;
-    private static final int BLOCK_SIZE = 512;
 
     private final DataWriter dataWriter;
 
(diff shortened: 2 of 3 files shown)


hal9002:
https://ci.novoda.com/job/download-manager-prb/211/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/Constants.java
```diff
@@ -123,9 +123,14 @@
     public static final String MIMETYPE_APK = "application/vnd.android.package";
 
     /**
+     * The size of a tar block
+     */
+    public static final int BLOCK_SIZE = 512;
```
blundell: to match the jdoc should this be `TAR_BLOCK_SIZE` or drop the tar in the jdoc

blundell:
:+1:

library/src/main/java/com/novoda/downloadmanager/lib/Constants.java
rock3r: +1 for `TAR_BLOCK_SIZE`

library/src/main/java/com/novoda/downloadmanager/lib/Constants.java
```diff
@@ -123,9 +123,14 @@
...
+
+    /**
      * The buffer size used to stream the data
      */
-    public static final int BUFFER_SIZE = 4096;
+    public static final int BUFFER_SIZE = 8 * BLOCK_SIZE;
```
rock3r: I'm curious, why not 8k? Has this been picked after testing or is it just a random value?

library/src/main/java/com/novoda/downloadmanager/lib/TarTruncator.java
```diff
@@ -18,11 +17,10 @@ public TarTruncator(DataWriter dataWriter) {
     public DownloadTask.State transferData(DownloadTask.State state, InputStream in) throws StopRequestException {
         DownloadTask.State newState = state;
         try {
-            byte[] buffer = new byte[BLOCK_SIZE];
-            int readLast = readBlock(in, buffer);
-            while (readsData(readLast) && isNotFullOfZeroes(buffer)) {
+            byte[] buffer = new byte[Constants.BUFFER_SIZE];
+            int readLast;
+            while ((readLast = readBuffer(in, buffer)) > 0 && isNotFullOfZeroes(buffer)) {
```
rock3r: Uggh assignments in loop conditions :/

library/src/test/java/com/novoda/downloadmanager/lib/TarTruncatorTest.java
```diff
@@ -28,7 +28,7 @@
     @Test
     public void itTruncatesProperlyTarFileThatContainsEndBlockMarker() throws Exception {
         givenATarFileWithEndBlockMarker();
-
+
```
rock3r: ... or not

library/src/main/java/com/novoda/downloadmanager/lib/Constants.java
amlcurran: To match the page size, which is generally 4K http://stackoverflow.com/questions/13351907/android-ram-page-size

rock3r: Ah I see. Nice! I usually go for two pages (8k) when doing disk I/O, but there's little difference I think.

library/src/main/java/com/novoda/downloadmanager/lib/Constants.java
Dorvaryn: :+1: Will change

library/src/main/java/com/novoda/downloadmanager/lib/TarTruncator.java
Dorvaryn: Yep not nice but this is one of the rare cases where I allow myself to use it. The other options all have more trouble dealing with reading + copy of the bytes only when matching conditions.

hal9002:
https://ci.novoda.com/job/download-manager-prb/212/
Test PASSed.

rock3r:
LGTM :+1:

library/src/main/java/com/novoda/downloadmanager/lib/TarTruncator.java
stefanhoth: Adding @charroch just because I can.

charroch: I take my preference with @Dorvaryn here

PR #150, opened 2015-08-19
Add hide intent to cancelled downloads

ronocod:
In a situation where we had 2 downloads, one in progress and one cancelled, swiping away the cancelled one didn't hide it permanently.

This was because we weren't adding the hide Intent to cancelled notifications.

This PR adds the hide intent to cancelled notifications and adds some boy scout refactoring.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/NotificationDisplayer.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/NotificationDisplayer.java
@@ -88,6 +88,7 @@ class NotificationDisplayer {
     }
 
     private void buildActionIntents(String tag, int type, Collection<DownloadBatch> cluster, NotificationCompat.Builder builder) {
+        DownloadBatch batch = cluster.iterator().next();
         if (type == DownloadNotifier.TYPE_ACTIVE || type == DownloadNotifier.TYPE_WAITING) {
             // build a synthetic uri for intent identification purposes
             Uri uri = new Uri.Builder().scheme("active-dl").appendPath(tag).build();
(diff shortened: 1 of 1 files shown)


hal9002:
https://ci.novoda.com/job/download-manager-prb/213/
Test PASSed.

library/src/main/java/com/novoda/downloadmanager/lib/NotificationDisplayer.java
```diff
@@ -96,34 +97,30 @@ private void buildActionIntents(String tag, int type, Collection<DownloadBatch>
...
-            DownloadBatch batch = cluster.iterator().next();
             Intent cancelIntent = new Intent(Constants.ACTION_CANCEL, null, context, DownloadReceiver.class);
             cancelIntent.putExtra(DownloadReceiver.EXTRA_BATCH_ID, batch.getBatchId());
             PendingIntent pendingCancelIntent = PendingIntent.getBroadcast(context, 0, cancelIntent, PendingIntent.FLAG_UPDATE_CURRENT);
             builder.addAction(R.drawable.dl__ic_action_cancel, context.getString(R.string.dl__cancel), pendingCancelIntent);
 
-        } else if (type == DownloadNotifier.TYPE_SUCCESS) {
-            DownloadBatch batch = cluster.iterator().next();
+        } else if (type == DownloadNotifier.TYPE_SUCCESS || type == DownloadNotifier.TYPE_CANCELLED) {
             // TODO: Decide how we handle notification clicks
```
friedger: What is left for this todo?

friedger:
Looks good

library/src/main/java/com/novoda/downloadmanager/lib/NotificationDisplayer.java
ronocod: I think it's related to the `clickIntent` which is now on Line `119`. If there's no extra default behaviour that we want to add here then the comment can probably be removed

PR #155, opened 2015-09-18
API to modify notifications

amlcurran:
This PR begins an API for clients to modify the notification shown when downloading.

Currently the only functionality available is to be able to customise the actions on a downloading notification:

added-action

Also fixed the functionality for clicking on a download notification and it doing something, which (I think) was previously broken, or at least didn't work well.


demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DemoNotificationCustomiser.java
```diff
@@ -0,0 +1,32 @@
...
+        this.context = context;
+    }
+
+    @Override
+    public void addActionsForBatch(NotificationCompat.Builder builder, long batchId) {
+        Intent intent = new Intent(context, MainActivity.class);
+        PendingIntent pendingIntent = PendingIntent.getActivity(context, 0, intent, PendingIntent.FLAG_UPDATE_CURRENT);
+        NotificationCompat.Action action = new NotificationCompat.Action(R.drawable.ic_notif_info, "View", pendingIntent);
+        builder.addAction(action);
+        Intent cancelIntent = new Intent(DownloadManager.ACTION_CANCEL, null, context, DownloadReceiver.class);
```
amlcurran: wondering about whether we can expose a static factory method on the DownloadManager class to hide all this implementation

zegnus: Yeah that might actually be a good idea, specially hiding the next `DownloadManager.` one :)

zegnus: From the client point of view, how do we know which actions are available?

amlcurran: we don't :/ but most are hidden, I'll investigate though :)

amlcurran: Only really seems to be this one that we need to expose for now at least!

amlcurran: https://github.com/novoda/download-manager/commit/bef5f6bf9170081692c4363683a70d6e121919cc :)

zegnus: :+1:

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/NotificationReceiver.java
```diff
@@ -0,0 +1,14 @@
...
+
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.Intent;
+
+public class NotificationReceiver extends BroadcastReceiver {
+
+    @Override
+    public void onReceive(Context context, Intent intent) {
+        context.startActivity(new Intent(context, MainActivity.class).addFlags(Intent.FLAG_ACTIVITY_NEW_TASK));
```
zegnus: Should be check the kind of action received here?

amlcurran: it's only the demo app so probably not worth it, keepin' it simple :)

zegnus: Yes but as the action is hidden now, we could use the demo as an example of how to use this. What do you think?

amlcurran: The action isn't hidden (because you still need it to do checks like the one you mention), just there is a convenience method now

library/src/main/java/com/novoda/downloadmanager/lib/DefaultNotficationCustomiser.java
```diff
@@ -0,0 +1,25 @@
...
+class DefaultNotficationCustomiser implements NotificationCustomiser {
+
+    private final Context context;
+
+    public DefaultNotficationCustomiser(Context context) {
+        this.context = context;
+    }
+
+    @Override
+    public void addActionsForBatch(NotificationCompat.Builder builder, long batchId) {
```
blundell: of course you're assuming the only thing clients are going to do to that builder is "addAction" :-)

blundell:
seems legit :+1:

library/src/main/java/com/novoda/downloadmanager/lib/DefaultNotficationCustomiser.java
amlcurran: Good point! Should we do something or assume that people will be well behaved? :)

blundell: well we either wrap builder and control what they change which is a lot of effort, or for now leave it or rename `addActionsForBatch` to `modifyNotiicationForBatch`? not ideal

amlcurran: lets go with `modify` for now, I like the idea of wrapping the builder but we'll be changing this tons in the next couple of weeks :+1:

amlcurran: https://github.com/novoda/download-manager/commit/17daa52db6a1c57cb36a3839e44ff9d0b3a5f1c8

blundell:
If ferran has further comment it can be in another PR :+1:

PR #156, opened 2015-09-29
Adds a method to exclude deleted batches from query

genypaul:
This PR adds a method to not take into account the batches that have been marked as `DELETED`.
It allows an app to be notified upon deletion right away.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/BatchQuery.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/BatchQuery.java
@@ -221,29 +221,45 @@ public class BatchQuery {
         }
 
         public BatchQuery build() {
-            if (criteriaIdBuilder != null) {
-                builder.withInnerCriteria(criteriaIdBuilder.build());
-                if (criteriaStatusBuilder != null) {
-                    builder.and();
-                }
-            }
 
-            if (criteriaStatusBuilder != null) {
-                builder.withInnerCriteria(criteriaStatusBuilder.build());
-                if (criteriaExtraDataBuilder != null) {
-                    builder.and();
-                }
-            }
-
-            if (criteriaExtraDataBuilder != null) {
-                builder.withInnerCriteria(criteriaExtraDataBuilder.build());
-            }
+            List<Criteria.Builder> criteriaBuilders = setCriteriaBuilders();
+            setCriteriaListFrom(criteriaBuilders);
 
             Criteria criteria = builder.build();
             String selection = criteria.getSelection();
             String sortOrder = criteria.getSort();
             String[] selectionArguments = criteria.getSelectionArguments();
             return new BatchQuery(selection, selectionArguments, sortOrder);
+        }
+
+        private void setCriteriaListFrom(List<Criteria.Builder> criteriaBuilders) {
+            for (Criteria.Builder criteriaBuilder : criteriaBuilders){
+                builder.withInnerCriteria(criteriaBuilder.build());
+                if (isNotLast(criteriaBuilder, criteriaBuilders)) {
+                    builder.and();
+                }
+            }
+        }
+
+        private boolean isNotLast(Criteria.Builder criteriaBuilder, List<Criteria.Builder> criteriaBuilders) {
+            return criteriaBuilders.indexOf(criteriaBuilder) != (criteriaBuilders.size() - 1);
+        }
+
+        private List<Criteria.Builder> setCriteriaBuilders() {
+            List<Criteria.Builder> criteriaBuilderList = new ArrayList<>();
+            if (criteriaIdBuilder != null) {
+                criteriaBuilderList.add(criteriaIdBuilder);
+            }
+            if (criteriaStatusBuilder != null) {
+                criteriaBuilderList.add(criteriaStatusBuilder);
+            }
+            if (criteriaExtraDataBuilder != null) {
+                criteriaBuilderList.add(criteriaExtraDataBuilder);
+            }
+            if (criteriaNoDeletionBuilder != null) {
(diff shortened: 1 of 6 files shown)


hal9002:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadProvider.java
```diff
@@ -800,7 +800,9 @@ public int update(final Uri uri, final ContentValues values, final String where,
             case BATCHES_ID:
                 SqlSelection batchSelection = getWhereClause(uri, where, whereArgs, match);
                 count = db.update(DownloadContract.Batches.BATCHES_TABLE_NAME, values, batchSelection.getSelection(), batchSelection.getParameters());
-                notifyStatusIfBatchesStatusChanged(values);
+                if (values.containsKey(DownloadContract.Batches.COLUMN_STATUS) || values.containsKey(DownloadContract.Batches.COLUMN_DELETED)){
```
zegnus: That is what triggers the immediate notification :)

library/src/main/java/com/novoda/downloadmanager/lib/DownloadProvider.java
```diff
@@ -520,7 +520,7 @@ private void notifyDownloadStatusChanged() {
         getContext().getContentResolver().notifyChange(downloadsUriProvider.getDownloadsWithoutProgressUri(), null);
     }
 
-    private void notifyBatchesStatusChanged() {
+    private void notifyBatchesChanged() {
```
amlcurran: can we rename this method to make it explicit it excludes progress?

library/src/main/java/com/novoda/downloadmanager/lib/BatchQuery.java
```diff
@@ -15,6 +15,7 @@
     public static final BatchQuery ALL = new BatchQuery(null, null, null);
     private static final int LOW_END_FAILED_STATUS_CODE = 400;
     private static final int HIGH_END_FAILED_STATUS_CODE = 600;
+    private static final int BATCH_DELETED = 1;
```
amlcurran: Could move this into DownloadsContract.Batches maybe?

library/src/main/java/com/novoda/downloadmanager/lib/BatchQuery.java
```diff
@@ -211,29 +221,45 @@ public Builder withExtraData(String extraData) {
...
+            setCriteriaListFrom(criteriaBuilders);
+
+            Criteria criteria = builder.build();
+            String selection = criteria.getSelection();
+            String sortOrder = criteria.getSort();
+            String[] selectionArguments = criteria.getSelectionArguments();
+            return new BatchQuery(selection, selectionArguments, sortOrder);
+        }
+
+        private void setCriteriaListFrom(List<Criteria.Builder> criteriaBuilders) {
```
amlcurran: are there any tests around setting multiple criteria? Looks complex and would be good to make sure we don't break it :)

library/src/main/java/com/novoda/downloadmanager/lib/BatchQuery.java
genypaul: It seems like a good place indeed :)

library/src/main/java/com/novoda/downloadmanager/lib/BatchQuery.java
genypaul: There are some tests around query building in `BatchQueryTest`. But I did not see any with multiple criteria other than multiple ids + multiple statuses. I could maybe improve it to take into account other multiple criteria.

amlcurran: OK, well if those tests haven't broken (they haven't, the PR is still open!) then that's probably OK :)

library/src/main/java/com/novoda/downloadmanager/lib/DownloadProvider.java
genypaul: Actually. I'm more wondering why we only notify the `withoutProgress` uri and not the `allBatches` uri as well? Is there a reason

amlcurran: No you're right - we should notify both :+1:

genypaul: https://github.com/novoda/download-manager/commit/b36b53ce71b6f4c28d307e1ee969832f97934199

amlcurran:
LGTM :+1: but need more eyes

library/src/main/java/com/novoda/downloadmanager/lib/BatchQuery.java
```diff
@@ -211,29 +220,45 @@ public Builder withExtraData(String extraData) {
...
                     builder.and();
                 }
             }
+        }
 
+        private boolean isNotLast(Criteria.Builder criteriaBuilder, List<Criteria.Builder> criteriaBuilders) {
+            return criteriaBuilders.indexOf(criteriaBuilder) != (criteriaBuilders.size() - 1);
+        }
+
+        private List<Criteria.Builder> setCriteriaBuilders() {
```
blundell: `combineCriteriaBuilders` ?

blundell:
:+1:

library/src/main/java/com/novoda/downloadmanager/lib/BatchQuery.java
genypaul: Better. Too much `set*` and `build*` :smirk:
https://github.com/novoda/download-manager/commit/7f71c4f5c16578217f2a5382649e105575fa71cd

PR #157, opened 2015-10-02
Add a broadcast for batch failure

flo-genymobile:
### Broadcast batch failure 

The broadcast is only done in case of error. (won't be triggered if a download is cancelled for example)

* Rename the BatchCompletionBroadcaster into BatchTerminationBroadcaster
* extend its API to send a broadcast if a failure occured while downloading.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/BatchTerminationBroadcaster.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/BatchTerminationBroadcaster.java
@@ -3,21 +3,30 @@ package com.novoda.downloadmanager.lib;
 import android.content.Context;
 import android.content.Intent;
 
-class BatchCompletionBroadcaster {
+class BatchTerminationBroadcaster {
 
     static final String ACTION_BATCH_COMPLETE = "com.novoda.downloadmanager.action.BATCH_COMPLETE";
+    static final String ACTION_BATCH_FAILED = "com.novoda.downloadmanager.action.BATCH_FAILED";
+
     static final String EXTRA_BATCH_ID = DownloadReceiver.EXTRA_BATCH_ID;
 
     private final Context context;
     private final String packageName;
 
-    BatchCompletionBroadcaster(Context context, String packageName) {
+    BatchTerminationBroadcaster(Context context, String packageName) {
         this.context = context;
         this.packageName = packageName;
     }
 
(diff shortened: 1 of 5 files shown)


hal9002:

amlcurran:
LGTM :+1:

ronocod:
LGTM :+1:

zegnus:
wait before merging let me see :P

ouchadam:
seems legit, leaving to @zegnus :+1:

zegnus:
This is based on this check

```
/**
     * Returns whether the status is an error (i.e. 4xx or 5xx).
     */
    public static boolean isError(int status) {
        return (status >= 400 && status < 600);
    }
```

`UNKNOWN_ERROR = 491;` so great for this case but
`CANCELED = 490`

Do we consider canceled as a failure?

ouchadam:
hmm I think it should be it's own status

flo-genymobile:
i've kept the implementation *as it was*: 

```
if (DownloadStatus.isCancelled(batchStatus)) {
             batchRepository.setBatchItemsCancelled(batchId);
} else if (DownloadStatus.isError(batchStatus)) {
             batchRepository.setBatchItemsFailed(batchId, downloadId);
            batchTerminationBroadcaster.notifyBatchFailedFor(batchId);
} else if (DownloadStatus.isSuccess(batchStatus)) {
            batchCompletionBroadcaster.notifyBatchCompletedFor(batchId);
            batchTerminationBroadcaster.notifyBatchCompletedFor(batchId);
}
```

the check for cancel is done before the check for error. I didn't want to change the behavior. I am simply triggering the broadcast when we endup in `isError`.

zegnus:
Oh ok so this works if we keep this exact order of execution, but what I'm saying is that onError() will trigger a failure broadcast and that includes the cancel status

flo-genymobile:
what `onError` are you talking about?

zegnus:
Sorry, this check `DownloadStatus.isError(batchStatus)`

zegnus:
Solved this offline, we'll create a new `isFailure` status removing the canceled as a failure.

flo-genymobile:
ok, as suggested by @zegnus, creating a new `isFailure` method that would go like `return isError() && !isCancel()` and call it instead of `isError` in this method. this way we make sure that the behavior doesnt rely on the order in which methods are called

friedger:
As discussed

PR #158, opened 2015-10-08
Track batch first time start

flo-genymobile:
### Track and broacast when a batch starts for the first time

Need to be able to detect when a batch starts for the first time so that external app can react on this. 

* add a new column to the Batch table, default to zero
* Up the DB version to 2 and handle migration

* when a batch download starts, check the column
 * if value is zero, mark batch has started and broadcast an event

* expose the broadcast ACTION so that app using the lib can listen for it

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
@@ -132,6 +132,7 @@ class BatchRepository {
         boolean hasCompleteItems = statusCountMap.hasCountFor(DownloadStatus.SUCCESS);
         boolean hasSubmittedItems = statusCountMap.hasCountFor(DownloadStatus.SUBMITTED);
         boolean hasOtherItems = statusCountMap.hasNoItemsWithStatuses(STATUSES_EXCEPT_SUCCESS_SUBMITTED);
+
         if (hasCompleteItems && hasSubmittedItems && !hasOtherItems) {
             return DownloadStatus.RUNNING;
         }
@@ -143,6 +144,33 @@ class BatchRepository {
         }
 
         return DownloadStatus.UNKNOWN_ERROR;
+    }
+
+    boolean isBatchDownloadStartingForTheFirstTime(long batchId) {
+        Cursor cursor = null;
+        int hasStarted = 0;
+        try {
+            String[] projection = {DownloadContract.Batches.COLUMN_HAS_STARTED};
+            String[] selectionArgs = {String.valueOf(batchId)};
+
+            cursor = resolver.query(
+                    downloadsUriProvider.getBatchesUri(),
+                    projection,
+                    DownloadContract.Batches._ID + " = ?",
+                    selectionArgs,
+                    null
+            );
+
+            if (cursor.moveToFirst()) {
+                hasStarted = cursor.getInt(0);
+            }
+        } finally {
+            if (cursor != null) {
+                cursor.close();
+            }
+        }
+
+        return hasStarted == DownloadContract.Batches.BATCH_HAS_STARTED ? false : true;
     }
 
     public DownloadBatch retrieveBatchFor(FileDownloadInfo download) {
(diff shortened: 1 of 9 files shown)


hal9002:

library/src/main/java/com/novoda/downloadmanager/lib/DatabaseHelper.java
```diff
@@ -108,7 +108,17 @@ public void onCreate(@NonNull SQLiteDatabase db) {
      */
     @Override
     public void onUpgrade(@NonNull SQLiteDatabase db, int oldVersion, final int newVersion) {
-        // no upgrade path yet
+        if (oldVersion == 1 && newVersion == 2) {
+            try {
+                db.execSQL(
+                        "ALTER TABLE " + DownloadContract.Batches.BATCHES_TABLE_NAME
```
Electryc: This SQL can be extracted into a meaningful constant

library/src/main/java/com/novoda/downloadmanager/lib/DatabaseHelper.java
```diff
@@ -108,7 +108,17 @@ public void onCreate(@NonNull SQLiteDatabase db) {
...
+                                + "ADD " + DownloadContract.Batches.COLUMN_HAS_STARTED + " BOOLEAN NOT NULL DEFAULT 0;"
+                );
+            } catch (SQLException ex) {
+                LLog.e("couldn't update table in downloads database");
```
Electryc: What's with this `LLog`?

Electryc:
Can we include an example of how this is used in the sample app?

library/src/main/java/com/novoda/downloadmanager/lib/DatabaseHelper.java
```diff
@@ -108,7 +113,18 @@ public void onCreate(@NonNull SQLiteDatabase db) {
      */
     @Override
     public void onUpgrade(@NonNull SQLiteDatabase db, int oldVersion, final int newVersion) {
-        // no upgrade path yet
+        if (oldVersion == 1 && newVersion == 2) {
+            upgradeFromVersionOneToVersionTwo(db);
+        }
```
flo-genymobile: I've up the DB_VERSION to 2 and this is how the migration is done. 
Shout if you see something is missing here since this is quite critical

xrigau:
:+1: seems legit

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
```diff
@@ -145,6 +146,33 @@ int calculateBatchStatus(long batchId) {
...
+            if (cursor.moveToFirst()) {
+                hasStarted = cursor.getInt(0);
+            }
+        } finally {
+            if (cursor != null) {
+                cursor.close();
+            }
+        }
+
+        return hasStarted == DownloadContract.Batches.BATCH_HAS_STARTED ? false : true;
```
ronocod: Can simplify this to `return hasStarted != DownloadContract.Batches.BATCH_HAS_STARTED;`

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
```diff
@@ -350,6 +350,11 @@ private boolean updateLocked() {
             }
 
             if (!isActive && downloadReadyChecker.canDownload(downloadBatch)) {
+                boolean isBatchStartingForTheFirstTime = batchRepository.isBatchStartingForTheFirstTime(downloadBatch.getBatchId());
```
zegnus: This will hit heavily the db, extract so that we only do this check once

zegnus: Or check if this is only happening once per downloadBatch :)

flo-genymobile: i've put this check here cause it will be triggered only once per batch. so the impact should be minor.

library/src/main/java/com/novoda/downloadmanager/lib/BatchRepository.java
flo-genymobile: :+1:

flo-genymobile:
@zegnus @ronocod @Electryc comments have been addressed

zegnus:
LGTM

ronocod:
:sheep:

PR #159, opened 2015-10-13
Send broadcast when the download was cancelled as well as successful

xrigau:
Title ^

The code modified is inside another `if-else` that says if the download is`TYPE_SUCCESS || TYPE_CANCELLED` then create the notification, but then we would only fire a broadcast event for `SUCCESS`, now we do both.

Before | After
--- | ---
before | after

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
@@ -1099,8 +1099,7 @@ public class DownloadManager {
         }
 
         private long getErrorCode(int status) {
-            if (400 <= status && status < DownloadStatus.MIN_ARTIFICIAL_ERROR_STATUS
-                    || 500 <= status && status < 600) {
+            if (isHttpClientError(status) || isHttpServerError(status)) {
                 // HTTP status code
                 return status;
             }
@@ -1134,6 +1133,14 @@ public class DownloadManager {
                 default:
                     return ERROR_UNKNOWN;
             }
+        }
+
+        private boolean isHttpClientError(int status) {
+            return 400 <= status && status < DownloadStatus.MIN_ARTIFICIAL_ERROR_STATUS;
+        }
+
+        private boolean isHttpServerError(int status) {
+            return 500 <= status && status < 600;
         }
 
         private int translateStatus(int status) {
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
@@ -871,7 +871,7 @@ public class DownloadManager {
                 .setTitle(title)
                 .setDescription(description)
                 .setMimeType(mimeType)
-                .setNotificationVisibility((showNotification) ? NotificationVisibility.ONLY_WHEN_COMPLETE : NotificationVisibility.HIDDEN);
+                .setNotificationVisibility(showNotification ? NotificationVisibility.ONLY_WHEN_COMPLETE : NotificationVisibility.HIDDEN);
 
         if (isMediaScannerScannable) {
             request.allowScanningByMediaScanner();
@@ -1099,8 +1099,8 @@ public class DownloadManager {
         }
 
         private long getErrorCode(int status) {
-            if ((400 <= status && status < DownloadStatus.MIN_ARTIFICIAL_ERROR_STATUS)
-                    || (500 <= status && status < 600)) {
+            if (400 <= status && status < DownloadStatus.MIN_ARTIFICIAL_ERROR_STATUS
+                    || 500 <= status && status < 600) {
                 // HTTP status code
                 return status;
             }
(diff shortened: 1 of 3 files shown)


library/src/main/java/com/novoda/downloadmanager/lib/NotificationDisplayer.java
```diff
@@ -114,17 +114,15 @@ private void buildActionIntents(String tag, int type, Collection<DownloadBatch>
...
-                Intent clickIntent = new Intent(action, uri, context, DownloadReceiver.class);
-                clickIntent.putExtra(DownloadManager.EXTRA_NOTIFICATION_CLICK_DOWNLOAD_IDS, new long[]{batchId});
-                clickIntent.putExtra(DownloadManager.EXTRA_NOTIFICATION_CLICK_DOWNLOAD_STATUSES, new int[]{batchStatus});
-                clickIntent.putExtra(DownloadReceiver.EXTRA_BATCH_ID, batchId);
-                builder.setContentIntent(PendingIntent.getBroadcast(context, 0, clickIntent, PendingIntent.FLAG_UPDATE_CURRENT));
-            }
+            Intent clickIntent = new Intent(action, uri, context, DownloadReceiver.class);
+            clickIntent.putExtra(DownloadManager.EXTRA_NOTIFICATION_CLICK_DOWNLOAD_IDS, new long[]{batchId});
+            clickIntent.putExtra(DownloadManager.EXTRA_NOTIFICATION_CLICK_DOWNLOAD_STATUSES, new int[]{batchStatus});
+            clickIntent.putExtra(DownloadReceiver.EXTRA_BATCH_ID, batchId);
```
flo-genymobile: out of curiosity why do we have `EXTRA_NOTIFICATION_CLICK_DOWNLOAD_IDS` and `EXTRA_BATCH_ID` ? they both seem to have only the `batchId` of the current download

flo-genymobile:
LGTM

library/src/main/java/com/novoda/downloadmanager/lib/NotificationDisplayer.java
```diff
@@ -114,17 +114,15 @@ private void buildActionIntents(String tag, int type, Collection<DownloadBatch>
             hideIntent.putExtra(DownloadReceiver.EXTRA_BATCH_ID, batchId);
             builder.setDeleteIntent(PendingIntent.getBroadcast(context, 0, hideIntent, 0));
 
-            if (type == DownloadNotifier.TYPE_SUCCESS) {
```
flo-genymobile: so this means that now we do this for any `TYPE`. Is that what we want? OR should you extend the if condition instead of removing it?

ouchadam:
would be really really good to have an extra that explains this is from a cancel

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
```diff
@@ -1099,8 +1099,8 @@ private long getPausedReason(int status) {
         }
 
         private long getErrorCode(int status) {
-            if ((400 <= status && status < DownloadStatus.MIN_ARTIFICIAL_ERROR_STATUS)
-                    || (500 <= status && status < 600)) {
+            if (400 <= status && status < DownloadStatus.MIN_ARTIFICIAL_ERROR_STATUS
```
amlcurran: extract methods? no idea what these magic numbers actually mean

library/src/main/java/com/novoda/downloadmanager/lib/NotificationDisplayer.java
xrigau: not sure - I assume some of this is just legacy code from AOSP's download manager

library/src/main/java/com/novoda/downloadmanager/lib/NotificationDisplayer.java
xrigau: from the PR descr.

>The code modified is inside another `if-else` that says if the download is`TYPE_SUCCESS || TYPE_CANCELLED` then create the notification, but then we would only fire a broadcast event for `SUCCESS`, now we do both.

so this only happens for cancelled and success

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
xrigau: lol no idea either but the PRB wouldn't stop closing the PR until I fixed 3 PMD issues so I removed unnecessary parenthesis

xrigau:
@ouchadam the status of the batch in the intent comes as cancelled already

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
amlcurran: worth making the effort to boyscout - if you don't know what it means do you know you've not broken anything :)

xrigau: how's this d81eda1e1ac54dde6c268d56f09885383233c957 ?

library/src/main/java/com/novoda/downloadmanager/lib/NotificationDisplayer.java
flo-genymobile: ok

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManager.java
amlcurran: looks ace

amlcurran:
:+1:

ouchadam:
:+1: lets do it

Issue #85, opened 2015-07-09
Define a Download 


ouchadam:
Currently we have

- `FileDownloadInfo` The uri which is enqueued, given an id and downloaded

- `DownloadBatch` The id that groups FileDownloadInfo ids together

- `Download` The publicly exposed model which exposes the DownloadBatch information (such as total size)

We need better define the difference in these and whether or not the public API should know about a "batch"

PR #162, opened 2015-10-15
Adds more fields to the Download object and passes that to the customiser

### LINKED ISSUE #142: Clicking on notification after download complete crashes the demo-simple app
1. Open the demo-simple app. 
2. Download the podcast
3. Once the download is finished, click on the "Family of Penguins - Download Complete" notification.

I originally saw this in my app and wanted to check if the demo apps behaved any differently.

```
07-27 09:51:17.268  19439-19614/com.novoda.downloadmanager.demo.simple E/AndroidRuntime﹕ FATAL EXCEPTION: DownloadReceiver
    Process: com.novoda.downloadmanager.demo.simple, PID: 19439
    java.lang.SecurityException: Provider com.novoda.downloadmanager.demo.simple/com.novoda.downloadmanager.lib.DownloadProvider does not allow granting of Uri permissions (uri content://com.novoda.downloadmanager.demo.simple/all_downloads/1)
            at android.os.Parcel.readException(Parcel.java:1465)
            at android.os.Parcel.readException(Parcel.java:1419)
            at android.app.ActivityManagerProxy.startActivity(ActivityManagerNative.java:2096)
            at android.app.Instrumentation.execStartActivity(Instrumentation.java:1419)
            at android.app.ContextImpl.startActivity(ContextImpl.java:1047)
            at android.app.ContextImpl.startActivity(ContextImpl.java:1029)
            at android.content.ContextWrapper.startActivity(ContextWrapper.java:311)
            at android.content.ContextWrapper.startActivity(ContextWrapper.java:311)
            at com.novoda.downloadmanager.lib.DownloadReceiver.openDownload(DownloadReceiver.java:159)
            at com.novoda.downloadmanager.lib.DownloadReceiver.handleNotificationBroadcast(DownloadReceiver.java:108)
            at com.novoda.downloadmanager.lib.DownloadReceiver.access$000(DownloadReceiver.java:31)
            at com.novoda.downloadmanager.lib.DownloadReceiver$1.run(DownloadReceiver.java:92)
            at android.os.Handler.handleCallback(Handler.java:733)
            at android.os.Handler.dispatchMessage(Handler.java:95)
            at android.os.Looper.loop(Looper.java:136)
            at android.os.HandlerThread.run(HandlerThread.java:61)
```

xrigau:
Title ^

This was done so that the client can customise the notification depending on the status of the download, set the title, etc.

**EDIT: We've done a bigger refactor so that passing objects to the library is more flexible than it used to be. Before you were forced to implement interfaces in the application's `Application` class, whereas now you can use a `Builder` and just implement 1 interface.
We've also moved many notification-related classes to a `.notifications` package. We couldn't move all of them but at least it's a starting point.**

Also, fixes #142

Paired with @blundell


demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DemoNotificationCustomiser.java
```diff
@@ -18,12 +18,12 @@ public DemoNotificationCustomiser(Context context) {
     }
 
     @Override
-    public void modifyQueuedOrDownloadingNotification(NotificationCompat.Builder builder, long batchId) {
+    public void modifyQueuedOrDownloadingNotification(NotificationCompat.Builder builder, com.novoda.downloadmanager.Download download) {
```
ronocod: Can you use an import here?

xrigau: sadly not, since this is one of the demos and there's a class called `Download` in there already :panda_face:

library/src/main/java/com/novoda/downloadmanager/Download.java
```diff
@@ -3,11 +3,17 @@
...
 
     private final long id;
+    private final String title;
+    private final String description;
+    private final int status;
     private final long currentSize;
     private final long totalSize;
 
-    public Download(long id, long currentSize, long totalSize) {
+    public Download(long id, String title, String description, int status, long currentSize, long totalSize) {
```
amlcurran: why do we need to pass the status? The customiser knows what status the download is right?

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DemoNotificationCustomiser.java
blundell: should the demo class be `BeardDownloadItem` or something of the sorts

library/src/main/java/com/novoda/downloadmanager/Download.java
xrigau: no, he doesn't :-1: - method name says `modifyQueuedOrDownloadingNotification` but if we want to add more (say you want to customise the notification for completed downloads) then we'd have to add more methods which is not ideal IMO - or if you want to differentiate between `Queued` and `Downloading` at the moment you can't

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DemoNotificationCustomiser.java
xrigau: sure :+1:

library/src/main/java/com/novoda/downloadmanager/Download.java
amlcurran: I know, I think it should be split, one method for one customisation I think that's fine (although might have to perhaps consider an abstract class which overrides from the default one)

amlcurran: we shouldn't push logic onto the client as much as possible I reckon :)

library/src/main/java/com/novoda/downloadmanager/lib/DefaultNotficationCustomiser.java
```diff
@@ -16,8 +17,8 @@ public DefaultNotficationCustomiser(Context context) {
     }
 
     @Override
-    public void modifyQueuedOrDownloadingNotification(NotificationCompat.Builder builder, long batchId) {
-        Intent cancelIntent = DownloadManager.createCancelBatchIntent(batchId, context);
+    public void modifyQueuedOrDownloadingNotification(NotificationCompat.Builder builder, Download download) {
```
xrigau: @blundell

blundell: `modifyQueuedNotification`
`modifyDownloadingNotification`
`modifyCompletedNotification` ?

library/src/main/java/com/novoda/downloadmanager/Download.java
blundell: :+1: agr33

library/src/main/java/com/novoda/downloadmanager/lib/NotificationCustomiser.java
```diff
@@ -2,7 +2,9 @@
 
 import android.support.v4.app.NotificationCompat;
 
+import com.novoda.downloadmanager.Download;
+
 public interface NotificationCustomiser {
 
-    void modifyQueuedOrDownloadingNotification(NotificationCompat.Builder builder, long batchId);
+    void modifyQueuedOrDownloadingNotification(NotificationCompat.Builder builder, Download download);
```
flo-genymobile: my understanding is that a `Batch` is composed of several downloads. so passing a `Download` object here is a bit confusing isnt it? cause what you really want is the Batch. 
There is a semantic problem I think, cause when we say **a download** it actually means **a video stream and several ads stream** (and this is represented by a **Batch**) 

iam not even sure this makes any sense :)

xrigau: yeah well this is the DLM library so in order to make it simpler for clients the idea was to leak less impl. details I think. you can discuss more here if you want https://github.com/novoda/download-manager/issues/85 :smiley:

library/src/main/java/com/novoda/downloadmanager/Download.java
xrigau: >we shouldn't push logic onto the client

makes sense! :+1: I'll do

library/src/main/java/com/novoda/downloadmanager/lib/NotificationCustomiser.java
flo-genymobile: right :) sorry for that.

but so it means, that for the client of the lib we only expose the concept of `Download`? 
if thats the case, i couldnt agree more :+1: 

however, this is confusing for the internal of the library but this will be sorted out one day (i guess)

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DemoApplication.java
```diff
@@ -28,10 +32,30 @@ public boolean isAllowedToDownload(Download download) {
...
     }
 
+    @Override
+    public DownloadingNotificationCustomiser getDownloadingNotificationCustomiser() {
+        return notificationCustomiser;
+    }
+
+    @Override
+    public CompleteNotificationCustomiser getCompleteNotificationCustomiser() {
+        return null;
```
xrigau: we are now going to change the public interface so that the user doesn't have to implement these methods if they don't want to i.e. no returning null yayyay

library/src/main/java/com/novoda/downloadmanager/lib/NotificationCustomiser.java
xrigau: yeah that's the idea (even tho it's not 100% there)

xrigau:
closing whilst we refactor

xrigau:
SIK refactoring done

xrigau:
PR description updated

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DemoApplication.java
```diff
@@ -4,14 +4,13 @@
 import android.hardware.SensorManager;
 
 import com.novoda.downloadmanager.Download;
+import com.novoda.downloadmanager.lib.DownloadManagerModules;
 import com.novoda.downloadmanager.lib.DownloadClientReadyChecker;
-import com.novoda.downloadmanager.lib.NotificationCustomiser;
-import com.novoda.downloadmanager.lib.NotificationCustomiserProvider;
 
-public class DemoApplication extends Application implements DownloadClientReadyChecker, NotificationCustomiserProvider {
+public class DemoApplication extends Application implements DownloadManagerModules.Provider {
```
xrigau: now you only need to implement 1 interface

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DemoApplication.java
```diff
@@ -21,23 +20,23 @@ public void onCreate() {
     }
 
     @Override
-    public boolean isAllowedToDownload(Download download) {
-        // Here you would add any reasons you may not want to download
-        // For instance if you have some type of geo-location lock on your download capability
-        return oneRuleToBindThem.shouldWeDownload(download);
+    public DownloadManagerModules provideDownloadManagerModules() {
+        return DownloadManagerModules.Builder.from(this)
```
xrigau: and you use this builder for EZPZ usage

library/src/main/java/com/novoda/downloadmanager/lib/DefaultsDownloadManagerModules.java
```diff
@@ -0,0 +1,114 @@
...
+        this.cancelledCustomiser = cancelledCustomiser;
+        this.failedCustomiser = failedCustomiser;
+        this.readyChecker = readyChecker;
+        this.imageRetriever = imageRetriever;
+    }
+
+    @Override
+    public NotificationImageRetriever getNotificationImageRetriever() {
+        if (imageRetriever == null) {
+            return new OkHttpNotificationImageRetriever();
```
xrigau: we need defaults incase the client doesn't want to provide any of these

library/src/main/java/com/novoda/downloadmanager/lib/NotificationCustomiser.java
```diff
@@ -2,7 +2,55 @@
...
-public interface NotificationCustomiser {
+import com.novoda.downloadmanager.notifications.CancelledNotificationCustomiser;
+import com.novoda.downloadmanager.notifications.CompleteNotificationCustomiser;
+import com.novoda.downloadmanager.Download;
+import com.novoda.downloadmanager.notifications.DownloadingNotificationCustomiser;
+import com.novoda.downloadmanager.notifications.FailedNotificationCustomiser;
+import com.novoda.downloadmanager.notifications.QueuedNotificationCustomiser;
 
-    void modifyQueuedOrDownloadingNotification(NotificationCompat.Builder builder, long batchId);
+class NotificationCustomiser implements QueuedNotificationCustomiser, DownloadingNotificationCustomiser,
```
xrigau: we wanted to keep this class so that we can still pass it in to the notification displayer

library/src/main/java/com/novoda/downloadmanager/lib/NotificationDisplayer.java
```diff
@@ -121,6 +123,31 @@ private void buildActionIntents(String tag, int type, Collection<DownloadBatch>
             Intent clickIntent = createClickIntent(action, batchId, batchStatus, uri);
             builder.setContentIntent(PendingIntent.getBroadcast(context, 0, clickIntent, PendingIntent.FLAG_UPDATE_CURRENT));
         }
+
+        customiseNotification(type, builder, batch);
+    }
+
+    private void customiseNotification(int type, NotificationCompat.Builder builder, DownloadBatch batch) {
```
xrigau: @amlcurran all these methods :100:

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DemoApplication.java
amlcurran: builders are trendy these days huh :tongue:

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DemoNotificationCustomiser.java
```diff
@@ -5,11 +5,13 @@
...
 
+import com.novoda.downloadmanager.Download;
+import com.novoda.downloadmanager.notifications.DownloadingNotificationCustomiser;
+import com.novoda.downloadmanager.notifications.QueuedNotificationCustomiser;
 import com.novoda.downloadmanager.demo.R;
 import com.novoda.downloadmanager.lib.DownloadManager;
-import com.novoda.downloadmanager.lib.NotificationCustomiser;
 
-class DemoNotificationCustomiser implements NotificationCustomiser {
+class DemoNotificationCustomiser implements QueuedNotificationCustomiser, DownloadingNotificationCustomiser {
```
amlcurran: :+1:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManagerModules.java
```diff
@@ -0,0 +1,101 @@
...
+
+    CancelledNotificationCustomiser getCancelledNotificationCustomiser();
+
+    FailedNotificationCustomiser getFailedNotificationCustomiser();
+
+    class Builder {
+
+        private final Context context;
+
+        private QueuedNotificationCustomiser queuedCustomiser;
```
amlcurran: why not initialise these to the sane defaults, and then you don't need the null checks and nullable annotations in the class above

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DemoApplication.java
blundell: I know right, feels like the goto pattern

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManagerModules.java
blundell: then an evil client could pass `null` and crash it right .. which you could argue is there fault.  Some of these can't have sane defaults as they need context in the constructor

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DemoApplication.java
xrigau: specially when the entry point method is called `from` :heart:

blundell:
should fix: https://github.com/novoda/download-manager/issues/142

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManagerModules.java
amlcurran: I would so say it's their fault but I'm mean :imp: but fairy nuff :+1:

library/src/main/java/com/novoda/downloadmanager/lib/NotificationDisplayer.java
amlcurran: :sparkles:

blundell:
@amlcurran loads of thumbs up, not much merging :smiley_cat: any chance?

blundell:
probably best if we merge https://github.com/novoda/download-manager/pull/164 into this first - @amlcurran could you move your eyes over to that pretty please :star2: :ferris_wheel:

xrigau:
now this has become a monster PR :sweat_smile:

library/src/main/java/com/novoda/downloadmanager/lib/DefaultsDownloadManagerModules.java
amlcurran: still think this may be better done in the builder - once you do "builder.build()" I feel like you're supposed to have a fully-instantiated class

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DemoApplication.java
juankysoriano: :+1:

amlcurran:
Looks alright to me, quite difficult to read because of adding new features and refactoring :(

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DemoApplication.java
```diff
@@ -21,23 +20,23 @@ public void onCreate() {
...
 
         /**
          * @return for our demo we expect always to return true ... unless you want to conquer the galaxy
          */
-        public boolean shouldWeDownload(Download download) {
+        @Override
+        public boolean isAllowedToDownload(Download download) {
+            // Here you would add any reasons you may not want to download
+            // For instance if you have some type of geo-location lock on your download capability
             return download.getTotalSize() > SensorManager.GRAVITY_DEATH_STAR_I;
```
juankysoriano: LOL

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DemoApplication.java
zegnus: :+1:

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DemoApplication.java
zegnus: aand a builder with no optional params... :dancer: 

I like it anyway :)

xrigau:
yeah we should've used a `feature-branch` but didn't think we'd do so many changes

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DemoApplication.java
xrigau: :star:

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/batches/BatchDownloadsActivity.java
```diff
@@ -18,7 +18,7 @@
 import com.novoda.downloadmanager.demo.extended.BeardDownload;
 import com.novoda.downloadmanager.demo.extended.QueryForDownloadsAsyncTask;
 import com.novoda.downloadmanager.lib.DownloadManager;
-import com.novoda.downloadmanager.lib.NotificationVisibility;
+import com.novoda.downloadmanager.notifications.NotificationVisibility;
```
zegnus: :+1:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadBatch.java
```diff
@@ -125,4 +149,77 @@ public boolean scanCompletedMediaIfReady(DownloadScanner downloadScanner) {
...
+    public String getTitle() {
+        return info.getTitle();
+    }
+
+    public long getFirstDownloadBatchId() {
+        return downloads.get(0).getId();
+    }
+
+    public Statistics getLiveStatistics(LongSparseArray<Long> downloadSpeed) {
+        long currentBytes = 0;
```
juankysoriano: Aren't this calculations responsibility of `Statistics` ?

amlcurran: Also, did you create this code or move this code? This will result in a ton of allocations D:

blundell: moved it.  It's not done inside Statistics to keep the same behaviour of before i.e. this is calculated once on created and then the values are read. (rather than 3 methods that would do 3 separate loops) this was in PR #164

blundell: this is not ideal I know, but it's one step better than before and I'm happy with that.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadBatch.java
```diff
@@ -46,6 +50,26 @@ public int getStatus() {
         return status;
     }
 
+    public boolean isQueuedForWifi() {
```
zegnus: :+1:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManagerModules.java
juankysoriano: I am with @amlcurran , if the client passes null then we should instead throw a `DeveloperError` but using the default instead of what he's passing is breaking the contract. The `withBlablabla` is a promise we are making to him that we'll be using what he provides, even `null`.

library/src/main/java/com/novoda/downloadmanager/lib/DefaultsDownloadManagerModules.java
blundell: this class is used elsewhere to select the defaults (outside of the builder) and so having it like this removes that duplication of selecting defaults in two places

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManagerModules.java
blundell: if someone passes null we could throw a dev error or annotate the `withBlaBla` with `@notnull`

xrigau: I agree with @blundell - IMO using `@NotNull` would be a good idea

juankysoriano: Yes, that makes the trick. I am happy with that.

xrigau: 95601fd

library/src/main/java/com/novoda/downloadmanager/notifications/NotificationDisplayer.java
```diff
@@ -251,21 +286,21 @@ private Notification buildStackedNotification(
         final NotificationCompat.InboxStyle inboxStyle = new NotificationCompat.InboxStyle(builder);
 
         for (DownloadBatch batch : currentBatches) {
-            inboxStyle.addLine(getDownloadTitle(batch.getInfo()));
+            inboxStyle.addLine(getDownloadTitle(batch));
         }
 
-        if (type == DownloadNotifier.TYPE_ACTIVE) {
```
zegnus: looks like a switch :P

I'm the `switch` man it seems :P

zegnus:
Many changes.... LGTM @amlcurran @juankysoriano ?

library/src/main/java/com/novoda/downloadmanager/notifications/NotificationDisplayer.java
juankysoriano: I am with you on this.

zegnus:
If there are more comments those can be addressed in another PR, merging

juankysoriano:
@zegnus that's what I was about to comment and to do!

PR #163, opened 2015-10-15
Restart stale downloads

biafra23:
### Bug description ###

While a download is in progress, the DownloadManager can be killed. In this situation the database will contain downloads with state `downloading`.

When the DownloadManager starts again the `downloading` ones won't be re-submitted as the system thinks that those are already in progress.

### Solution implemented ###

We have created a helper method that fixes all the states of the downloads (in `DownloadService.onCreate()`) and let the DownloadManager to restart them.

Paired with @zegnus


demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DemoApplication.java
```diff
@@ -18,6 +20,9 @@ public void onCreate() {
         super.onCreate();
         notificationCustomiser = new DemoNotificationCustomiser(DemoApplication.this);
         oneRuleToBindThem = new OneRuleToBindThem();
+
+        DownloadManager downloadManager = DownloadManagerBuilder.from(this).build();
+        downloadManager.sanityCheckBatchStatuses();
```
amlcurran: why not just "restart" or something more simple?

blundell:
isn't this something that should be done in the `DownloadService.onCreate` i.e. not by the client,

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DemoApplication.java
blundell: yeah or `.onCreateApplication()`

ronocod:
:X: there's a DB file in this commit that should not be here :cop:

demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/DemoApplication.java
zegnus: Not `restart` because this does not restart paused by the app downloads or other valid cases, this fixes invalid states of the downloads.

`onCreateApplication` yeah maybe as this should be called at that moment

zegnus:
@blundell We tried there and for some reason wasn't working, I will give it another try though

@ronocod Good catch, removing

library/src/main/java/com/novoda/downloadmanager/lib/DownloadsRepository.java
```diff
@@ -195,6 +198,71 @@ public void setDownloadSubmitted(FileDownloadInfo info) {
         contentResolver.update(info.getAllDownloadsUri(), contentValues, null, null);
     }
 
+    @Nullable
+    String getCurrentDownloadingBatchId() {
+        return getCurrentBatchIdWithColumnStatus(DownloadStatus.RUNNING);
```
xrigau: does this assume there's only 1 download downloading at a time?? what if there are more?

zegnus: Correct, we only support sequential downloads for now.

In case that we want to make parallel downloads in the future this would return an array

xrigau: >we only support sequential downloads for now.

what do you mean? the library supports multiple downloads at a time at the moment, the value can be modified by specifying a value as metadata in the manifest `com.novoda.downloadmanager.MaxConcurrentDownloads`

zegnus: You're right we missed that, we'll fix it

blundell: what Xavi said :x:

zegnus:
Closing to support multiple downloads in parallel

zegnus:
@xrigau @blundell Support for parallel download done :)

flo-genymobile:
:+1: guys this looks good!

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
```diff
@@ -179,6 +175,24 @@ public FileDownloadInfo create(FileDownloadInfo.Reader reader) {
...
         }, downloadsUriProvider);
 
+        unlockStaleDownloads();
+
+        updateThread = new HandlerThread("DownloadManager-UpdateThread");
+        updateThread.start();
+        updateHandler = new Handler(updateThread.getLooper(), updateCallback);
+    }
+
+    void unlockStaleDownloads() {
```
xrigau: why not private?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
```diff
@@ -179,6 +175,24 @@ public FileDownloadInfo create(FileDownloadInfo.Reader reader) {
...
+        List<String> batchesToBeUnlocked = downloadsRepository.getCurrentDownloadingBatchIds();
+
+        if (batchesToBeUnlocked.isEmpty()) {
+            batchesToBeUnlocked = downloadsRepository.getCurrentSubmittedBatchIds();
```
xrigau: are downloading and submitted excluding? I wonder if this should be
```java
List<String> batchesToBeUnlocked = downloadsRepository.getCurrentDownloadingBatchIds();
batchesToBeUnlocked.addAll(downloadsRepository.getCurrentSubmittedBatchIds());
// the rest would be the same
```

zegnus: I don't think that is possible to have one batch with no downloading/completed but submitted ones, so in theory or is submitted or is downloading.

We would have to avoid duplicates in there so instead of a list should be a set because a downloading one will contain submitted entries almost for sure.

Will do because maybe in parallel batches there might be an edge case

library/src/main/java/com/novoda/downloadmanager/lib/DownloadsRepository.java
```diff
@@ -195,6 +199,79 @@ public void setDownloadSubmitted(FileDownloadInfo info) {
         contentResolver.update(info.getAllDownloadsUri(), contentValues, null, null);
     }
 
+    @NonNull
```
blundell: we don't use these annotations elsewhere, lets not start now.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadsRepository.java
```diff
@@ -195,6 +199,79 @@ public void setDownloadSubmitted(FileDownloadInfo info) {
...
+    List<String> getCurrentDownloadingBatchIds() {
```
blundell: why default protection?

zegnus: https://github.com/novoda/download-manager/commit/4109830a9d6bbef156656abfc0a35f4d2e861ae8

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
zegnus: https://github.com/novoda/download-manager/commit/cba7a587a14190535f42002a32a7acddc5ed8978

library/src/main/java/com/novoda/downloadmanager/lib/DownloadsRepository.java
```diff
@@ -195,6 +198,69 @@ public void setDownloadSubmitted(FileDownloadInfo info) {
...
+            if (cursor != null) {
+                cursor.close();
+            }
+        }
+    }
+
+    /**
+     * @return Number of rows updated
+     */
+    int updateRunningOrSubmittedDownloadsToPending() {
```
blundell: same with this ferran why default scope?

xrigau:
Pauls comment and then :shipit:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadsRepository.java
zegnus: https://github.com/novoda/download-manager/commit/9afdcc3980649e1670b6fa884e6468909ff93eb5

PR #164, opened 2015-10-15
Moves more notification code to package

blundell:
Moves behaviour onto the DownloadBatch so that it is not showing its innards

Moves the actions specifically used for notifications to that package.

Renames status translator to the `PublicFacing` pattern, not ideal but matches other code for now.

Removes unused code.

Creates a factory for the `DownloaNotifier` so that we can create it all in 1 package.

Still more to do, refactoring towards a complete notification package, but this is starting to give our download manager more flexibility.

(merging into https://github.com/novoda/download-manager/pull/162)

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadBatch.java
@@ -1,8 +1,12 @@
 package com.novoda.downloadmanager.lib;
 
+import android.support.v4.util.LongSparseArray;
+
+import com.novoda.downloadmanager.notifications.NotificationVisibility;
+
 import java.util.List;
 
-class DownloadBatch {
+public class DownloadBatch {
 
     public static final DownloadBatch DELETED = new DownloadBatch(-1, null, null, -1, -1L, -1L);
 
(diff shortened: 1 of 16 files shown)


blundell:
@xrigau

blundell:
I've noticed my PR description bullets could have been separate commits :-( I am sorry!

library/src/main/java/com/novoda/downloadmanager/lib/DownloadBatch.java
```diff
@@ -125,4 +149,77 @@ public boolean scanCompletedMediaIfReady(DownloadScanner downloadScanner) {
...
+
+    public String getTitle() {
+        return info.getTitle();
+    }
+
+    public long getFirstDownloadBatchId() {
+        return downloads.get(0).getId();
+    }
+
+    public Statistics getLiveStatistics(LongSparseArray<Long> downloadSpeed) {
```
blundell: by moving this to here, I noticed that the `DownloadBatch` already has `final` fields for `currentSizeBytes` and `totalSizeBytes` could these be used here? I'm not sure, didn't want to change that behaviour, but if anyone knows differently?

library/src/test/java/com/novoda/downloadmanager/notifications/DownloadNotifierTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mock;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+import static org.mockito.MockitoAnnotations.initMocks;
+
+public class DownloadNotifierTest {
```
xrigau: :+1:

xrigau: ah just been moved :expressionless:

xrigau:
:+1: nice action

PR #165, opened 2015-10-19
Adds a static text for downloading so that it can be overwritten

xrigau:
If the client doesn't provide a description with the request for a download, it used to show a text with the remaining time, the problem being that it could not be changed from the client (from the customiser).

Using a static resource allows us to change it via resource overriding. This is not the ideal solution but it's good enough for now.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/notifications/NotificationDisplayer.java
+++ b/library/src/main/java/com/novoda/downloadmanager/notifications/NotificationDisplayer.java
@@ -214,7 +214,7 @@ public class NotificationDisplayer {
 
         if (currentBatches.size() == 1) {
             DownloadBatch batch = currentBatches.iterator().next();
-            return buildSingleNotification(type, builder, batch, remainingText, percentText);
+            return buildSingleNotification(type, builder, batch, percentText);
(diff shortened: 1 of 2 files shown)


xrigau:
I've created an issue to improve this #166

ouchadam:
makes sense :+1:

library/src/main/java/com/novoda/downloadmanager/notifications/NotificationDisplayer.java
```diff
@@ -239,7 +239,7 @@ private Notification buildSingleNotification(
         if (type == SynchronisedDownloadNotifier.TYPE_ACTIVE) {
             String description = batch.getDescription();
             if (TextUtils.isEmpty(description)) {
-                setSecondaryNotificationText(builder, style, remainingText);
+                setSecondaryNotificationText(builder, style, context.getString(R.string.dl__downloading));
```
blundell: why didn't you do `context.getString(R.string.dl__remaining, remainingText)`

`<string name="dl__remaining">%s$1</string>`

so as to not break the current behaviour of all other apps? :x:

PR #167, opened 2015-10-20
Handle failed notifications so that they get dismissed and click events work

xrigau:
Title ^

### Before

the notifications for a failed download wouldn't fire any intent on click, and they'd reappear after being dismissed every time the DownloadService was recreated.

### After

the click works and they don't reappear after being dismissed

Fixes #51 , #143

PR #169, opened 2015-10-20
If user clicks on a failed download then we want to hide the notification forever

xrigau:
Title ^

#167 fixes the 'failed download' notification reappearing when the user dismisses it but not when the user clicks on it, so this PR fixes that

Also did small boyscouting for the `batchId`


blundell:
:+1:

blundell:
tests .. soon we should

Issue #168, opened 2015-10-20
Remove the URI from the Intent used for the click events on Notifications


xrigau:
https://github.com/novoda/download-manager/pull/167/files#diff-14235d3c65570f071b2574801705fc2aR128

That Uri seems like it isn't used anywhere so we should make sure it is not and then get rid of it.

xrigau:
Fixed in #170

PR #170, opened 2015-10-20
Uses multiple actions for click events on notifications so they can be handled differently

### LINKED ISSUE #161: Rename `ACTION`s in `Constants` to be more specific and meaningful
Raised by @amlcurran in https://github.com/novoda/download-manager/pull/160/files#r41991443

The Action names are too generic at the moment, they should be more specific, and maybe they could be split in many or joined together (OPEN vs. LIST)

Actions are here: https://github.com/novoda/download-manager/blob/57f9aa/library/src/main/java/com/novoda/downloadmanager/lib/Constants.java#L42-L67

xrigau:
#### Before

We used to have generic actions (inherited from the AOSP) to handle click events - this was confusing and hard to work around it.

#### After

We can differentiate between each specific action and handle them all differently if we have to in the `DownloadReceiver`.
Also solved a bug since we were checking for the batch status but it was the translated status instead of the raw one, this was fixed with just using another action :dancer: 

Fixes #161 and #168

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadReceiver.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadReceiver.java
@@ -65,9 +65,13 @@ public class DownloadReceiver extends BroadcastReceiver { // TODO split this int
             case CONNECTIVITY_ACTION:
                 checkConnectivityToStartService(context);
                 break;
-            case NotificationDisplayer.ACTION_OPEN:
-            case NotificationDisplayer.ACTION_LIST:
-            case NotificationDisplayer.ACTION_HIDE:
+            case NotificationDisplayer.ACTION_DOWNLOAD_CANCELLED_CLICK:
+            case NotificationDisplayer.ACTION_DOWNLOAD_FAILED_CLICK:
+            case NotificationDisplayer.ACTION_DOWNLOAD_SUCCESS_CLICK:
+            case NotificationDisplayer.ACTION_DOWNLOAD_RUNNING_CLICK:
+            case NotificationDisplayer.ACTION_DOWNLOAD_SUBMITTED_CLICK:
+            case NotificationDisplayer.ACTION_DOWNLOAD_OTHER_CLICK:
+            case NotificationDisplayer.ACTION_NOTIFICATION_DISMISSED:
             case ACTION_CANCEL:
                 handleSystemNotificationAction(context, intent);
                 break;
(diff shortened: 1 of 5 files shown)


library/src/main/java/com/novoda/downloadmanager/lib/DownloadReceiver.java
```diff
@@ -102,31 +106,33 @@ public void run() {
...
         long batchId = getBatchId(intent);
+
+        String action = intent.getAction();
         switch (action) {
-            case NotificationDisplayer.ACTION_LIST:
+            case NotificationDisplayer.ACTION_DOWNLOAD_CANCELLED_CLICK:
+            case NotificationDisplayer.ACTION_DOWNLOAD_FAILED_CLICK:
+            case NotificationDisplayer.ACTION_DOWNLOAD_SUCCESS_CLICK:
                 sendNotificationClickedIntent(context, ids, statuses);
-                if (DownloadStatus.isFailure(statuses[0])) {
```
xrigau: this didn't work cause the status was translated - now there's no need to check for the status here, we can just check the action :boom:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadsUriProvider.java
```diff
@@ -2,7 +2,7 @@
 
 import android.net.Uri;
 
-public class DownloadsUriProvider {
+public class DownloadsUriProvider { // Why is this a singleton if all fields are final?
```
xrigau: does anybody know?

library/src/main/java/com/novoda/downloadmanager/notifications/NotificationDisplayer.java
```diff
@@ -116,39 +107,60 @@ private void buildIcon(int type, NotificationCompat.Builder builder) {
...
 
-    private void buildActionIntents(String tag, int type, Collection<DownloadBatch> cluster, NotificationCompat.Builder builder) {
+    private void buildActionIntents(int type, Collection<DownloadBatch> cluster, NotificationCompat.Builder builder) {
         DownloadBatch batch = cluster.iterator().next();
         long batchId = batch.getBatchId();
         int batchStatus = batch.getStatus();
 
         if (type == SynchronisedDownloadNotifier.TYPE_ACTIVE || type == SynchronisedDownloadNotifier.TYPE_WAITING) {
-            // build a synthetic uri for intent identification purposes
-            Uri uri = new Uri.Builder().scheme("active-dl").appendPath(tag).build();
```
xrigau: I've removed this since it wasn't used (addresses the TODO)

library/src/main/java/com/novoda/downloadmanager/notifications/NotificationDisplayer.java
```diff
@@ -116,39 +107,60 @@ private void buildIcon(int type, NotificationCompat.Builder builder) {
...
-            builder.setContentIntent(PendingIntent.getBroadcast(context, 0, clickIntent, PendingIntent.FLAG_UPDATE_CURRENT));
             builder.setOngoing(true);
-
         } else if (type == SynchronisedDownloadNotifier.TYPE_SUCCESS
                 || type == SynchronisedDownloadNotifier.TYPE_CANCELLED
                 || type == SynchronisedDownloadNotifier.TYPE_FAILED) {
-            long firstDownloadBatchId = batch.getFirstDownloadBatchId(); // TODO why can't we just use getBatchId()?
-            Uri uri = ContentUris.withAppendedId(downloadsUriProvider.getAllDownloadsUri(), firstDownloadBatchId);
-
-            Intent hideIntent = new Intent(ACTION_HIDE, uri, context, DownloadReceiver.class);
```
xrigau: extracted to a `createNotificationDismissedIntent` method

library/src/main/java/com/novoda/downloadmanager/notifications/NotificationDisplayer.java
```diff
@@ -116,39 +107,60 @@ private void buildIcon(int type, NotificationCompat.Builder builder) {
...
+            Intent dismissedIntent = createNotificationDismissedIntent(batchId);
+            builder.setDeleteIntent(PendingIntent.getBroadcast(context, 0, dismissedIntent, 0));
             builder.setAutoCancel(true);
-
-            String action = batch.isError() ? ACTION_LIST : ACTION_OPEN;
-            Intent clickIntent = createClickIntent(action, batchId, batchStatus, uri);  // TODO: the Uri doesn't seem to be ever read
-            builder.setContentIntent(PendingIntent.getBroadcast(context, 0, clickIntent, PendingIntent.FLAG_UPDATE_CURRENT));
         }
 
+        Intent clickIntent = createClickIntent(batchId, batchStatus);
```
xrigau: the click intent is needed in all cases I believe, so instead of doing it in the if AND the else, it's done here now

xrigau:
Fixes #161 and #168

library/src/main/java/com/novoda/downloadmanager/lib/DownloadsUriProvider.java
blundell: @zegnus knows

blundell:
makes sense :+1: looks like we could use some strategy pattern here instead of all these switch cases in the future :-)

PR #172, opened 2015-10-22
Use the 'HIDDEN' state when hiding a notification

xrigau:
This prevents notifications from reappearing when you kill the app after dismissing a notification and start the app again.

#### Steps to reproduce

 - Download something
 - When download completes or fails, dismiss the notification by swiping it
 - Kill the app
 - Start the app again

#### Before

 - The notification would reappear

#### After

 - No notification reappears

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadReceiver.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadReceiver.java
@@ -165,7 +165,7 @@ public class DownloadReceiver extends BroadcastReceiver { // TODO split this int
                 if ((DownloadStatus.isCancelled(status) || DownloadStatus.isCompleted(status))
                         && (visibility == ACTIVE_OR_COMPLETE || visibility == ONLY_WHEN_COMPLETE)) {
                     ContentValues values = new ContentValues(1);
-                    values.put(DownloadContract.Batches.COLUMN_VISIBILITY, NotificationVisibility.ONLY_WHEN_ACTIVE);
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/novoda/downloadmanager/lib/DownloadReceiver.java
```diff
@@ -165,7 +165,7 @@ private void hideNotification(Context context, long batchId) {
                 if ((DownloadStatus.isCancelled(status) || DownloadStatus.isCompleted(status))
                         && (visibility == ACTIVE_OR_COMPLETE || visibility == ONLY_WHEN_COMPLETE)) {
                     ContentValues values = new ContentValues(1);
-                    values.put(DownloadContract.Batches.COLUMN_VISIBILITY, NotificationVisibility.ONLY_WHEN_ACTIVE);
```
xrigau: if the method says 'hide' does anybody know why this used this flag instead of the `HIDE` one?

ronocod: `ONLY_WHEN_ACTIVE` will only show when the download is active, for completed downloads either can be used, but I think `HIDDEN` is easier to understand :+1:

xrigau: strange cause it was showing the notification back anyway after dismissing and restarting the download service :scream:

ronocod:
LGTM :+1:

zegnus:
LGTM

xrigau:
Turns out this didn't completely fixed the issue, this does #173

PR #173, opened 2015-10-22
Makes notifications uniques by setting different data on each of them

xrigau:
So it turns out the real issue with notifications reappearing after being dismissed was the intents weren't unique. I thought it had to do with the flags we use or the way we handle the notification visibility in the DB, and maybe it did initially but now the problem was the uniqueness of the intents. So the bug could only be reproduced when having multiple failed/completed notifications, with just one it worked fine, that made it harder to pin point.

Anyway this is fixed by making each intent used for the notifications unique as mentioned here http://stackoverflow.com/a/15456633/1180029 and the problem is fixed.

I've done this for both the intent used for dismissing a notification and for the intent used for clicking a notification and it all works fine. See gifs below (the `Before` one shows how 2 notifications for previously cancelled downloads reappear after I dismiss them and start another download):

Before | After
--- | ---
before | after


xrigau:
MRW notifications don't reappear

blundell:
:+1: this is a fix on to of removing the old uris?

xrigau:
https://github.com/novoda/download-manager/pull/170/files#r42522893 ??

yeah so that one shouldn've been removed I think and there was no URI for the other intent (the one used for when the notification is swiped away).

I think having the method with `unique` in the name makes it more explicit now

PR #174, opened 2015-11-13
Add "download-manager" subdirectory for the used cache dir

zegnus:
### Bug description ###

We have noticed that client's content inside the `getCacheDir()` is removed when a download starts

### Solution implemented ###

When the download manager instance is instantiated, goes through a process of cleaning up invalid files, calling the method `removeSpuriousFiles` https://github.com/novoda/download-manager/blob/master/library/src/main/java/com/novoda/downloadmanager/lib/StorageManager.java#L350, this method gets **all** the files inside the `getCacheDir()` and deletes them all (line https://github.com/novoda/download-manager/blob/master/library/src/main/java/com/novoda/downloadmanager/lib/StorageManager.java#L391)

There is a commented code that tries to discover the process owner of the file in order to prevent this bug but:
- It is commented (the library used is not available and there is an alternative library `Os.lstat().st_uid` but it is only available above Lollipop)
- It won't work if you include this library inside your project, the library runs in the same process as the client (or could potentially do)

The solution is to ensure that the files are stored inside a subdirectory `download-manager` instead of using the root one

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/StorageManager.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/StorageManager.java
@@ -294,7 +294,7 @@ class StorageManager {
     }
 
     public static File getDownloadDataDirectory(Context context) {
-        return context.getCacheDir();
+        return new File(context.getCacheDir().toString() + "/download-manager/");
     }
 
     /**
--- a/library/src/main/java/com/novoda/downloadmanager/lib/StorageManager.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/StorageManager.java
@@ -83,6 +83,8 @@ class StorageManager {
     private final ContentResolver contentResolver;
     private final DownloadsUriProvider downloadsUriProvider;
 
+    private final static String FILE_SEPARATOR = File.separator;
+
     StorageManager(
             ContentResolver contentResolver, 
             File externalStorageDir, 
@@ -294,7 +296,7 @@ class StorageManager {
     }
 
     public static File getDownloadDataDirectory(Context context) {
-        return new File(context.getCacheDir().toString() + "/download-manager/");
+        return new File(context.getCacheDir().getPath() + FILE_SEPARATOR + "download-manager" + FILE_SEPARATOR);
     }
 
     /**


ronocod:
Makes sense, LGTM :+1:

ouchadam:
:+1: also LGTM buttttttttttttttttttttttttt what happens to existing downloads from previous versions?

StylingAndroid:
:+1: LGTM

library/src/main/java/com/novoda/downloadmanager/lib/StorageManager.java
```diff
@@ -294,7 +294,7 @@ File getDownloadDataDirectory() {
     }
 
     public static File getDownloadDataDirectory(Context context) {
-        return context.getCacheDir();
+        return new File(context.getCacheDir().toString() + "/download-manager/");
```
juankysoriano: `context.getCacheDir().getPath()`.

blundell:
do we have hanging files now that the clean up doesn't happen?

library/src/main/java/com/novoda/downloadmanager/lib/StorageManager.java
ouchadam: could also use `File.separator` but I suppose we're pretty safe with the environment :wink:

juankysoriano:
I don't understand something, sorry guys. 

```
    /**
     * Removes files in the systemcache and downloads data dir without corresponding entries in
     * the downloads database.
     * This can occur if a delete is done on the database but the file is not removed from the
     * filesystem (due to sudden death of the process, for example).
     * This is not a very common occurrence. So, do this only once in a while.
     */
```

So it will remove the files only if there is a mismatch between the database and the `getCacheDir()` folder, isn't that ok? Or is the comment on that method not accurate and therefore we need to rewrite it?

zegnus:
@blundell Could potentially yes. 

The method does two things:
- Remove files that are in the database, this method hasn't changed
- Remove files that are in the getCacheDir but are not in the Database. This is the method that will delete files in a subdirectory. The javadoc says that this should not happen unless there is a sudden process kill or some other weird circumstance.

Still, this is a cache directory and not used for a long term persistance.

zegnus:
@juankysoriano That is AOSP code maybe we can ask a Google why of this :) I've elaborated a bit on my reply to @blundell

library/src/main/java/com/novoda/downloadmanager/lib/StorageManager.java
zegnus: https://github.com/novoda/download-manager/commit/506a32d0a7aebde6fd556612cc026d06703d81a6

ouchadam:
let's do it! :+1:

PR #176, opened 2015-11-19
Sets the current size as the total when adding completed downloads

ouchadam:
Previously the current size was defaulting to 0 when adding completed downloads, now it uses the total size since the file is already fully downloaded


xrigau:
:+1:

zegnus:
LGTM :+1:

PR #178, opened 2015-11-24
Notify pause fast

zegnus:
### Feature requested ###

When pausing, the client is notified after the `InputStream` is fully closed. The closing time depends on the device and can take up to 40s. In this scenario the client would see a `downloading` status even after pressing the `pause` button.

### Solution implemented ###

- We have implemented a new status `pausing` that will be triggered immediately to the client
- This Pull Request also includes the status `submitted` as a `pending` status, as it was missing in the master code and it was causing that the query for queued items was returning all items but the submitted ones. While in submitted status, the list would not show that item and then the item would show again when the item was properly processed (the list had items appearing and disappearing)

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.4-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.9-all.zip
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
@@ -231,8 +231,10 @@ public class DownloadService extends Service {
      * Enqueue an {#updateLocked()} pass to occur in future.
      */
     private void enqueueUpdate() {
-        updateHandler.removeMessages(MSG_UPDATE);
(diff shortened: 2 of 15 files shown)


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.4-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.9-all.zip
```
blundell: this guy, waiting for some PR's from you!

!][(https://media.giphy.com/media/BK8hW6YBpnvSU/giphy.gif)

zegnus: lol !! xDD

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
```diff
@@ -405,6 +407,7 @@ private void downloadOrContinueBatch(List<FileDownloadInfo> downloads) {
         for (FileDownloadInfo info : downloads) {
             if (!DownloadStatus.isCompleted(info.getStatus()) && !info.isSubmittedOrRunning()) {
                 download(info);
+                return;
```
juankysoriano: Why this return? shouldn't we continue processing the remaining elements?

zegnus: No, this was a bug where a tester found out a way to put two batches in the downloading status. Only one file at a time can be submitted as the file could belong to different batches (hence the bug).

Parallel download (if active) are dealt in a different way, not here.

ouchadam:
looks legit to me :+1:

PR #180, opened 2016-01-12
Refactor tar truncator to support record based tar files.

Dorvaryn:
Tar files that are large are usually written as records, that means that we might get padded zeros in writing those chuncks.

That was breaking the truncation logic that was in place.

The new approach is more resilient as it will read by records of 20 blocking factor (default blocking factor of tar) and wait for the end of the stream to try and backtrack within the last read record to find the last block entry before the EOF marker. It will then copy the record up to that point.

This will work for all record size smaller or equal to the default record size of 20.
We could parametrize the blocking factor to make this working with an arbitrary blocking factor.

What do you think ?

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/Constants.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/Constants.java
@@ -108,9 +108,14 @@ class Constants {
     public static final int TAR_BLOCK_SIZE = 512;
 
     /**
+     * The default blocking factor used by tar
+     */
+    private static final int DEFAULT_TAR_BLOCKING_FACTOR = 20;
+
+    /**
      * The size of a tar record
      */
-    public static final int TAR_RECORD_SIZE = TAR_BLOCK_SIZE * 20;
+    public static final int TAR_RECORD_SIZE = TAR_BLOCK_SIZE * DEFAULT_TAR_BLOCKING_FACTOR;
 
     /**
      * The buffer size used to stream the data
--- a/library/src/main/java/com/novoda/downloadmanager/lib/Constants.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/Constants.java
@@ -108,6 +108,11 @@ class Constants {
     public static final int TAR_BLOCK_SIZE = 512;
 
     /**
+     * The size of a tar record
+     */
+    public static final int TAR_RECORD_SIZE = TAR_BLOCK_SIZE * 20;
+
+    /**
      * The buffer size used to stream the data
      */
     public static final int BUFFER_SIZE = 8 * TAR_BLOCK_SIZE;
--- a/library/src/main/java/com/novoda/downloadmanager/lib/TarTruncator.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/TarTruncator.java
@@ -42,17 +42,17 @@ class TarTruncator implements DataTransferer {
         }
     }
 
-    private int truncateEOFMarker(byte[] buffer, int length) {
+    private static int truncateEOFMarker(byte[] buffer, int length) {
         int position = length;
 
-        while (position >= Constants.TAR_BLOCK_SIZE && isFullOfZeroes(buffer, position - Constants.TAR_BLOCK_SIZE, position)) {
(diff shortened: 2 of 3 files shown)


library/src/main/java/com/novoda/downloadmanager/lib/TarTruncator.java
```diff
@@ -17,11 +17,22 @@ public TarTruncator(DataWriter dataWriter) {
...
+            byte[] swappingRef;
+            int read;
+            int readLast = 0;
+
+            while ((read = readRecord(in, buffer)) > 0) {
+                newState = dataWriter.write(newState, buffer2, readLast);
+                swappingRef = buffer2;
+                buffer2 = buffer;
+                buffer = swappingRef;
+                readLast = read;
```
jonreeve: This is a bit hard to parse atm. Get it eventually, but some renaming would help a bit

library/src/main/java/com/novoda/downloadmanager/lib/TarTruncator.java
```diff
@@ -31,31 +42,30 @@ public TarTruncator(DataWriter dataWriter) {
...
 
-    private boolean isNotFullOfZeroes(byte[] buffer) {
-        for (byte b : buffer) {
-            if (b != BYTE_ZERO) {
-                return true;
-            }
+    private int truncateEOFMarker(byte[] buffer, int length) {
+        int position = length;
+
+        while (position >= Constants.TAR_BLOCK_SIZE && isFullOfZeroes(buffer, position - Constants.TAR_BLOCK_SIZE, position)) {
```
jonreeve: isFullOfZeroes - nice!

library/src/main/java/com/novoda/downloadmanager/lib/Constants.java
```diff
@@ -108,6 +108,11 @@
     public static final int TAR_BLOCK_SIZE = 512;
 
     /**
+     * The size of a tar record
+     */
+    public static final int TAR_RECORD_SIZE = TAR_BLOCK_SIZE * 20;
```
ataulm: >The new approach is more resilient as it will read by records of 20 blocking factor (default blocking factor of tar)

You mentioned parameterising the 20 - yday I recommended doing it as part of this PR and leave 20 as the default. but now thinking on it, yagni; let's do it when someone raises the issue. The 20 can be extracted as a private constant though as `DEFAULT_TAR_BLOCKING_FACTOR` as it's not an arbitrary number.

library/src/main/java/com/novoda/downloadmanager/lib/TarTruncator.java
```diff
@@ -31,31 +42,30 @@ public TarTruncator(DataWriter dataWriter) {
...
+            if (buffer[i] != BYTE_ZERO) {
+                return false;
+            }
         }
-        return read;
+        return true;
     }
 
-    private int readBlock(InputStream fileInputStream, byte[] buffer) throws IOException {
+    private int readRecord(InputStream fileInputStream, byte[] buffer) throws IOException {
```
ataulm: could be static

library/src/main/java/com/novoda/downloadmanager/lib/TarTruncator.java
```diff
@@ -31,31 +42,30 @@ public TarTruncator(DataWriter dataWriter) {
         }
     }
 
-    private boolean isNotFullOfZeroes(byte[] buffer) {
-        for (byte b : buffer) {
-            if (b != BYTE_ZERO) {
-                return true;
-            }
+    private int truncateEOFMarker(byte[] buffer, int length) {
```
ataulm: `truncateEofMarker` - acronyms are treated as words

ataulm: is this method more `findMarkerToTruncateEof()`. atm (truncate being a verb) it sounds like it will truncate the EOF marker and I don't think it is.

ataulm:
the explanation you gave yesterday made sense, and this seems to line up with your explanation but I'm not confident enough reading it to assert that it lines up -> :+1:

jonreeve:
LGTM :+1:

PR #181, opened 2016-02-12
There's no need to enforce WRITE_EXTERNAL_STORAGE permission for internal downloads.

mcroghan:
The problem: Currently, DownloadProvider.insertDownload() enforces WRITE_EXTERNAL_STORAGE permission for all downloads using a DESTINATION_FILE_URI, and throws a SecurityException if that permission is missing. However, that permission isn't actually required if the download is destined for internal storage.

The solution: Check to see if the file path contains the app package name. If it does, it's using internal storage and there's no need for WRITE_EXTERNAL_STORAGE permission; therefore, waive enforcement of that permission.

image

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadProvider.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadProvider.java
@@ -408,10 +408,12 @@ public final class DownloadProvider extends ContentProvider {
                 dest = DownloadsDestination.DESTINATION_CACHE_PARTITION;
             }
             if (dest == DownloadsDestination.DESTINATION_FILE_URI) {
-                getContext().enforcePermission(
-                        android.Manifest.permission.WRITE_EXTERNAL_STORAGE, Binder.getCallingPid(), Binder.getCallingUid(),
-                        "need WRITE_EXTERNAL_STORAGE permission to use DESTINATION_FILE_URI");
-                checkFileUriDestination(values);
+                if (!getFileUriDestinationPath(values).contains("/" + getContext().getPackageName() + "/")) {
+                    // external, not internal storage
+                    getContext().enforcePermission(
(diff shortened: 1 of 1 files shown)


hal9002:

blundell:
seems legit! :+1:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadProvider.java
```diff
@@ -408,10 +408,12 @@ private Uri insertDownload(Uri uri, ContentValues values, SQLiteDatabase db, int
                 dest = DownloadsDestination.DESTINATION_CACHE_PARTITION;
             }
             if (dest == DownloadsDestination.DESTINATION_FILE_URI) {
-                getContext().enforcePermission(
-                        android.Manifest.permission.WRITE_EXTERNAL_STORAGE, Binder.getCallingPid(), Binder.getCallingUid(),
-                        "need WRITE_EXTERNAL_STORAGE permission to use DESTINATION_FILE_URI");
-                checkFileUriDestination(values);
+                if (!getFileUriDestinationPath(values).contains("/" + getContext().getPackageName() + "/")) {
```
amlcurran: doesn't the internal storage directory have the package name too? I thought it was `/data/data/{package-name}/files`?

mcroghan: Agh, you're right (Assume you mean external). It's something like that and will contain the package.

Our use case only wants to use internal storage and we don't want to request the external permission. I'll try to find a reliable and correct (!) way of determining which kind of storage is being used within this method. I'll close this PR - thanks (and sorry!) for you time.

amlcurran: No problem! Perhaps you do some file path checking against [`Environment#getDataDirectory()`](http://developer.android.com/reference/android/os/Environment.html#getDataDirectory()) instead?

PR #182, opened 2016-02-15
Add new listener for DownloadManagerModules being released

blundell:
- This means if anyone makes a module that needs some cleanup before GC, they can do that.

Paired with @hhaouat


xrigau:
:100:

PR #183, opened 2016-02-18
There's no need to enforce WRITE_EXTERNAL_STORAGE permission for internal downloads. (try #2)

mcroghan:
This is a second try for closed PR #181.
 
The problem: Currently, DownloadProvider.insertDownload() enforces WRITE_EXTERNAL_STORAGE permission for all downloads using a DESTINATION_FILE_URI, and throws a SecurityException if that permission is missing. However, that permission isn't actually required if the download is destined for internal storage.

The solution: Check to see if the file path begins with Environment.getDataDirectory(). If it does, it's using internal storage and there's no need for WRITE_EXTERNAL_STORAGE permission; therefore, waive enforcement of that permission.

@amlcurran, thanks for the suggestion - I think this will do it.

image

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadProvider.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadProvider.java
@@ -408,10 +408,12 @@ public final class DownloadProvider extends ContentProvider {
                 dest = DownloadsDestination.DESTINATION_CACHE_PARTITION;
             }
             if (dest == DownloadsDestination.DESTINATION_FILE_URI) {
-                getContext().enforcePermission(
-                        android.Manifest.permission.WRITE_EXTERNAL_STORAGE, Binder.getCallingPid(), Binder.getCallingUid(),
-                        "need WRITE_EXTERNAL_STORAGE permission to use DESTINATION_FILE_URI");
-                checkFileUriDestination(values);
+                if (!getFileUriDestinationPath(values).contains("/" + getContext().getPackageName() + "/")) {
+                    // external, not internal storage
+                    getContext().enforcePermission(
+                            android.Manifest.permission.WRITE_EXTERNAL_STORAGE, Binder.getCallingPid(), Binder.getCallingUid(),
+                            "need WRITE_EXTERNAL_STORAGE permission to use DESTINATION_FILE_URI");
+                }
             } else if (dest == DownloadsDestination.DESTINATION_SYSTEMCACHE_PARTITION) {
                 getContext().enforcePermission(
                         "android.permission.ACCESS_CACHE_FILESYSTEM", Binder.getCallingPid(), Binder.getCallingUid(),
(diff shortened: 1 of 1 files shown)


hal9002:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadProvider.java
```diff
@@ -408,10 +408,12 @@ private Uri insertDownload(Uri uri, ContentValues values, SQLiteDatabase db, int
                 dest = DownloadsDestination.DESTINATION_CACHE_PARTITION;
             }
             if (dest == DownloadsDestination.DESTINATION_FILE_URI) {
-                getContext().enforcePermission(
-                        android.Manifest.permission.WRITE_EXTERNAL_STORAGE, Binder.getCallingPid(), Binder.getCallingUid(),
-                        "need WRITE_EXTERNAL_STORAGE permission to use DESTINATION_FILE_URI");
-                checkFileUriDestination(values);
+                if (!getFileUriDestinationPath(values).startsWith(Environment.getDataDirectory() + "/")) {
```
rock3r: Just a note, it's probably wise to use `File#getAbsolutePath()` for this comparison

mcroghan: Hmm. Thanks, @rock3r  - I'm struggling with this, though, because at this point in the code this destination file path is nowhere near being turned into an actual File that you could call getAbsolutePath() on, and there's a whole lot of logic elsewhere, largely in the Helpers class, including the generateSaveFile(), getFullPath(), and chooseFileName() methods, which come into play before actually turning this into a File. I'd be reluctant to bring all that code in to play at this point. Also, as far as I can tell, the only way the COLUMN_FILE_NAME_HINT in the DB could be a relative URI is if you passed one in via Request.setDestinationUri() (rather than using one of the other setDestination... methods). If you did that, and you used a path that begins with the result of Environment.getDataDirectory()...well, then I think it would actually just work as an internal-storage download, and I'm not sure that's problematic. It's quite possible I'm all wet, though. This is from an hour or two of poking through the code.

rock3r: I think that's more insight than I have then :) Mine was just a generic recommendation, I'm not very familiar with the download manager code to be honest.

ouchadam:
:+1: looks legit to me! will leave to @amlcurran to hit the merge button

amlcurran:
Looks good to me too! :shipit: 

Thanks for the feature @mcroghan !

alveru:
Hello !

I am getting the following error in my application when setting to use the internal app directory:

Exception: need WRITE_EXTERNAL_STORAGE permission to use DESTINATION_FILE_URI: uid 10134 does not have android.permission.WRITE_EXTERNAL_STORAGE

Why do I need this permission on Android Marshmallow if I am using the internal app data folder?

Am I doing something wrong?  I am using this version of library:

compile 'com.novoda:download-manager:0.2.38'

Thanks !!

PR #184, opened 2016-03-02
Refactor BatchRepository

fourlastor:
This PR splits `BatchRepository` in 4 collaborators (each of which is covering one area that `BatchRepository` was covering) and renames it to `BatchFacade`:
- `BatchStatusService`
- `BatchStartingService`
- `BatchDeletionService`
- `BatchRetrievalService`

Tests have been added around the various methods of `BatchFacade` before refactoring, after the refactoring was successful, I moved the tests down one level to the collaborator and added tests for `BatchFacade` to be sure it uses its collaborators.

The only exception to this is `BatchRepository.retrieveBatchesFor` which I found **extremely** hard to test due to the intensive use of `Cursor`


library/build.gradle
```diff
@@ -50,6 +50,8 @@ dependencies {
 
     testCompile 'junit:junit:4.12'
     testCompile 'org.mockito:mockito-core:1.10.19'
+    testCompile 'org.powermock:powermock-module-junit4:1.6.4'
```
blundell: :-1: powermock is not cool, I assume you've added this to either mock static methods or override final classes? A better way is to highlight where these design issues are in the code so that we can make the code better rather than using a brute force tool to get around the issue

library/src/main/java/com/novoda/downloadmanager/lib/BatchDeletionService.java
```diff
@@ -0,0 +1,91 @@
...
+import com.novoda.notils.string.StringUtils;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+class BatchDeletionService {
+
+    private final DownloadDeleter downloadDeleter;
+    private ContentResolver resolver;
```
blundell: should this be final?

otherwise split it from the others ie. whiteline space

blundell:
can you explain why the name change from `BatchRepository` to `BatchFacade`?  A facade is an implementation detail that you don't usually expose. The old name explained that query these methods was to do with query from a datastore.

library/build.gradle
fourlastor: Discussed together and agreed that it's the least evil to be able to test it ;)

fourlastor:
@blundell restored `BatchRepository` as the class name, also renamed the collaborators to be `*Repository` as we talked about yesterday :dancer: 

Also fixed `BatchDownloadsRepository` since it was breaking the tests (I introduced a regression :dancer: )

library/build.gradle
blundell: :+1:

library/src/main/java/com/novoda/downloadmanager/lib/BatchRetrievalException.java
```diff
@@ -0,0 +1,15 @@
+package com.novoda.downloadmanager.lib;
+
+final class BatchRetrievalException extends RuntimeException {
+    static BatchRetrievalException failedQueryForBatches() {
+        return new BatchRetrievalException("Failed to query for batches");
+    }
+
+    static BatchRetrievalException failedQueryForBatch(long batchId) {
```
blundell: I think its a bit strange to have static factory methods on exceptions? Most people don't expect this with exceptions what about just using the constructor?

```java
public BatchRetrievalException() {
        super("Failed to query for batches");
    }

public BatchRetrievalException(long batchId) {
        super("Failed to query for batch with batchId = " + batchId);
    }
```

library/src/main/java/com/novoda/downloadmanager/lib/BatchStatusRepository.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    int calculateBatchStatusFromDownloads(long batchId) {
+        Cursor cursor = queryForDownloadStatusesByBatch(batchId);
+
+        statuses.clear();
+
+        marshallInStatuses(cursor);
+
+        cursor.close();
```
blundell: can you put this in a try finally?

library/src/main/java/com/novoda/downloadmanager/lib/Cursors.java
```diff
@@ -0,0 +1,27 @@
...
+    public static long getLong(Cursor c, String column) {
+        return c.getLong(columnIndexFor(c, column));
+    }
+
+    public static String getString(Cursor c, String column) {
+        return c.getString(columnIndexFor(c, column));
+    }
+
+    private static int columnIndexFor(Cursor c, String column) {
+        return c.getColumnIndexOrThrow(column);
```
blundell: not sure it's wise to have this `OrThrow` hidden inside a method that does not explicitly say that. When I first read line 12 I was wondering why you didn't use `OrThrow` but you do just that the method is named wrong. I would inline `columnIndexFor` or rename it

library/src/main/java/com/novoda/downloadmanager/lib/Cursors.java
```diff
@@ -0,0 +1,27 @@
...
+
+    public static int getInt(Cursor c, String column) {
+        return c.getInt(columnIndexFor(c, column));
+    }
+
+    public static long getLong(Cursor c, String column) {
+        return c.getLong(columnIndexFor(c, column));
+    }
+
+    public static String getString(Cursor c, String column) {
```
blundell: not sure why we have these static helper methods? I would get rid of this class, next thing someone will be mocking these methods with powermock! ew

library/src/main/java/com/novoda/downloadmanager/lib/DownloadsUriProvider.java
```diff
@@ -59,14 +55,6 @@ private static DownloadsUriProvider newInstance() {
     }
 
     /**
-     * The content URI for accessing publicly accessible downloads (i.e., it requires no
-     * permissions to access this downloaded file)
-     */
-    public Uri getPubliclyAccessibleDownloadsUri() {
```
blundell: this is removed because there is no public access now?

library/src/main/java/com/novoda/downloadmanager/lib/Statuses.java
```diff
@@ -0,0 +1,113 @@
...
+        return DownloadStatus.UNKNOWN_ERROR;
+    }
+
+    @Override
+    public String toString() {
+        StringBuilder stringBuilder = new StringBuilder("{");
+
+        int size = statusCounts.size();
+        for (int i = 0; i < size; i++) {
+            stringBuilder.append("[status: ").append(statusCounts.keyAt(i)).append(", count: ").append(statusCounts.valueAt(i)).append("]");
```
blundell: would be nicer with each append on a new line

library/src/main/java/com/novoda/downloadmanager/lib/BatchRetrievalException.java
fourlastor: Because it doesn't give it a name mainly. Also, with constructors, usually people tend to end up with the message as a parameter, leaking that information outside the class and hiding when the exception has come to a point where it makes sense to split it.

People are used to it because it's easier, imho not a good reason to keep the habit :)

library/src/main/java/com/novoda/downloadmanager/lib/BatchStatusRepository.java
fourlastor: :+1: will do

library/src/main/java/com/novoda/downloadmanager/lib/Cursors.java
fourlastor: Makes sense, I'll append `orThrow` to the name.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadsUriProvider.java
fourlastor: It wasn't used, I suppose simply no one noticed it before.

library/src/main/java/com/novoda/downloadmanager/lib/Cursors.java
fourlastor: The main idea is to simplify the use of `Cursor` which is awkward at least.

You don't need to mock this method, test wise you've the same problem (using cursor will make that part of the code hard to test), it just removes noise from the client.

library/src/main/java/com/novoda/downloadmanager/lib/Cursors.java
fourlastor: Actually I'll inline :dancer:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadsUriProvider.java
fourlastor: Oh wait, I might misunderstood the scope of this class: is this to be used by the client?

library/src/main/java/com/novoda/downloadmanager/lib/BatchStatusRepository.java
fourlastor: 4f941f3

library/src/main/java/com/novoda/downloadmanager/lib/Cursors.java
fourlastor: 24a972c

library/src/main/java/com/novoda/downloadmanager/lib/Cursors.java
fourlastor: 0b3e92a

library/src/main/java/com/novoda/downloadmanager/lib/BatchRetrievalException.java
blundell: like I showed the message should be inside the class and so people putting it outside is not a valid reason. With the above it enforces that you cannot have the full message as a param.

Creating an exception through a factory method is very strange, when something is exceptional you should know exactly why & what exception you are using.
If you want to separate these two exceptions by name i.e. they are that different - then they should be two different classes.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadsUriProvider.java
blundell: yes be careful if something is public in a lin you have to really think about it. I think it was client facing :x:

library/src/main/java/com/novoda/downloadmanager/lib/Cursors.java
```diff
@@ -0,0 +1,23 @@
+package com.novoda.downloadmanager.lib;
+
+import android.database.Cursor;
+
+final class Cursors {
```
blundell: I don't think this class is a wise move :x: its secretly a `CursorUtil` and we don't want that. Decorate the cursor if you don't think it's methods are good enough

library/src/main/java/com/novoda/downloadmanager/lib/DownloadsUriProvider.java
fourlastor: going to restore it

fourlastor: a2267e3

fourlastor:
Please test this again.

fourlastor:
@blundell what do you think of `BetterCursor` instead of `Cursors`? It's basically a wrapper with added methods. I didn't use the Android fw wrapper because it would've made the tests more complicated (not mocked exception, as usual : / )

Could be part of notils if you like the idea :tongue:

library/src/main/java/com/novoda/downloadmanager/lib/BatchRetrievalException.java
fourlastor: 6a1dada

library/src/main/java/com/novoda/downloadmanager/lib/db/BetterCursor.java
```diff
@@ -0,0 +1,27 @@
...
+    public static BetterCursor wrap(Cursor cursor) {
+        return new BetterCursor(cursor);
+    }
+
+    BetterCursor(Cursor cursor) {
+        super(cursor);
+    }
+
+    public int getInt(String columnName) {
+        return getInt(getColumnIndexOrThrow(columnName));
```
blundell: I understand your sentiment here, trying to enforce the idea of throwing all the time, which is good! However anyone who normally uses cursor will be really confused by using this method ie.. `getInt` that does `getIntOrThrow`. It seems the throw is now implicit and you should always favor explicit code. This might be a bit of over engineering, lets keep it simple? The rest of the PR is ready to ship! :)

fourlastor: as in? Renaming it to `getXOrThrow`? Or just removing this class? I don't know if removing this class would really make the rest of the code cleaner :P

blundell: just remove the class tbh

fourlastor:
@blundell  7195788

blundell:
LGETM :+1: see if you can find a second thumbs up now :wink:

library/src/test/java/com/novoda/downloadmanager/lib/BatchStatusRepositoryTests.java
```diff
@@ -0,0 +1,322 @@
...
+import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.anyCollection;
+import static org.mockito.Matchers.anyString;
+import static org.mockito.Matchers.same;
+import static org.mockito.Mockito.*;
+import static org.powermock.api.mockito.PowerMockito.mockStatic;
+import static org.powermock.api.mockito.PowerMockito.whenNew;
+
+@RunWith(Enclosed.class)
+public class BatchStatusRepositoryTests {
```
xrigau: syk tests :+1:

library/src/main/java/com/novoda/downloadmanager/lib/Statuses.java
```diff
@@ -0,0 +1,118 @@
...
+                return status;
+            }
+        }
+
+        return DownloadStatus.UNKNOWN_ERROR;
+    }
+
+    @Override
+    public String toString() {
+        StringBuilder stringBuilder = new StringBuilder("{");
```
xrigau: :+1:

xrigau:
:+1: nice action, most changes are tests and that's goood! you shall get a :shipit:

PR #186, opened 2016-04-19
Fix status mapping

PaNaVTEC:
### Some context ###
In the a previous commit: https://github.com/novoda/download-manager/commit/b6d286a2d052971e0e71935b3ea86839cdec0690 we are now throwing an exception every time an error occurs when transferring data independently if the download is paused, queued, etc, which is right. But know we get "paused" status when is internally "waiting for network".

### Problem description ###
The download mapper,  was wrong since we were mapping "queued" to "paused" where it needs to be: "pending".
The other problems occurs when the `DownloadService` restarts the downloads when the network resumes after a disconnection. In this case, we are not calling `downloadReadyChecker.canDownload()` method and therefore we are not propagating that "event" to the client.

### Solution implemented ###
For the first sub-problem: Fix the mapper.
For the second: Force the downloadTask to trigger the check rules on the clients to send them that "event"

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
@@ -345,7 +345,7 @@ public class DownloadService extends Service {
                 continue;
             }
 
-            if (!isActive && downloadReadyChecker.canDownload(downloadBatch)) {
+            if (downloadReadyChecker.canDownload(downloadBatch) && !isActive) {
                 boolean isBatchStartingForTheFirstTime = batchRepository.isBatchStartingForTheFirstTime(downloadBatch.getBatchId());
                 if (isBatchStartingForTheFirstTime) {
                     handleBatchStartingForTheFirstTime(downloadBatch);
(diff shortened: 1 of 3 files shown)


library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
```diff
@@ -345,7 +345,7 @@ private boolean updateLocked() {
                 continue;
             }
 
-            if (!isActive && downloadReadyChecker.canDownload(downloadBatch)) {
+            if (downloadReadyChecker.canDownload(downloadBatch) && !isActive) {
```
ouchadam: :x: I'm still not sold on this, it doesn't make sense to check if inactive downloads `canDownload`, we'll end up spamming `canDownload` for no reason which will lead to other problems for example:

if a client has a space check in `canDownload` and has an active download which the device has enough space for with other paused items which the device does not have space for, the library will be incorrectly checking against the paused items and if the logic to handle out of space errors is inside the client side `canDownload` it'll be triggered even though there is enough space for the active download

joetimmins:
There's no unit tests around either of these classes - do we need some?

library/src/main/java/com/novoda/downloadmanager/lib/DownloadTask.java
```diff
@@ -335,6 +336,16 @@ private void runInternal() {
         storageManager.incrementNumDownloadsSoFar();
     }
 
+    private void hackToForceClientsRefreshRulesIfConnectionDropped(int finalStatus) {
```
ouchadam: :100:

PR #193, opened 2016-08-05
Migrate fields in Request to java.net.URI

eduardb:
#### Problem

`Request`s cannot be used in tests because of `android.net.Uri` dependency

#### Solution
I decided to migrate to `java.net.URI` instead, and deprecate the previous constructor and setter.

Thanks @ouchadam for testing again the change on all4 😂

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/Request.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/Request.java
@@ -92,7 +92,6 @@ public class Request {
     /**
      * @param uri the HTTP URI to download.
      */
-    @Deprecated
     public Request(URI uri) {
         validateUriScheme(uri.toString(), uri.getScheme());
         this.uri = uri;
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/novoda/downloadmanager/lib/Request.java
```diff
@@ -81,21 +82,51 @@
     private int notificationVisibility = NotificationVisibility.ONLY_WHEN_ACTIVE;
 
     /**
-     * @param uri the HTTP URI to download.
+     * @param uri the HTTP Uri to download.
      */
+    @Deprecated
```
blundell: write in the javadoc what is the alternative to use

eduardb: Good point, done in 0c04ae8

jonreeve:
LGTM 👍

PR #194, opened 2016-08-08
Update bintray release plugin

eduardb:
To fix the following build failure:
```
Execution failed for task ':library:mavenAndroidJavadocs'.
> Javadoc generation failed.
```

Although 0.3.5 is the latest version of the plugin, it is known to have the same issue. So we should use 0.3.4


rock3r:
LGTM 👍

Issue #201, opened 2016-11-23
Can't download to SD Card


Hackmodford:
My app downloads audiobooks from a server. I have given the user the option to choose a preferred storage method. If the user chooses their SD card I would like to download files to their SD card instead of internal memory.

When I set the download path to the SD card using request.setDestinationUri. The download fails with an unknown error.

This is the path I am trying to use.
/storage/external_SD/Android/data/com.hughesmedia.big_finish/files/Download

Is there some limitation that I'm not aware of?

Thanks

Hackmodford:
I turned on verbos logging and got this result.

`11-25 10:11:28.647 W/EasyPermissions: hasPermissions: API version < M, returning true by default
11-25 10:11:30.544 V/DownloadManager: initiating download with UID 10014 
11-25 10:11:30.584 V/DownloadManager: Service onCreate 
11-25 10:11:30.585 I/DownloadManager: in removeSpuriousFiles 
11-25 10:11:30.586 I/DownloadManager: in trimDatabase 
11-25 10:11:30.586 V/DownloadManager: starting query, database is not null; projection[0] is _id; selection is status >= '200'; selectionArgs is null; sort is last_modified_timestamp. 
11-25 10:11:30.590 V/DownloadManager: created cursor android.database.sqlite.SQLiteCursor@16c4cb2 on behalf of 29450 
                                      
                                      --------- beginning of amazon_main
11-25 10:11:30.595 D/DownloadManager: invalid selection [(control is null or control = ? ) AND (status = ? or status = ?)) GROUP BY (batch_id] triggered java.lang.IllegalArgumentException: unrecognized column or keyword 
11-25 10:11:30.596 V/DownloadManager: starting query, database is not null; projection[0] is DISTINCT batch_id; selection is (control is null or control = ? ) AND (status = ? or status = ?)) GROUP BY (batch_id; selectionArgs[0] is 0; selectionArgs[1] is 192; selectionArgs[2] is 189; sort is null. 
11-25 10:11:30.599 V/DownloadManager: created cursor android.database.sqlite.SQLiteCursor@1e271403 on behalf of 29450 
11-25 10:11:30.605 V/DownloadManager: Service onStart 
11-25 10:11:30.607 V/DownloadManager: starting query, database is not null; projection is null; selection is null; selectionArgs is null; sort is _id ASC. 
[...]

PR #203, opened 2016-11-25
Allow downloads to SD card

Hackmodford:
These changes allow the download manager to download to a directory retrieved from context.getExternalFilesDirs(null);

This fixes the problem I had in issue #201

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
@@ -139,9 +139,14 @@ public class DownloadService extends Service {
         ContentResolver contentResolver = getContentResolver();
         File downloadDataDir = StorageManager.getDownloadDataDirectory(this);
         File externalStorageDir = Environment.getExternalStorageDirectory();
+        File[] externalStorageDirs = new File[0];
+        if (android.os.Build.VERSION.SDK_INT >= 19 )
+        {
+            externalStorageDirs = this.getExternalFilesDirs(null);
+        }
         File internalStorageDir = Environment.getDataDirectory();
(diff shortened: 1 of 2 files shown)


stefanhoth:

xrigau:
ok to test

library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
```diff
@@ -139,9 +139,14 @@ public void onCreate() {
         ContentResolver contentResolver = getContentResolver();
         File downloadDataDir = StorageManager.getDownloadDataDirectory(this);
         File externalStorageDir = Environment.getExternalStorageDirectory();
+        File[] externalStorageDirs = new File[0];
+        if (android.os.Build.VERSION.SDK_INT >= 19 )
```
xrigau: @Hackmodford what happens if this is running on a pre-19 device? is it not supported?

xrigau: also @Hackmodford it'd be awesome if you can use the novoda IDE settings for the formatting when you submit a PR in order to keep some consistency in the codebase 😉  https://github.com/novoda/novoda/tree/master/ide-settings/intellij

xrigau: ah so the download failure was only in 19 and newer! thanks for clarifying 👍

Hackmodford: Yes, in 19+ you can get an array of storage directories that can be used. Some are emulated external dirs and some are the actual SD card.

PR #212, opened 2017-03-20
Replace deprecated intent by job scheduler

zegnus:
### Problem description

A client targeting android N will make the current DownloadManager to fail to reconnect when there is a network change.

This is due to that the permission `android.net.conn.CONNECTIVITY_CHANGE` is being deprecated in favour of using a `JobScheduler`. For reference [connectivityManager](https://developer.android.com/reference/android/net/ConnectivityManager.html#CONNECTIVITY_ACTION) and [android N background optimizations](https://developer.android.com/about/versions/nougat/android-7.0.html#background_optimizations)

We can still subscribe a broadcast event for `CONNECTIVITY_CHANGE` programatically but the application will only receive events when its in foreground.

### Solution

Connectivity reconnections are part of the download manager's responsibility and the solution has been designed with the objective of keeping this functionality in place for all android versions.

We have moved the scheduling of the downloads from a service, and triggered by a connectivity change event, to a scheduling a job. The library that we have chosen is the [`JobScheduler`](https://github.com/evernote/android-job) library from `Evernote` as provides backward compatibility in one library.

#### Architectural changes

The major change in architecture is the following:
- Previously the management of the download of all batches and files happened inside an infinite thread loop living in a service.
- This implementation replaces this notion by scheduling a job.

The code under the new class `DownloadServiceJob` is a modification of the original `DownloadService`. No refactoring of the class has been done that wasn't strictly necessary. The code is still Google's legacy.

**Jobs**

A Job is a background task responsible of processing batches until all batches are in a status that no more work can be done on them. The job will loop one file at a time with the aim of a full download. Then it will start again with the next file/batch. And it will finally end.

A job can end in three different status:
- `SUCCESS`: All files and batches and processed correctly and everything has been downloaded or one or more batches cannot proceed with the download because they are paused
- `RESCHEDULE`: A job will be automatically rescheduled if we are waiting for a valid network or for a client to free up restrictions
- `FAILURE`: If there is any other status will make the job to fail. In this case it will need a _manual_ scheduling of a new job

A job will be scheduled every time that there is an action in a download. As an example when a download is scheduled, deleted, network change, change in the database of any sort (but not deletion)

There is also several mechanism in place in order to avoid multiple jobs to proceed on an already running batch.

Notifications are also provided with a refresh update frequency of two seconds.

Deletion has been improved by aborting a download in the write to disc method instead of waiting for a new pass in the database. This makes the download to finish earlier and avoids unnecessary method calls.

The network change rescheduling is going to be guaranteed by the following mechanisms:
- For all devices < Android M the broadcast event will be used
- For all devices >= Android M a programmatically set up network change notification will be used instead
- A job will be rescheduled automatically when the connectivity fails. The job is set up to be executed between one and two seconds. If after two seconds the network is not available the job will run in a linear fashion back-off mode of 5 seconds.

Due to the nature of the job scheduler and the back-off mechanism, the previously two described events plus the usage of the job scheduler guarantees that the download will be resume as soon as possible. Either trusting the validity of the notification changes or the job scheduling.

As a general advice, it is recommended that the clients use a unique instance of the `DownloadManager` living in their `Application` class.


demo-extended/src/main/java/com/novoda/downloadmanager/demo/extended/batches/BatchDownloadsActivity.java
```diff
@@ -67,6 +73,18 @@ public void onClick(@NonNull View v) {
                 });
 
         setupQueryingExample();
+
+        ConnectivityManager connectivityManager = (ConnectivityManager) getSystemService(Context.CONNECTIVITY_SERVICE);
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
```
zegnus: Any other version will use the connectivity change broadcast that's still available

library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
```diff
@@ -31,7 +37,26 @@ public DownloadManagerBuilder withVerboseLogging() {
...
     public DownloadManager build() {
         ContentResolver contentResolver = context.getContentResolver();
-        return new DownloadManager(context, contentResolver, verboseLogging);
+        DownloadManager downloadManager = new DownloadManager(context, contentResolver, verboseLogging);
+        updateOnNetworkChanges(downloadManager);
+        return downloadManager;
     }
 
+    private void updateOnNetworkChanges(final DownloadManager downloadManager) {
+        if (deviceDoesNotSupportConnectivityChangesBroadcast()) {
```
zegnus: Lower versions will still receive the broadcast `CONNECTIVITY_CHANGE`

library/src/main/java/com/novoda/downloadmanager/lib/DownloadProvider.java
```diff
@@ -755,13 +752,6 @@ public int update(final Uri uri, final ContentValues values, final String where,
         int count;
         boolean startService = false;
 
-        if (values.containsKey(DownloadContract.Downloads.COLUMN_DELETED)) {
-            if (values.getAsInteger(DownloadContract.Downloads.COLUMN_DELETED) == 1) {
-                // some rows are to be 'deleted'. need to start DownloadService.
-                startService = true;
```
zegnus: No need to schedule a job anymore, the deletion happens immediately in the job itself

library/src/main/java/com/novoda/downloadmanager/lib/DownloadTask.java
```diff
@@ -587,11 +544,17 @@ private void transferData(State state, InputStream in, OutputStream out) throws
 
         State newState = dataTransferer.transferData(state, in);
         handleEndOfStream(newState);
+        LLog.v("end transfer data");
     }
 
     private final NotifierWriter.WriteChunkListener checkOnWrite = new NotifierWriter.WriteChunkListener() {
         @Override
         public void chunkWritten(FileDownloadInfo downloadInfo) throws StopRequestException {
+            if (clock.intervalLessThan(Clock.Interval.ONE_SECOND)) {
```
zegnus: Write to disc in chunks can happen really fast, in order to do not hammer the client we only check once every second.

library/src/main/java/com/novoda/downloadmanager/lib/jobscheduler/DownloadJob.java
```diff
@@ -0,0 +1,121 @@
...
+    private static final boolean ENFORCED_EXECUTION = true;
+    private static final long BACKOFF_MILLIS = TimeUnit.SECONDS.toMillis(5);
+    private static final long EXECUTION_START_MILLIS = TimeUnit.SECONDS.toMillis(1);
+    private static final long EXECUTION_END_MILLIS = TimeUnit.SECONDS.toMillis(2);
+    private static final Object LOCK = new Object();
+
+    @NonNull
+    @Override
+    protected Result onRunJob(Params params) {
+        LLog.v("job starts right now");
```
zegnus: This job runs in a background thread by default

library/src/main/java/com/novoda/downloadmanager/lib/jobscheduler/DownloadJob.java
```diff
@@ -0,0 +1,121 @@
...
+            LLog.v("job failure");
+            return Result.FAILURE;
+        }
+    }
+
+    private void ensureDownloadServiceJobInstanceExists(final DownloadServiceJob[] downloadServiceJob) {
+        new Handler(Looper.getMainLooper()).post(new Runnable() {
+            @Override
+            public void run() {
+                downloadServiceJob[0] = DownloadServiceJob.getInstance();
```
zegnus: `.getInstance()` needs to happen in the main thread due to content provider instantiations

library/src/main/java/com/novoda/downloadmanager/lib/jobscheduler/DownloadJob.java
```diff
@@ -0,0 +1,121 @@
...
+        scheduleJob(EXECUTION_START_MILLIS, STANDARD_EXECUTION);
+    }
+
+    public static void scheduleEnforcedJob() {
+        LLog.v("scheduling a job to start immediately in 1s");
+        scheduleJob(EXECUTION_START_MILLIS, ENFORCED_EXECUTION);
+    }
+
+    private static void scheduleJob(final long startMillis, final boolean enforceJob) {
+        new Handler(Looper.getMainLooper()).post(new Runnable() {
```
zegnus: We guarantee that we always schedule a job from the main thread

library/build.gradle
```diff
@@ -15,12 +15,12 @@ repositories {
...
 android {
-    compileSdkVersion 22
-    buildToolsVersion "22.0.1"
+    compileSdkVersion 25
+    buildToolsVersion '25.0.2'
 
     defaultConfig {
         minSdkVersion 14
-        targetSdkVersion 22
+        targetSdkVersion 25
```
ataulm: :x: this'll make apps' targetSdkVersion to 25. we can ditch this line entirely

library/src/main/java/com/novoda/downloadmanager/lib/Constants.java
```diff
@@ -130,7 +130,7 @@
     /**
      * The minimum amount of progress that has to be done before the progress bar gets updated
      */
-    public static final int MIN_PROGRESS_STEP = 4096;
+    public static final int MIN_PROGRESS_STEP = BUFFER_SIZE;
```
ataulm: 👍

demo-extended/build.gradle
```diff
@@ -29,6 +29,6 @@ android {
 
 dependencies {
     compile project(':library')
-    compile 'com.android.support:appcompat-v7:22.2.1'
-    compile 'com.android.support:recyclerview-v7:22.2.1'
+    compile 'com.android.support:appcompat-v7:25.2.0'
+    compile 'com.android.support:recyclerview-v7:25.2.0'
```
blundell: these version updates could have been done in a separate PR and the review and merge of that would have been pretty fast - now they are stuck next to this other code that i imagine will be discussed at length. Maybe look at splitting this PR?

zegnus:
On my way home I came up with a possible solution much simpler than this PR (and so this is why I closed this PR), will open a different PR if it works.

zegnus:
This PR is superseded by https://github.com/novoda/download-manager/pull/213

library/build.gradle
rock3r: That would only happen in the (very unlikely) case in which the app didn't specify it itself. But yeah we could get rid of it.

PR #213, opened 2017-03-21
Restart service on disconnection

zegnus:
### Problem description

A client targeting android N will make the current DownloadManager to fail to reconnect when there is a network change.

This is due to that the permission `android.net.conn.CONNECTIVITY_CHANGE` is being deprecated in favour of using a `JobScheduler`. For reference [connectivityManager](https://developer.android.com/reference/android/net/ConnectivityManager.html#CONNECTIVITY_ACTION) and [android N background optimizations](https://developer.android.com/about/versions/nougat/android-7.0.html#background_optimizations)

We can still subscribe a broadcast event for `CONNECTIVITY_CHANGE` programatically but the application will only receive events when its in foreground.

### Solution

Connectivity reconnections are part of the download manager's responsibility and the solution has been designed with the objective of keeping this functionality in place for all android versions.

The network change rescheduling is going to be guaranteed by the following mechanisms:

For all devices < Android M the broadcast event will be used
For all devices >= Android M a programmatically set up network change notification will be used instead.

A job will be rescheduled automatically when the connectivity fails. The job is set up to be executed between one and two seconds. If after two seconds the network is not available the job will run in a linear fashion back-off mode of 5 seconds.

The library that we have chosen is the [`JobScheduler`](https://github.com/evernote/android-job) library from `Evernote` as provides backward compatibility in one library.

Due to the nature of the job scheduler and the back-off mechanism, the previously two described events plus the usage of the job scheduler guarantees that the download will be resume as soon as possible. Either trusting the validity of the notification changes or the job scheduling.

As a general advice, it is recommended that the clients use a unique instance of the DownloadManager living in their Application class.

### DIFF
--- a/demo-extended/build.gradle
+++ b/demo-extended/build.gradle
@@ -3,14 +3,14 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.0'
+        classpath 'com.android.tools.build:gradle:1.2.3'
     }
 }
 apply plugin: 'com.android.application'
 
 android {
     compileSdkVersion 22
-    buildToolsVersion '25.0.0'
+    buildToolsVersion "22.0.1"
 
     defaultConfig {
         applicationId "com.novoda.downloadmanager.demo.extended"
--- a/demo-extended/build.gradle
+++ b/demo-extended/build.gradle
@@ -3,19 +3,19 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.0'
+        classpath 'com.android.tools.build:gradle:1.2.3'
     }
 }
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 25
-    buildToolsVersion '25.0.2'
+    compileSdkVersion 22
+    buildToolsVersion "22.0.1"
 
     defaultConfig {
         applicationId "com.novoda.downloadmanager.demo.extended"
         minSdkVersion 14
-        targetSdkVersion 25
+        targetSdkVersion 22
         versionCode 1
         versionName "1.0"
 
@@ -29,6 +29,6 @@ android {
 
 dependencies {
     compile project(':library')
-    compile 'com.android.support:appcompat-v7:25.2.0'
-    compile 'com.android.support:recyclerview-v7:25.2.0'
+    compile 'com.android.support:appcompat-v7:22.2.1'
+    compile 'com.android.support:recyclerview-v7:22.2.1'
 }
--- a/demo-extended/build.gradle
+++ b/demo-extended/build.gradle
@@ -3,19 +3,19 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.2.3'
+        classpath 'com.android.tools.build:gradle:2.3.0'
     }
 }
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 22
-    buildToolsVersion "22.0.1"
+    compileSdkVersion 25
+    buildToolsVersion '25.0.2'
 
     defaultConfig {
         applicationId "com.novoda.downloadmanager.demo.extended"
         minSdkVersion 14
-        targetSdkVersion 22
+        targetSdkVersion 25
         versionCode 1
         versionName "1.0"
 
@@ -29,6 +29,6 @@ android {
 
 dependencies {
     compile project(':library')
-    compile 'com.android.support:appcompat-v7:22.2.1'
-    compile 'com.android.support:recyclerview-v7:22.2.1'
+    compile 'com.android.support:appcompat-v7:25.2.0'
+    compile 'com.android.support:recyclerview-v7:25.2.0'
 }
--- a/demo-extended/build.gradle
+++ b/demo-extended/build.gradle
@@ -3,14 +3,14 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.2.3'
+        classpath 'com.android.tools.build:gradle:2.3.0'
     }
 }
 apply plugin: 'com.android.application'
 
 android {
     compileSdkVersion 22
-    buildToolsVersion "22.0.1"
+    buildToolsVersion '25.0.0'
 
     defaultConfig {
         applicationId "com.novoda.downloadmanager.demo.extended"
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
@@ -47,20 +47,22 @@ public class DownloadManagerBuilder {
 
     @TargetApi(Build.VERSION_CODES.LOLLIPOP)
     private void updateOnNetworkChanges(final DownloadManager downloadManager) {
-        if (deviceDoesNotSupportConnectivityChangesBroadcast()) {
-            ConnectivityManager connectivityManager = (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
-            NetworkRequest networkRequest = new NetworkRequest.Builder().addCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET).build();
-            connectivityManager.registerNetworkCallback(networkRequest, new ConnectivityManager.NetworkCallback() {
-                @Override
-                public void onAvailable(Network network) {
-                    Log.v("network is active now from inside the download manager");
-                    downloadManager.forceStart();
-                }
-            });
+        if (deviceSupportsConnectivityChangesBroadcast()) {
+            return;
         }
+
+        ConnectivityManager connectivityManager = (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
+        NetworkRequest networkRequest = new NetworkRequest.Builder().addCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET).build();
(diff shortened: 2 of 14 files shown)


demo-extended/build.gradle
```diff
@@ -3,14 +3,14 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.2.3'
+        classpath 'com.android.tools.build:gradle:2.3.0'
```
blundell: https://github.com/novoda/download-manager/pull/212/files#r107037244

library/src/main/java/com/novoda/downloadmanager/lib/GlobalState.java
```diff
@@ -4,7 +4,7 @@
 
 import com.novoda.downloadmanager.lib.logger.LLog;
 
-class GlobalState {
+public class GlobalState {
```
blundell: this means clients of the library can see this class as well. Is that wise?

demo-extended/build.gradle
zegnus: I think that I was forced to do this, let me check

library/src/main/java/com/novoda/downloadmanager/lib/jobscheduler/DownloadJob.java
```diff
@@ -0,0 +1,47 @@
...
+
+import com.evernote.android.job.Job;
+import com.evernote.android.job.JobRequest;
+import com.novoda.downloadmanager.lib.DownloadService;
+import com.novoda.downloadmanager.lib.GlobalState;
+import com.novoda.downloadmanager.lib.logger.LLog;
+
+import java.util.concurrent.TimeUnit;
+
+public class DownloadJob extends Job {
```
blundell: I would keep this in the `lib` package - are you expecting clients of the library to instantiate this? (i don't think so)

zegnus: Yup I'll move it so that I can also revert the GlobalState visibility change

library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
```diff
@@ -12,6 +19,8 @@
  */
 public class DownloadManagerBuilder {
 
+    private static final int ANDROID_N = 24;
```
Mecharyry: Why not using `Build.VERSION_CODES` like below?

zegnus: Not available in the current build version dependencies

library/src/main/java/com/novoda/downloadmanager/lib/jobscheduler/DownloadJob.java
zegnus: https://github.com/novoda/download-manager/pull/213/commits/088c57a1bdf1a12c60f5014a9caccde29a3ca3ea

demo-extended/build.gradle
zegnus: https://github.com/novoda/download-manager/pull/213/commits/20ae14c7e8041622e7f4dddc7022bc93929004d8

library/src/main/java/com/novoda/downloadmanager/lib/GlobalState.java
zegnus: https://github.com/novoda/download-manager/pull/213/commits/088c57a1bdf1a12c60f5014a9caccde29a3ca3ea

library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
rock3r: Maybe we should bump the target SDK version given we code against it and the changes are specifically to support N+

zegnus: Some comments were against any update:

https://github.com/novoda/download-manager/pull/212/files#r107037244

https://github.com/novoda/download-manager/pull/212/files#r107033372

blundell: my comments weren't against updates - they where against two separate things happening in 1 PR

library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
```diff
@@ -31,7 +40,27 @@ public DownloadManagerBuilder withVerboseLogging() {
...
         ContentResolver contentResolver = context.getContentResolver();
-        return new DownloadManager(context, contentResolver, verboseLogging);
+        DownloadManager downloadManager = new DownloadManager(context, contentResolver, verboseLogging);
+        updateOnNetworkChanges(downloadManager);
+        return downloadManager;
+    }
+
+    @TargetApi(Build.VERSION_CODES.LOLLIPOP)
+    private void updateOnNetworkChanges(final DownloadManager downloadManager) {
+        if (deviceDoesNotSupportConnectivityChangesBroadcast()) {
```
blundell: use a guard clause for easier reading + 1 less level of indentation & preferring the positive

```

if (deviceSupportsConnectivityChangesBroadcast()) {
 return;
}

ConnectivityManager connectivityManager = (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
             NetworkRequest networkRequest = new NetworkRequest.Builder().addCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET).build();
            connectivityManager.registerNetworkCallback(networkRequest, new ConnectivityManager.NetworkCallback() {
                 @Override
                 public void onAvailable(Network network) {
                     Log.v("network is active now from inside the download manager");
                     downloadManager.forceStart();
                 }
             });
```

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManagerJobCreator.java
```diff
@@ -0,0 +1,15 @@
+package com.novoda.downloadmanager.lib;
+
+import com.evernote.android.job.Job;
+import com.evernote.android.job.JobCreator;
+
+class DownloadManagerJobCreator implements JobCreator {
+    @Override
+    public Job create(String tag) {
```
blundell: where is this method called I can't see it?

library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
rock3r: I don't think either comment is against updates; besides Paul's, Ataul is saying to remove the target SDK and leave that to the library users (you can't remove the `minSdkVersion` tho).

library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
```diff
@@ -31,7 +40,27 @@ public DownloadManagerBuilder withVerboseLogging() {
...
+            ConnectivityManager connectivityManager = (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
+            NetworkRequest networkRequest = new NetworkRequest.Builder().addCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET).build();
+            connectivityManager.registerNetworkCallback(networkRequest, new ConnectivityManager.NetworkCallback() {
+                @Override
+                public void onAvailable(Network network) {
+                    Log.v("network is active now from inside the download manager");
```
blundell: here you used `Log` elsewhere you used `LLog`

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManagerJobCreator.java
zegnus: https://github.com/novoda/download-manager/blob/restart-service-on-disconnection/library/src/main/java/com/novoda/downloadmanager/lib/DownloadManagerJobCreator.java#L9

zegnus: I though you were referring to the tag.

That method is called by the jobScheduler library through the library instantiation `JobManager.create(context).addJobCreator(jobCreator);`

library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
zegnus: https://github.com/novoda/download-manager/pull/213/commits/34430a640db99374b6ad3c94d04e107b3f641bf2

library/src/main/java/com/novoda/downloadmanager/lib/DownloadJob.java
```diff
@@ -0,0 +1,45 @@
...
+    public static final String TAG = "download_job_tag";
+
+    private static final long BACKOFF_MILLIS = TimeUnit.SECONDS.toMillis(5);
+    private static final long EXECUTION_START_MILLIS = TimeUnit.SECONDS.toMillis(1);
+    private static final long EXECUTION_END_MILLIS = TimeUnit.SECONDS.toMillis(10);
+
+    @NonNull
+    @Override
+    protected Result onRunJob(Params params) {
+        Context context = GlobalState.getContext();
```
blundell: can't this be null in some circumstances? What if a job is scheduled then the app has been killed and wiped

zegnus: I have tried to crash it but with no success. What I think that is happening is that the context is held by a static field. Then the jobScheduler is instantiated by the same context so that as long as the job is alive (scheduled) the context will always have a non null reference.

PR #216, opened 2017-03-31
v2/ Fix paused downloads

Mecharyry:
#### Problem
When queueing downloads too quickly, the downloads would hang. 

`DownloadUpdater` would iterate through the list of all downloads, submitting the next queued download to the executor. It avoided submitting the same download twice by checking the `Download.stage` was not already `SUBMITTED`. 

This didn't work because we stored `SUBMITTED` in the database but the `write/read` did not happen fast enough. 

#### Solution
Instead of storing into the `database` we keep a list of `SUBMITTED` downloads in a `List` in the `Service`.

#### Screen Captures

Empty Before | Empty After
---|---
before_dm_empty | after_dm_empty

Before | After
---|---
before_dm_downloading | after_dm


library/src/main/java/com/novoda/downloadmanager/domain/DownloadStatus.java
```diff
@@ -11,7 +11,6 @@
     public static DownloadStatus from(DownloadStage stage) {
         switch (stage) {
             case QUEUED:
-            case SUBMITTED:
```
ouchadam: a `DownloadStatus` should *always*  link to a Download stage, otherwise downloads will disappear/flicker/become statusless

library/src/main/java/com/novoda/downloadmanager/domain/DownloadFile.java
```diff
@@ -12,15 +12,13 @@
     private final long totalSize;
     private final String localUri;
     private final FileStatus fileStatus;
-    private final String fileIdentifier;
```
ouchadam: Deffo need to keep the individual file identifiers. How else would you know what the individual files are? File uris are not ids!

ouchadam:
>This didn't work because we stored SUBMITTED in the database but the write/read did not happen fast enough.

is pork pies

The database write/reads are synchronous
 
```
downloadHandler.setDownloadSubmitted(downloadId);
downloadHandler.isSubmitted(downloadId); // is true
```

It sounds like the bug is due to us inspecting a previous snapshot instead requerying all downloads after changing the status

library/src/main/java/com/novoda/downloadmanager/domain/DownloadStatus.java
ataulm: and this still holds true - a Download will go from QUEUED -> RUNNING.

The submitted state is a pseudo-state that only needs to exist as a concept for whoever is submitting the downloads - doesn't need to be persisted.

library/src/main/java/com/novoda/downloadmanager/domain/DownloadFile.java
ataulm: They are IDs. The local URIs must be unique otherwise you'll get into an unstable state when you try to download two different (or same) remote files to the same local path. Since the local URIs must be unique, they're sufficient as IDs.

PR #217, opened 2017-04-04
v2 Fix hanging downloads

ataulm:
#### Problem
When queueing downloads too quickly, the downloads would hang. 

The Service was spawning multiple download threads, meaning unexpectedly asynchronous operations were being performed, for example:

`DownloadUpdater` would iterate through the list of all downloads, submitting the next queued download to the executor. It avoided submitting the same download twice by checking the `Download.stage` was not already `SUBMITTED`, but this relied on the whole process being synchronous.

#### Solution
We only create one thread per Service now, and we ensure that the Service can die by unbinding from it properly.

#### Screen Captures

Empty Before | Empty After
---|---
before_dm_empty | after_dm_empty

Before | After
---|---
before_dm_downloading | after_dm

#### Paired with
@Mecharyry


.gitignore
```diff
@@ -25,12 +25,12 @@ proguard/
...
 *.log
 
-# IDEA Ignores
+# IDE Ignores
 *.iml
 *.ipr
 *.iws
 .idea/
-.gradle/
+captures/
```
ataulm: When you generate memory dumps via AS, we don't want to store these.

library/src/main/assets/migrations/1_SETUP.SQL
```diff
@@ -1,15 +1,15 @@
 CREATE TABLE IF NOT EXISTS 'request' (
                                 _id INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
                                 request_timestamp TEXT NOT NULL,
-                                UNIQUE (request_timestamp) ON CONFLICT REPLACE
+                                UNIQUE (request_timestamp) ON CONFLICT REPLACE --TODO: should abort this - else risk two downloads with same ID in DatabaseInteractions
```
ouchadam: worth pulling out and replacing with github issues?

library/src/main/java/com/novoda/downloadmanager/DatabaseInteraction.java
```diff
@@ -1,21 +1,22 @@
...
 import com.novoda.downloadmanager.Provider;
 import com.novoda.downloadmanager.demo.simple.DB;
-import com.novoda.downloadmanager.domain.Download;
-import com.novoda.downloadmanager.domain.DownloadFile;
-import com.novoda.downloadmanager.domain.DownloadId;
-import com.novoda.downloadmanager.domain.DownloadRequest;
-import com.novoda.downloadmanager.domain.DownloadStage;
-import com.novoda.downloadmanager.domain.DownloadStatus;
-import com.novoda.downloadmanager.domain.ExternalId;
+import com.novoda.downloadmanager.Download;
```
ouchadam: shouldn't there be no imports?

library/src/main/java/com/novoda/downloadmanager/DatabaseInteraction.java
```diff
@@ -115,17 +129,28 @@ public void submitRequest(DownloadRequest downloadRequest) {
         DB.Download.setDownloadIdentifier(downloadRequest.getExternalId().asString(), values);
         contentResolver.insert(Provider.DOWNLOAD, values);
 
+        ensureLocalFileUriIsUniqueForAllFilesIn(downloadRequest);
+
         ContentValues[] fileValues = createFileValues(downloadRequest.getFiles(), downloadId);
         contentResolver.bulkInsert(Provider.FILE, fileValues);
     }
 
+    private void ensureLocalFileUriIsUniqueForAllFilesIn(DownloadRequest downloadRequest) {
```
ouchadam: does this throw when submitting the request via the `downloader`? (if so good!)

library/src/main/java/com/novoda/downloadmanager/DownloadObserver.java
```diff
@@ -32,8 +29,8 @@ void release() {
         contentResolver.unregisterContentObserver(contentObserver);
     }
 
-    interface OnUpdate {
-        void onUpdate();
+    interface Callback {
```
ouchadam: this is more of a `listener` than callback since multiple events will be triggered

ouchadam:
only the file identifier to readd and then this LGTM ! 👍 awesome work

library/src/main/java/com/novoda/downloadmanager/DownloadObserver.java
ataulm: any documentation that proves this definition and I'll change :D

library/src/main/java/com/novoda/downloadmanager/DownloadRequest.java
```diff
@@ -27,60 +27,41 @@ public ExternalId getExternalId() {
...
     }
 
-    public static class File {
+    public static final class File {
 
         private final String uri;
         private final String localUri;
-        private final String identifier;
 
-        public File(String uri, String localUri, String identifier) {
```
ataulm: we really really don't. any reason?

library/src/main/java/com/novoda/downloadmanager/DatabaseInteraction.java
ataulm: yes it does

library/src/main/java/com/novoda/downloadmanager/DatabaseInteraction.java
ataulm: :thinking:

library/src/main/assets/migrations/1_SETUP.SQL
ataulm: yep maybe in another PR since there's other TODOs you left in v2 branch anyhow

library/src/main/java/com/novoda/downloadmanager/DownloadRequest.java
ataulm: we discussed offline - will add this back (not super convinced :joy:).

library/src/main/assets/migrations/1_SETUP.SQL
ouchadam: 👍 would be good to move all of them to issues and perhaps drop all the v1 issues as well

PR #230, opened 2017-04-26
Fix batch failure and locked to pause

zegnus:
### Problem
We have spotted two issues:
1. Sometimes when a download is in progress and we lose connection slowly, the full batch is marked as failed. But if we regain connection the download should just continue.
2. When pausing a download, sometimes the batch stays stuck in `queued` mode and the download don't continue. The download should be paused and we should be able to resume it.

### Solution
1. We have analysed the error status when the batch is marked as failure. There is a connection and the response status code is 404. We have added 404 to the list of retriable statuses and this fixes the issue. All retriable downloads will finally fail after a few trials. This is a one line change fix.

2. The `paused by application` status is checked once every second in order to do not hammer the client at every written chunk. It could be possible that the user clicks on pause, the check is inside one second (so that isn't executed) and in the write the file finishes to download. In this case two things happens:
 1. The pause has marked all the files to control = 1 (paused) immediately
 2. The file has finished with 200, the batch is marked as pending for the next iteration to restart
The next iteration grabs the pending queue, but because the file is marked as paused, the batch is stuck.

The solution has been to check if there is any file paused when the file finishes completely with success. If that's the case, the final status is going to be paused, instead of success.

Paired with @takecare

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadTask.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadTask.java
@@ -281,7 +281,7 @@ class DownloadTask implements Runnable {
             finalizeDestinationFile(state);
             finalStatus = DownloadStatus.SUCCESS;
 
-            if (batchHasPausedFiles(originalDownloadBatch.getBatchId())) { // check if a file is paused by app and set PAUSED_BY_APP to finalStatus
+            if (batchHasPausedFiles(originalDownloadBatch.getBatchId())) {
                 finalStatus = DownloadStatus.PAUSED_BY_APP;
             }
         } catch (StopRequestException error) {
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadTask.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadTask.java
@@ -281,9 +281,7 @@ class DownloadTask implements Runnable {
             finalizeDestinationFile(state);
             finalStatus = DownloadStatus.SUCCESS;
 
-            // check if a file is paused by app and set PAUSED_BY_APP to finalStatus
-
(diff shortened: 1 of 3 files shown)


library/src/main/java/com/novoda/downloadmanager/lib/DownloadTask.java
```diff
@@ -887,6 +902,7 @@ private static long getHeaderFieldLong(URLConnection conn, String field, long de
     private static boolean isStatusRetryable(int status) {
         switch (status) {
             case HTTP_DATA_ERROR:
+            case HTTP_NOT_FOUND:
```
jackSzm: 👍

library/src/main/java/com/novoda/downloadmanager/lib/DownloadTask.java
```diff
@@ -279,6 +280,10 @@ private void runInternal() {
 
             finalizeDestinationFile(state);
             finalStatus = DownloadStatus.SUCCESS;
+
+            if (batchHasPausedFiles(originalDownloadBatch.getBatchId())) {
+                finalStatus = DownloadStatus.PAUSED_BY_APP;
```
jackSzm: You first set `finalStatus` as SUCCESS and then have an `if` to change it to `PAUSED_BY_THE_APP`. How about having it in the if - else manner ?

takecare: good q. honestly, in practice it shouldn't be too different?

jackSzm: nah , in practice no difference , but imho lil bit easier to follow :)

takecare: done

PR #233, opened 2017-06-12
Null check on downloadsCursor

mcroghan:
Hi folks,

Submitting a fix (I'm pretty sure - I can't actually repro but the stack trace is pretty clear) for an issue that's been causing a considerable number of crashes in our apps). Thanks for your consideration. (Also, sorry for the false start a bit earlier when I submitted a PR from my company's repo - I need to use *that* fork as the actual library for our apps until/unless this gets merged in, so I want to keep it separate from this PR.)

Thanks again,
Mike Croghan

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadsRepository.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadsRepository.java
@@ -55,7 +55,9 @@ class DownloadsRepository {
 
             return downloads;
         } finally {
-            if (downloadsCursor != null) downloadsCursor.close();
+            if (downloadsCursor != null) {
+                downloadsCursor.close();
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


stefanhoth:

library/src/main/java/com/novoda/downloadmanager/lib/DownloadsRepository.java
```diff
@@ -55,7 +55,7 @@ public DownloadsRepository(SystemFacade systemFacade, ContentResolver contentRes
 
             return downloads;
         } finally {
-            downloadsCursor.close();
+            if (downloadsCursor != null) downloadsCursor.close();
```
tasomaniac: I'm not sure if this repo has different code style but we normally don't do one liner if statements.

mcroghan: Gotcha - updated.

Issue #209, opened 2017-02-25
download fails if app removed from recent list


Hackmodford:
I have setup my app to only allow 1 concurrent download.
If I start a download and wait for it to have progress, then remove the app from the recent list I get some strange behavior.

The notification gets removed a few seconds after doing this. Which to me looks like the download gets paused/stopped. Then the notification comes back like it is trying to resume the download, but it fails to resume the download. Then it proceeds to begin downloading the next file in the queue.

Is this expected behavior?

stefanhoth:
Thanks for using download manager. To your questions:

1. To my knowledge, removing an app from recents kills the process and with doing so the download in progress. So far I assume expected behaviour.

2. Usually download manager would be able to resume the download. If it can't the likely answer is that the server doesn't support resuming, either because it doesn't allow to jump ahead or because your session expires after one download. Could this be the case here?

Pinging @ouchadam for detailed answers with actual knowledge.

ouchadam:
like @stefanhoth mentioned, the killing of the process also killing the download is expected. This library takes the stock shared system service `DownloadManager` and reimplements it as a app process (I believe the original intention was to allow access to the private app storage).

As for the resuming, are you able to pause/resume under normal conditions? 

It's likely ending the process early puts the database in an inconsistent state.

also worth noting there's an effort to rewrite this library with the same functionality from scratch dropping all the inherited AOSP code.

https://github.com/novoda/download-manager/tree/v2

tasomaniac:
If we use an Android `Service` with posting `Notification` using `startForeground();` method, the process shouldn't be killed when user swipes the app from Recents. 

You can test it in music apps, when you swipe the app, the notification + music playback should continue. 

We can potentially use the same mechanism to do that.

Hackmodford:
Still does this in the latest version. I like @tasomaniac's service idea.

Hackmodford:
Pull request #237 fixes this issue for me.

PR #237, opened 2017-07-31
Bring download service to foreground

joetimmins:
## Problem
On Android O, the downloads service is stopped by the system when the activity using it is no longer present. This is because it is not registered as a foreground service.

## Solution
In order to move the downloads service to the foreground, we need two things - an active download, and the notification for that active download. The downloads service uses a `DownloadNotifier` implementation to delegate all of its notification handling, so it seemed logical to pass in a callback to that class that would move the service to and from the background depending on the state of the current downloads.

### Screenshots
downloads-continue

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
@@ -35,6 +35,7 @@ import android.os.IBinder;
 import android.os.Message;
 import android.os.Process;
 import android.support.annotation.NonNull;
+import android.support.v4.util.SimpleArrayMap;
 
 import com.novoda.downloadmanager.lib.logger.LLog;
 import com.novoda.downloadmanager.notifications.DownloadNotifier;
@@ -452,7 +453,7 @@ public class DownloadService extends Service {
     }
 
     private void updateUserVisibleNotification(Collection<DownloadBatch> batches) {
-        downloadNotifier.updateWith(batches, notificationCreatedCallback);
+        downloadNotifier.updateWith(batches, notificationsCreatedCallback);
     }
 
     @Override
@@ -460,11 +461,21 @@ public class DownloadService extends Service {
         LLog.e("I want to dump but nothing to dump into");
     }
 
-    final DownloadNotifier.NotificationCreatedCallback notificationCreatedCallback = new DownloadNotifier.NotificationCreatedCallback() {
+    final DownloadNotifier.NotificationsCreatedCallback notificationsCreatedCallback = new DownloadNotifier.NotificationsCreatedCallback() {
         @Override
-        public void onNotificationCreated(NotificationTag tag, Notification notification) {
-            if (DownloadStatus.isRunning(tag.status())) {
-                DownloadService.this.startForeground(tag.hashCode(), notification);
+        public void onNotificationCreated(SimpleArrayMap<NotificationTag, Notification> taggedNotifications) {
+            boolean noActiveDownloads = true;
+
+            for (int i = 0; i < taggedNotifications.size(); i++) {
+                NotificationTag currentTag = taggedNotifications.keyAt(i);
+                if (currentTag.status() == DownloadNotifier.TYPE_ACTIVE) {
+                    DownloadService.this.startForeground(currentTag.hashCode(), taggedNotifications.get(currentTag));
+                    noActiveDownloads = false;
+                    break;
+                }
+            }
+            if (noActiveDownloads) {
+                DownloadService.this.stopForeground(false);
             }
         }
     };
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
@@ -17,7 +17,6 @@
 package com.novoda.downloadmanager.lib;
 
 import android.app.AlarmManager;
-import android.app.Notification;
 import android.app.PendingIntent;
 import android.app.Service;
 import android.content.ContentResolver;
@@ -35,12 +34,10 @@ import android.os.IBinder;
 import android.os.Message;
 import android.os.Process;
 import android.support.annotation.NonNull;
-import android.support.v4.util.SimpleArrayMap;
 
 import com.novoda.downloadmanager.lib.logger.LLog;
 import com.novoda.downloadmanager.notifications.DownloadNotifier;
 import com.novoda.downloadmanager.notifications.DownloadNotifierFactory;
-import com.novoda.downloadmanager.notifications.NotificationTag;
 
 import java.io.File;
 import java.io.FileDescriptor;
@@ -53,8 +50,6 @@ import java.util.Map;
 import java.util.concurrent.ExecutorService;
 
 import static android.text.format.DateUtils.MINUTE_IN_MILLIS;
-import static com.novoda.downloadmanager.notifications.DownloadNotifier.NotificationsCreatedCallback;
-import static com.novoda.downloadmanager.notifications.DownloadNotifier.TYPE_ACTIVE;
 
 /**
  * Performs background downloads as requested by applications that use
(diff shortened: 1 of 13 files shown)


library/src/main/java/com/novoda/downloadmanager/lib/DownloadService.java
```diff
@@ -88,6 +88,8 @@
     private NetworkChecker networkChecker;
     private DestroyListener destroyListener;
 
+    private final NotificationCreatedListener notificationsCreatedCallback = new NotificationCreatedListener(this);
```
ouchadam: is it worth `new'ing` up in the `onCreate` to avoid the possibility of the service being interacted with before it has been fully created?

joetimmins: Can do

library/src/main/java/com/novoda/downloadmanager/notifications/SynchronisedDownloadNotifier.java
```diff
@@ -77,92 +72,68 @@ public void notifyDownloadSpeed(long id, long bytesPerSecond) {
...
     }
 
-    private List<Integer> getStaleTagsThatWereNotRenewed(SimpleArrayMap<String, Collection<DownloadBatch>> clustered) {
+    private List<Integer> getStaleTagsThatWereNotRenewed(SimpleArrayMap<NotificationTag, Collection<DownloadBatch>> clusteredBatches) {
         List<Integer> staleTags = new ArrayList<>();
 
         for (int i = activeNotifications.size() - 1; i >= 0; i--) {
-            String tag = activeNotifications.keyAt(i);
-            if (!clustered.containsKey(tag)) {
+            NotificationTag tag = activeNotifications.keyAt(i);
```
ouchadam: 👍

library/src/main/java/com/novoda/downloadmanager/notifications/NotificationTag.java
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Build tag used for collapsing several {@link DownloadBatch} into a single
+ * {@link Notification}.
+ */
+public class NotificationTag {
+
+    private final int status;
+    private final String identifier;
+
+    public static NotificationTag create(DownloadBatch batch, String packageName) {
```
Mecharyry: We return `null` in the `else`, we could make this super explicit to the clients of this method and add the `@Nullable` annotation.

library/src/main/java/com/novoda/downloadmanager/notifications/DownloadNotifier.java
```diff
@@ -1,13 +1,21 @@
...
+    int TYPE_SUCCESS = 3;
+    int TYPE_FAILED = 4;
+    int TYPE_CANCELLED = 5;
+
     void cancelAll();
 
     void notifyDownloadSpeed(long id, long bytesPerSecond);
 
-    void updateWith(Collection<DownloadBatch> batches);
+    void updateWith(Collection<DownloadBatch> batches, NotificationsCreatedListener notificationsCreatedCallback);
```
blundell: `NotificationsCreatedListener notificationsCreatedCallback`

is it a listener or a callback ... :-)

library/src/main/java/com/novoda/downloadmanager/notifications/NotificationTag.java
```diff
@@ -0,0 +1,68 @@
...
+            // Failed downloads always have unique notifications
+            return new NotificationTag(DownloadNotifier.TYPE_FAILED, String.valueOf(batch.getBatchId()));
+        } else if (batch.isCancelled() && batch.shouldShowCompletedItem()) {
+            // Cancelled downloads always have unique notifications
+            return new NotificationTag(DownloadNotifier.TYPE_CANCELLED, String.valueOf(batch.getBatchId()));
+        } else if (batch.isSuccess() && batch.shouldShowCompletedItem()) {
+            // Complete downloads always have unique notifications
+            return new NotificationTag(DownloadNotifier.TYPE_SUCCESS, String.valueOf(batch.getBatchId()));
+        } else {
+            return null;
```
blundell: want to add the nullable annotation to the method?

Also I notice this class and this method is public, so do you expect client libraries to call it? If so it'd be good if you add javadoc to explain its use (and document the null return)

library/src/main/java/com/novoda/downloadmanager/notifications/NotificationTag.java
```diff
@@ -0,0 +1,68 @@
...
+
+import android.app.Notification;
+
+import com.novoda.downloadmanager.lib.DownloadBatch;
+
+/**
+ * Build tag used for collapsing several {@link DownloadBatch} into a single
+ * {@link Notification}.
+ */
+public class NotificationTag {
```
Mecharyry: Generating a `toString` would make it easier to debug should anything go wrong that involves the `NotificationTag`.

library/src/main/java/com/novoda/downloadmanager/notifications/DownloadNotifier.java
```diff
@@ -1,13 +1,21 @@
 package com.novoda.downloadmanager.notifications;
 
 import com.novoda.downloadmanager.lib.DownloadBatch;
+import com.novoda.downloadmanager.lib.NotificationsCreatedListener;
 
 import java.util.Collection;
 
 public interface DownloadNotifier {
+    int TYPE_ACTIVE = 1;
```
blundell: moving these makes them publicly visible to clients, was this the intention?

library/src/test/java/com/novoda/downloadmanager/lib/NotificationsCreatedListenerTest.java
```diff
@@ -0,0 +1,81 @@
...
+    private NotificationsCreatedListener listener;
+
+    @Before
+    public void setUp() {
+        initMocks(this);
+        listener = new NotificationsCreatedListener(service);
+    }
+
+    @Test
+    public void givenActiveDownload_whenNotificationCreated_thenBringsServiceToForeground() {
```
blundell: you did add tests! Even tho the PR said you didn't :-) oh you

library/src/main/java/com/novoda/downloadmanager/notifications/DownloadNotifier.java
joetimmins: It's a listener now, used to be a callback though

library/src/main/java/com/novoda/downloadmanager/lib/NotificationsCreatedListener.java
```diff
@@ -0,0 +1,33 @@
...
+
+import android.app.Notification;
+import android.app.Service;
+import android.support.v4.util.SimpleArrayMap;
+
+import com.novoda.downloadmanager.notifications.NotificationTag;
+
+import static com.novoda.downloadmanager.notifications.DownloadNotifier.TYPE_ACTIVE;
+
+public class NotificationsCreatedListener {
```
Mecharyry: This can't be moved unless the packages are collapsed. It's created in `lib` and used in `notifications`. I'd rather the constructor be package the methods accessible because at least the clients can't create it to use it.

library/src/main/java/com/novoda/downloadmanager/notifications/NotificationTag.java
joetimmins: It needs to be public because it's used in two packages. Investigating to see if there's a nicer structure.

library/src/main/java/com/novoda/downloadmanager/notifications/DownloadNotifier.java
joetimmins: They need to be public because they're used in two packages now. Hopefully will be able to move em back into package-private scope shortly

library/src/main/java/com/novoda/downloadmanager/lib/NotificationsCreatedListener.java
joetimmins: Needs to be unfortunately, used in `lib` and `notifications` packages

library/src/test/java/com/novoda/downloadmanager/lib/NotificationsCreatedListenerTest.java
joetimmins: Oh yeah, I forgot I wrote that! Removed it now

joetimmins:
There's no need for an O-specific demo, there's no additional features for O other than it now working. It's backwards compatible too.

Will change the target & compile SDK versions for the demos to `26` though.

joetimmins:
Okay so moving the build tools version and compile SDK to `O` / `26` has caused quite a few warnings in the app. Can we merge this (as it works for build tools `N` / `25`) and I can tackle the new warnings in a separate PR?

blundell:
sure, should definitely update the demo as part of this piece of work 👍

library/src/main/java/com/novoda/downloadmanager/notifications/NotificationsCreatedListener.java
```diff
@@ -0,0 +1,31 @@
...
+import android.app.Notification;
+import android.app.Service;
+import android.support.v4.util.SimpleArrayMap;
+
+import static com.novoda.downloadmanager.notifications.SynchronisedDownloadNotifier.TYPE_ACTIVE;
+
+public class NotificationsCreatedListener {
+    private Service service;
+
+    public NotificationsCreatedListener(Service service) {
```
blundell: if it has to stay public, can we javadoc it then? (even if its just to say _not expected to be used by clients_)

tasomaniac:
This might be also fixing #209

Hackmodford:
I believe you're right @tasomaniac 
I will do some testing to confirm this.

tasomaniac:
Danke ☺

PR #238, opened 2017-08-08
Target Android O when building the library

joetimmins:
## Problem
Android O contains significant breaking changes to the way notifications work. It also makes sure that long running services have to be either a) foreground services with a visible notification, or b) killed with extreme prejudice. In order to make sure `download-manager` can handle these differences, we need to compile against Android O.

## Solution
Bump the `compileSdkVersion` and `buildToolsVersion` in the library's build.gradle file to `26`. This causes a load of error toasts to appear when the demo apps are downloading things. In order to fix those, we need to make sure we set a `NotificationChannel` on both the `NotificationManager` we are using and each individual `Notification` we fire through the `NotificationManager`. 

I also added the `NotificationChannelProvider` interface to `DownloadManagerModules` to enable clients of the library to set their own `NotificationChannel`s.


demo-extended/build.gradle
```diff
@@ -9,13 +9,13 @@ buildscript {
...
-    compileSdkVersion 25
-    buildToolsVersion '25.0.2'
+    compileSdkVersion 26
+    buildToolsVersion '26.0.1'
 
     defaultConfig {
         applicationId "com.novoda.downloadmanager.demo.extended"
         minSdkVersion 14
-        targetSdkVersion 25
+        targetSdkVersion 26
```
Mecharyry: @ataulm did you say we should remove the `targetSdkVersion`? I honestly don't remember.

library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
```diff
@@ -15,7 +15,6 @@
 /**
  * A client can specify whether the downloads are allowed to proceed by implementing
  * {@link com.novoda.downloadmanager.lib.DownloadClientReadyChecker} on your Application class
- *
```
Mecharyry: 😆

library/src/main/java/com/novoda/downloadmanager/lib/DownloadManagerModules.java
```diff
@@ -84,6 +88,11 @@ public Builder withNotificationImageRetrieverFactory(@NonNull NotificationImageR
             return this;
         }
 
+        public Builder withNotificationManagerCustomiser(@NonNull NotificationChannelProvider notificationChannelProvider) {
```
Mecharyry: The `with` doesn't match the `param`, was there a reason for this?

joetimmins: Just an oversight, will fix

demo-extended/build.gradle
joetimmins: The apps work fine targeting either 25 or 26, I left them as 26 to make the build.gradle files free of warnings.

library/src/main/java/com/novoda/downloadmanager/notifications/NotificationDisplayer.java
```diff
@@ -80,6 +85,19 @@ public Notification buildAndShowNotification(NotificationTag tag, Collection<Dow
         return notification;
     }
 
+    private NotificationCompat.Builder initNotificationBuilder() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            NotificationChannel notificationChannel = notificationChannelProvider.getNotificationChannel();
+            if (!notificationManager.getNotificationChannels().contains(notificationChannel)) {
```
Mecharyry: Worth adding the `contains` to the `notificationManager`?

joetimmins: We don't own `notificationManager`, it's Google's class

Mecharyry: Ah right, my bad 😄

demo-extended/build.gradle
ataulm: this is the demo so it's fine (and correct) - it's the core library that we export that should not have target specified

PR #239, opened 2017-08-10
Revert to API 25 (Nougat)

joetimmins:
## Problem
Moving to Android O (API 26) for the library means including the `26.0.1` support library. This unfortunately means any client apps cannot use any other support libraries, effectively forcing them to move to API 26, even though this is not strictly enforced. This is undesirable.

## Solution
Remove any Android O specific APIs, revert compileSdk and support libs to `25`, and make it easy to move to supporting Android O again in future. This meant including the `NotificationInitialiser` interface as part of `DownloadManagerModules`. When we want to target Android O again, any client app can put all O-specific APIs in a class implementing that interface.


library/src/main/java/com/novoda/downloadmanager/lib/DefaultNotificationInitialiser.java
```diff
@@ -0,0 +1,14 @@
...
+import android.app.NotificationManager;
+import android.content.Context;
+import android.support.v4.app.NotificationCompat;
+
+import com.novoda.downloadmanager.notifications.NotificationInitialiser;
+
+class DefaultNotificationInitialiser implements NotificationInitialiser {
+    @Override
+    public NotificationCompat.Builder initNotificationBuilder(Context context, NotificationManager notificationManager) {
+        return new NotificationCompat.Builder(context);
```
joetimmins: Right now it looks a bit weird to have `NotificationManager` passed in here as it's never used, but when an app that targets Android O wants to implement this interface, they will need to call `setNotificationChannel` on the `NotificationManager` as well as the `Builder` they will return, or their notifications won't work.

blundell:
what out of 26.0.1 was being used for O ooi?

joetimmins:
Having a buildToolsVersion of `25` means that having the `26` support lib causes a compile error. We need the support lib core utils in the `download-manager` library for `SimpleArrayMap`, amongst other things, which the `download-manager` uses internally as a holdover from when it was part of AOSP.

library/src/main/java/com/novoda/downloadmanager/lib/DefaultNotificationInitialiser.java
```diff
@@ -0,0 +1,14 @@
+package com.novoda.downloadmanager.lib;
+
+import android.app.NotificationManager;
+import android.content.Context;
+import android.support.v4.app.NotificationCompat;
+
+import com.novoda.downloadmanager.notifications.NotificationInitialiser;
+
+class DefaultNotificationInitialiser implements NotificationInitialiser {
```
zegnus: What do you think about renaming the interface to `NotificationCompatBuilderCreator`?

joetimmins: In the next PR, when we reintroduce Android O, there will be an implementation of this interface that modifies something on the `NotificationManager`, so it will do more than just create a `Builder` object. Also having both `Builder` and `Creator` in the name looks confusing 😜

zegnus: fair enough 👍

joluet:
Any news on this? We're using this library and it currently prevents us from targeting Android O as this introduces a crash (#244).

Mecharyry:
@joluet we are currently working on v2 of the library. @joetimmins do we have any plans to try and tackle this problem in the meantime?

joluet:
@Mecharyry @joetimmins This is getting more and more urgent. It would be great if you could release a workaround that allows targeting api level `26`. 

screen shot 2018-05-09 at 11 06 07 am

joelfernandes19:
Any updates on this?

MechaRyRy:
Hey @joluet and @joelfernandes19 I'm working on getting v2 merged this week 😄 just been adding some javadocs and a readme.

joelfernandes19:
That's great news @Mecharyry ! Thanks for the update. Looking forward to it. 

We're closing in on the deadline for app updates to target at least Android 8.0 (API level 26) from November 1, 2018 onward, so was looking at options we have.

MechaRyRy:
@joluet @joelfernandes19 I'm pleased to announce that version 2 of the `download-manager` has been released.

joelfernandes19:
@Mecharyry Awesome! Thanks for the update.

PR #241, opened 2017-08-15
Covering all possible null cursor cases in this DownloadsRepository.java

mcroghan:
Hi folks - despite the null check change I submitted in https://github.com/novoda/download-manager/pull/233, the Google Play console is continuing to report lots of app crashes (I think they're actually background service crashes) originating in this class. I inspected it more carefully this time and noted that there are several places in it - not just the one I noted and fixed the first time I looked at it - where cursors might be null but are treated as if they are known to be non-null. This PR is an attempt to address all of those issues.

I'm not sure this is how you would prefer to address some of these. In particular, the method getDownloadFor() seems never to be called in the current codebase, so I just annotated it @Nullable and made it return null in case of a null cursor, but that may or may not jibe with its intended use.

Interestingly, 99.9% of our crashes for these issues are on Android 6.0 devices, according to the Google Play console. So this must all be due to some underlying curser-related bug that Google blessed us with in Marshmallow, and then fixed for 7.0. So, if you test this, use some 6.0 devices/emulators. I'm doing the same.

Thanks, folks. I should mention, this library has been a lifesaver for our products. Thanks so much for building it and opening it up.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/lib/DownloadsRepository.java
+++ b/library/src/main/java/com/novoda/downloadmanager/lib/DownloadsRepository.java
@@ -5,6 +5,7 @@ import android.content.ContentUris;
 import android.content.ContentValues;
 import android.database.Cursor;
 import android.net.Uri;
+import android.support.annotation.Nullable;
 import android.text.TextUtils;
 
 import com.novoda.downloadmanager.lib.logger.LLog;
@@ -49,7 +50,7 @@ class DownloadsRepository {
             List<FileDownloadInfo> downloads = new ArrayList<>();
             FileDownloadInfo.Reader reader = new FileDownloadInfo.Reader(contentResolver, downloadsCursor);
 
-            while (downloadsCursor.moveToNext()) {
+            while (downloadsCursor != null && downloadsCursor.moveToNext()) {
                 downloads.add(downloadInfoCreator.create(reader));
             }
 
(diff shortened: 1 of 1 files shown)


hal9002:

joetimmins:
Thank you for the pull request @mcroghan. Can I ask you if this is happening on the newest version of the download-manager library (0.3.0)? Also, are there any specific OS versions or devices where this is happening? Not saying we won't look into this, just want to get the full picture 👍

mcroghan:
Hi Joe,

Yep, this is in 0.3.0, and yep, it seems to be exclusively on Android 6.0 devices. (I mentioned that in my PR comment.) Thanks!

- Mike

--
Mike Croghan
Principal Developer
USA TODAY NETWORK
Voice: 571-989-2764
mcroghan@gannett.com

________________________________

library/src/main/java/com/novoda/downloadmanager/lib/DownloadsRepository.java
```diff
@@ -65,15 +66,20 @@ public DownloadsRepository(SystemFacade systemFacade, ContentResolver contentRes
...
     public FileDownloadInfo getDownloadFor(long id) {
         Uri uri = ContentUris.withAppendedId(downloadsUriProvider.getAllDownloadsUri(), id);
         Cursor downloadsCursor = contentResolver.query(uri, null, null, null, null);
-        try {
-            downloadsCursor.moveToFirst();
-            FileDownloadInfo.Reader reader = new FileDownloadInfo.Reader(contentResolver, downloadsCursor);
-            return downloadInfoCreator.create(reader);
-        } finally {
-            downloadsCursor.close();
+        if (downloadsCursor != null) {
```
joetimmins: I think it would probably be clearer to have a quick return if null here - so something like
```
if (downloadsCursor != null) {
    return null;
```
Makes it a bit easier to see exactly what gets returned if the cursor is null and removes a level of indentation too.

library/src/main/java/com/novoda/downloadmanager/lib/DownloadsRepository.java
```diff
@@ -49,7 +50,7 @@ public DownloadsRepository(SystemFacade systemFacade, ContentResolver contentRes
             List<FileDownloadInfo> downloads = new ArrayList<>();
             FileDownloadInfo.Reader reader = new FileDownloadInfo.Reader(contentResolver, downloadsCursor);
 
-            while (downloadsCursor.moveToNext()) {
+            while (downloadsCursor != null && downloadsCursor.moveToNext()) {
```
joetimmins: Similar to below we can move the null check to line 48 and return `Collections.emptyList()` directly at that point

joetimmins:
> Hi Joe, Yep, this is in 0.3.0, and yep, it seems to be exclusively on Android 6.0 devices. (I mentioned that in my PR comment.) Thanks! - Mike

Haha yes so you did. A good lesson for me to *actually read* PR descriptions... 😅

PR #245, opened 2017-10-31
Add v2 base code

Mecharyry:
### Problem
Moving towards version two of download manager.

### Solution
Add the code that @zegnus has kindly provided us from https://github.com/zegnus/lite-download-manager 

The removal of download-manager v1 was done on the epic branch, just a delete all from the `download-manager` package 😂 

**Note** this does not include any unit tests, this is simply replacing the code.

### Screen Capture
download


Electryc:
>Remove the old code and replace

`+4,903 −0`

I don't see remove 😛

Mecharyry:
😂  I removed the code from the epic branch, I thought it would be super not helpful to have it all in here in the end @Electryc

team-props/lint-rules/config.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<lint>
+
+  <issue id="all">
```
Mecharyry: Had to ignore all lint warnings from the generated package, Room seems to have more than a few issues. We can decide to address these later if we decide.

Electryc:
@Mecharyry I see 👍

PR #247, opened 2017-10-31
Example test

Mecharyry:
### Problem
As per the ticket 
> add at least one test

### Solution
Add one test 🤔  add several tests! 😄 

Additionally I've updated all of the libraries that this project uses, might as well start off on the right foot 😄 

Also removed the other dummy tests that weren't needed here 😄


library/src/main/java/com/novoda/downloadmanager/RoomBatch.java
```diff
@@ -4,10 +4,12 @@
 import android.arch.persistence.room.Entity;
 import android.arch.persistence.room.Index;
 import android.arch.persistence.room.PrimaryKey;
+import android.support.annotation.NonNull;
 
 @Entity(indices={@Index("batch_id")})
 class RoomBatch {
 
+    @NonNull
```
ataulm: necessaire?

library/src/main/java/com/novoda/downloadmanager/RoomFile.java
```diff
@@ -14,6 +15,7 @@
 )
 class RoomFile {
 
+    @NonNull
```
ataulm: looks like necessary since you have it in multiple places..

app/build.gradle
```diff
@@ -1,13 +1,13 @@
...
-    compileSdkVersion 26
-    buildToolsVersion "26.0.2"
+    compileSdkVersion 27
+    buildToolsVersion "27.0.0"
 
     defaultConfig {
         applicationId "com.litedownloadmanager.demo"
         minSdkVersion 14
-        targetSdkVersion 26
+        targetSdkVersion 27
```
jonreeve: ✨

library/src/test/java/com/novoda/downloadmanager/CallbackThrottleByProgressIncreaseTest.java
```diff
@@ -0,0 +1,67 @@
...
+        verify(downloadBatchCallback).onUpdate(downloadBatchStatus);
+    }
+
+    private void givenStatusUpdate(DownloadBatchStatus downloadBatchStatus) {
+        given(downloadBatchStatus.percentageDownloaded()).willReturn(DOWNLOAD_PERCENTAGE);
+        callbackThrottleByProgressIncrease.setCallback(downloadBatchCallback);
+        callbackThrottleByProgressIncrease.update(downloadBatchStatus);
+        reset(downloadBatchCallback);
+    }
+
```
ataulm: extra line doesn't seem to match the ExampleUnitTest - dunno which was project convention

library/src/test/java/com/novoda/downloadmanager/CallbackThrottleByProgressIncreaseTest.java
```diff
@@ -0,0 +1,67 @@
...
+import org.junit.rules.ExpectedException;
+
+import static org.mockito.BDDMockito.given;
+import static org.mockito.Mockito.*;
+
+public class CallbackThrottleByProgressIncreaseTest {
+
+    private static final int DOWNLOAD_PERCENTAGE = 75;
+
+    @Rule
```
zegnus: @pablisco doing this doesn't remove the performance advantage of using `mock(...)`?

library/src/main/java/com/novoda/downloadmanager/RoomFile.java
Mecharyry: Yea this happened when I updated to the latest version of `Room` 😢

library/src/test/java/com/novoda/downloadmanager/CallbackThrottleByProgressIncreaseTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Test
+    public void throwsException_whenStoppingUpdates_withoutSettingCallback() {
+        thrown.expect(NullPointerException.class);
+
+        callbackThrottleByProgressIncrease.stopUpdates();
+    }
+
+    @Test
+    public void doesNothing_whenUpdating_withoutSettingCallback() {
```
jonreeve: the underscores seem a little arbitrary... I guess because the names are still quite long. I get that they're not very easy to name tho.

library/src/test/java/com/novoda/downloadmanager/CallbackThrottleByProgressIncreaseTest.java
```diff
@@ -0,0 +1,67 @@
...
+    public void doesNothing_whenDownloadPercentageIsTheSame() {
+        givenStatusUpdate(downloadBatchStatus);
+
+        callbackThrottleByProgressIncrease.update(downloadBatchStatus);
+
+        verifyZeroInteractions(downloadBatchCallback);
+    }
+
+    @Test
+    public void notifiesCallbackWithPreviousStatus_whenStoppingUpdates() {
```
jonreeve: Quite specific about the implementation details... Maybe instead of "notifies callback" we can just say "emits"? emitsLastStatusWhenStoppingUpdates ?

library/src/test/java/com/novoda/downloadmanager/CallbackThrottleByProgressIncreaseTest.java
Mecharyry: They aren't arbitrary 😄  

`expectedBehavior_whenStateUnderTest_furtherClauses` obviously trying to avoid the `FurtherClausesPart` 😄

ataulm: how c/would you name this one?

for me the underscores seem legit, separating each chunk of information. would you have preferred no underscores?

library/src/test/java/com/novoda/downloadmanager/CallbackThrottleByProgressIncreaseTest.java
```diff
@@ -0,0 +1,67 @@
...
+        callbackThrottleByProgressIncrease.stopUpdates();
+
+        verify(downloadBatchCallback).onUpdate(downloadBatchStatus);
+    }
+
+    private void givenStatusUpdate(DownloadBatchStatus downloadBatchStatus) {
+        given(downloadBatchStatus.percentageDownloaded()).willReturn(DOWNLOAD_PERCENTAGE);
+        callbackThrottleByProgressIncrease.setCallback(downloadBatchCallback);
+        callbackThrottleByProgressIncrease.update(downloadBatchStatus);
+        reset(downloadBatchCallback);
```
jonreeve: This seems to be hiding a little too much..

Mecharyry: Recommendations?

library/src/test/java/com/novoda/downloadmanager/CallbackThrottleByProgressIncreaseTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Test
+    public void doesNothing_whenUpdating_withoutSettingCallback() {
+        callbackThrottleByProgressIncrease.update(downloadBatchStatus);
+
+        verifyZeroInteractions(downloadBatchCallback, downloadBatchStatus);
+    }
+
+    @Test
+    public void doesNothing_whenDownloadPercentageIsTheSame() {
```
jonreeve: i.e. doesn't emit any update because the input status was unchanged? doesNotEmitWhenPercentageUnchanged?

library/src/test/java/com/novoda/downloadmanager/CallbackThrottleByProgressIncreaseTest.java
```diff
@@ -0,0 +1,67 @@
...
+        thrown.expect(NullPointerException.class);
+
+        callbackThrottleByProgressIncrease.stopUpdates();
+    }
+
+    @Test
+    public void doesNothing_whenUpdating_withoutSettingCallback() {
+        callbackThrottleByProgressIncrease.update(downloadBatchStatus);
+
+        verifyZeroInteractions(downloadBatchCallback, downloadBatchStatus);
```
jonreeve: hang on, does this check it doesn't call a callback you never set? If I get that right that doesn't sound like a test we should have. If you were just verifying it doesn't blow up, then that shouldn't need this bit

library/src/test/java/com/novoda/downloadmanager/CallbackThrottleByProgressIncreaseTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+    private CallbackThrottleByProgressIncrease callbackThrottleByProgressIncrease;
+
+    @Before
+    public void setUp() {
+        callbackThrottleByProgressIncrease = new CallbackThrottleByProgressIncrease();
+    }
+
+    @Test
+    public void throwsException_whenStoppingUpdates_withoutSettingCallback() {
```
jonreeve: mustHaveCallbackWhenStoppingUpdates ?

library/src/test/java/com/novoda/downloadmanager/CallbackThrottleByProgressIncreaseTest.java
Mecharyry: This verifies that there are zero interactions with any mocks if a callback is not set

library/src/test/java/com/novoda/downloadmanager/CallbackThrottleByProgressIncreaseTest.java
jonreeve: yeah, would prefer getting the names short enough to not have underscores. To me they're a smell the name is still too long

library/src/test/java/com/novoda/downloadmanager/CallbackThrottleByProgressIncreaseTest.java
jonreeve: Yeah I'm still sort of figuring it out... I think I'd need to look at the objects concerned. Thoughts that come to mind in no particular order:
 * is downloadBatchStatus a simple enough object to not need a mock? That could help. Looks like mocking a data object but maybe it's more complex.
 * for now at least I'd just stick the callback setting in each test that does it, in the fixture setup block.
 * maybe the last two lines are worth keeping as a givenPreviousUpdate(X)

library/src/test/java/com/novoda/downloadmanager/CallbackThrottleByProgressIncreaseTest.java
jonreeve: but you gave it no reference to the callback, right? Like, if I instantiate an object and *don't* give it a collaborator, I don't need to check it doesn't call that collaborator. That would be magic if it did. Maybe I'm missing something

ataulm: yeah I think this test is not necessary, either:

- remove the test
- remove the assertion to verify that the CallbackThrottleByProgressIncrease doesn't throw an exception
- throw an exception if a callback is not set and we call update, then assert that

Mecharyry: I'm reluctant to change the code this early on considering I don't know enough about it. I'll remove this test for now and when more is known potentially retest this class.

Mecharyry:
@jonreeve @ataulm I've updated the tests again 😄  take another look and let me know if you are happy yet!

jonreeve:
Sorry, don't mean to be a PITA. I think it's a massive improvement anyway. I'm a bit of a nut for short and fluent language in tests, but frankly so long as we're aiming towards that I'm happy.

PR #248, opened 2017-11-10
Dismiss notification

Mecharyry:
### Problem
We see only a single notification per batch, each notification wipes the previous and we cannot swipe the notification away when the download is completed.

Additionally, when deleting we post a notification that has the current downloaded percentage 😬  since we don't know what we want to do with this we should prevent notification updates when the status is deleted.

### Solution
Provide a unique id for each notification along with a channel id. When a notification is completed we should remove it from the service (detach) and reattach using the `NotificationManager`. This prevents the new notification from a new batch from wiping the old one. 

- Remove the notification whenever the download status is deleted.
- Stack the notification when the download is completed. This allows another download to show a notification without deleting the previous notification.
- Update the notification for all other statuses. 

##### Lollipop
Before | After
--- | ---
before_lollipop | after_lollipop

##### Nougat
Before | After
--- | ---
master mp4 | dm-notification mp4

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadService.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadService.java
@@ -6,5 +6,7 @@ interface DownloadService {
 
     void updateNotification(NotificationInformation notification);
 
-    void makeNotificationDismissible(NotificationInformation notificationInformation);
+    void stackNotification(NotificationInformation notificationInformation);
+
+    void dismissNotifications();
 }
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadService.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadService.java
@@ -8,5 +8,5 @@ interface DownloadService {
 
     void stackNotification(NotificationInformation notificationInformation);
 
-    void dismissNotifications();
+    void dismissNotification();
 }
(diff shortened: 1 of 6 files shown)


library/src/main/java/com/novoda/downloadmanager/DownloadService.java
```diff
@@ -6,5 +6,7 @@
 
     void updateNotification(NotificationInformation notification);
 
-    void makeNotificationDismissible(NotificationInformation notificationInformation);
+    void stackNotification(NotificationInformation notificationInformation);
+
+    void dismissNotification();
```
Hutch4: 👍

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerDownloader.java
```diff
@@ -132,11 +131,17 @@ private void updateNotification(DownloadBatchStatus liteDownloadBatchStatus, Dow
...
+        }
 
-        if (liteDownloadBatchStatus.status() != DOWNLOADING) {
-            downloadService.updateNotification(notificationInformation);
+        if (liteDownloadBatchStatus.status() == DOWNLOADED) {
+            downloadService.stackNotification(notificationInformation);
+            return;
         }
+
+        downloadService.updateNotification(notificationInformation);
```
Hutch4: Can I get a clarification here. 

We are updating the service with information, but if the status was `DELETION` we would `dismiss()` before we `update()` . Is this the desired behaviour?

Mecharyry: No that's not what happens on `DELETION`. If we `DELETE` the notification just gets dismissed, there is no update because of the return.

Hutch4: Very good point 👍

PR #249, opened 2017-11-13
No downloaded status

Mecharyry:
## Problem
When a download is completed and transitions to the DOWNLOADED state in the UI the database is not updated to reflect this change 🤔 

## Solution 
Mark a `downloadBatch` as complete and update the `db`.

### DIFF
--- a/app/src/main/java/com/novoda/downloadmanager/demo/DemoApplication.java
+++ b/app/src/main/java/com/novoda/downloadmanager/demo/DemoApplication.java
@@ -5,14 +5,8 @@ import android.os.Handler;
 import android.os.Looper;
 
 import com.facebook.stetho.Stetho;
-import com.novoda.downloadmanager.DownloadsPersistence;
-import com.novoda.downloadmanager.FileDownloader;
-import com.novoda.downloadmanager.FileSizeRequester;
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/novoda/downloadmanager/demo/DemoApplication.java
```diff
@@ -26,24 +20,10 @@ public void onCreate() {
...
-        FileSizeRequester fileSizeRequester = new CustomFileSizeRequester();
-        FileDownloader fileDownloader = new CustomFileDownloader();
-        DownloadsPersistence downloadsPersistence = new CustomDownloadsPersistence();
-        NotificationCreator notificationCreator = new CustomNotificationCreator(this, R.mipmap.ic_launcher_round);
-
         Handler handler = new Handler(Looper.getMainLooper());
 
         liteDownloadManagerCommands = DownloadManagerBuilder
                 .newInstance(this, handler, R.mipmap.ic_launcher_round)
-                .withFileDownloaderCustom(fileSizeRequester, fileDownloader)
```
Mecharyry: There's a whole bunch of custom classes here to demonstrate the uses of the library but for the moment we just want to work with the vanilla version 😄

PR #251, opened 2017-11-13
Flickering on network recovery

Mecharyry:
## Problem
When there is a network error with the first download batch and network recovery is in operation then the second completed batch flickers between `DOWNLOADED` and `QUEUED`.

`DownloadError` is created and passed through the constructor of some objects as `UNKNOWN`. It seems that `UNKNOWN` is used to represent a missing error 🤔  this should instead be an `Optional` or `@Nullable` field.

## Solution
When queueing downloads ignore all items that are already `DOWNLOADED`. Made `DownloadError` `@Nullable` wherever it is currently used.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadError.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadError.java
@@ -8,7 +8,7 @@ class DownloadError {
         FILE_CANNOT_BE_WRITTEN,
         STORAGE_UNAVAILABLE,
         NETWORK_ERROR_CANNOT_DOWNLOAD_FILE,
-        UNKNOWN;
+        UNKNOWN
     }
 
     private final Error error;
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadError.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadError.java
@@ -8,16 +8,12 @@ class DownloadError {
         FILE_CANNOT_BE_WRITTEN,
         STORAGE_UNAVAILABLE,
         NETWORK_ERROR_CANNOT_DOWNLOAD_FILE,
-        UNKNOWN
+        UNKNOWN;
     }
 
-    private Error error;
+    private final Error error;
 
-    DownloadError() {
-        this.error = Error.UNKNOWN;
-    }
-
-    void setError(Error error) {
+    DownloadError(Error error) {
         this.error = error;
     }
 
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadFileStatus.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadFileStatus.java
@@ -1,5 +1,7 @@
 package com.novoda.downloadmanager;
 
+import android.support.annotation.Nullable;
+
 class DownloadFileStatus {
 
     enum Status {
@@ -13,15 +15,14 @@ class DownloadFileStatus {
 
     private final DownloadFileId downloadFileId;
 
-    private DownloadError downloadError;
     private FileSize fileSize;
     private Status status;
+    private DownloadError downloadError;
 
-    DownloadFileStatus(DownloadFileId downloadFileId, DownloadFileStatus.Status status, FileSize fileSize, DownloadError downloadError) {
+    DownloadFileStatus(DownloadFileId downloadFileId, Status status, FileSize fileSize) {
         this.downloadFileId = downloadFileId;
         this.status = status;
         this.fileSize = fileSize;
-        this.downloadError = downloadError;
     }
 
     void update(FileSize fileSize) {
@@ -57,7 +58,7 @@ class DownloadFileStatus {
     }
 
     boolean isMarkedAsError() {
-        return status == Status.ERROR;
+        return status == Status.ERROR && downloadError != null;
     }
 
     void markAsQueued() {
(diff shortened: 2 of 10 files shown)


library/src/main/java/com/novoda/downloadmanager/DownloadError.java
```diff
@@ -11,13 +11,9 @@
         UNKNOWN
     }
 
-    private Error error;
+    private final Error error;
```
Hutch4: 👍

library/src/main/java/com/novoda/downloadmanager/DownloadFileStatus.java
```diff
@@ -57,7 +58,7 @@ void isMarkedAsPaused() {
     }
 
     boolean isMarkedAsError() {
-        return status == Status.ERROR;
+        return status == Status.ERROR && downloadError != null;
```
Hutch4: What is the use case for if the Status is `ERROR` but there is no `downloadError`? 
You could reach this case but passing in  `Status.ERROR` in the constructor.

library/src/main/java/com/novoda/downloadmanager/LiteDownloadBatchStatus.java
```diff
@@ -109,6 +114,10 @@ private void updateStatus(Status status, DownloadsBatchStatusPersistence persist
 
     @Override
     public DownloadError.Error getDownloadErrorType() {
```
Hutch4: should his be annotated `@Nullable`?

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerDownloader.java
```diff
@@ -95,15 +94,18 @@ private void waitForDownloadService() {
...
         downloadService.download(downloadBatch, downloadBatchCallback());
     }
 
-    private void updateStatusToQueuedIfNeeded(DownloadBatch downloadBatch) {
-        InternalDownloadBatchStatus liteDownloadBatchStatus = downloadBatch.status();
-
-        if (liteDownloadBatchStatus.status() != PAUSED) {
-            liteDownloadBatchStatus.markAsQueued(downloadsBatchPersistence);
+    private void updateStatusToQueuedIfNeeded(InternalDownloadBatchStatus downloadBatchStatus) {
+        if (downloadBatchStatus.status() != PAUSED) {
```
Hutch4: if the status is `ERROR`  should it be Queued?

Mecharyry: Yes, it should be re-queued as it might have been a network error.

library/src/main/java/com/novoda/downloadmanager/DownloadFileStatus.java
Mecharyry: Hmmm yes in theory this could happen. The point is that `DownloadFileStatus` shouldn't be created with an `ErrorStatus` that's why the `Error` was removed from the constructor signature. Unfortunately I'm not sure how to get around this without making everything **immutable** but by doing that we will end up with memory churn associated with the update callbacks.

Hutch4: Yeah thats fair. I guess it would have to be a redesign or something sort, that is out of scope of fixing this.  
Definitely wouldn't block the merging of this PR on this 👍

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerDownloader.java
```diff
@@ -95,15 +94,18 @@ private void waitForDownloadService() {
     }
 
     private void executeDownload(final DownloadBatch downloadBatch) {
-        updateStatusToQueuedIfNeeded(downloadBatch);
+        InternalDownloadBatchStatus downloadBatchStatus = downloadBatch.status();
+        if (downloadBatchStatus.status() == DOWNLOADED) {
+            return;
```
pablisco: Mid return may worsen readability

Mecharyry: df8eb5a

library/src/main/java/com/novoda/downloadmanager/LiteDownloadBatchStatus.java
```diff
@@ -92,9 +96,10 @@ public void markForDeletion() {
     }
 
     @Override
-    public void markAsError(DownloadError downloadError) {
+    public void markAsError(DownloadError downloadError, DownloadsBatchStatusPersistence persistence) {
         this.status = Status.ERROR;
         this.downloadError = downloadError;
+        updateStatus(status, persistence);
```
zegnus: This isn't needed ;)

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerDownloader.java
```diff
@@ -95,15 +94,16 @@ private void waitForDownloadService() {
     }
 
     private void executeDownload(final DownloadBatch downloadBatch) {
-        updateStatusToQueuedIfNeeded(downloadBatch);
-        downloadService.download(downloadBatch, downloadBatchCallback());
+        InternalDownloadBatchStatus downloadBatchStatus = downloadBatch.status();
+        if (downloadBatchStatus.status() != DOWNLOADED) {
```
zegnus: maybe its clearer with a fast return?

PR #254, opened 2017-11-15
CallbackThrottleByTime test

Mecharyry:
## Problem
Wanted to test the `CallbackThrottleByTime` class but it creates the objects responsible for threading internally. 

## Solution
Create a `Scheduler` interface to hide the threading mechanism that is used from the `CallbackThrottleByTime` logic. 

Create a `SchedulerFactory` for injecting the dependencies we need to create a `FixedRateTimerScheduler` in production code and still be able to vary it for testing purposes. 

Fully test both the `CallbackThrottleByTime` and the `FixedRateTimerScheduler` class 😄

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/CallbackThrottleByTime.java
+++ b/library/src/main/java/com/novoda/downloadmanager/CallbackThrottleByTime.java
@@ -1,49 +1,40 @@
 package com.novoda.downloadmanager;
 
-import java.util.Timer;
-import java.util.TimerTask;
-
 class CallbackThrottleByTime implements CallbackThrottle {
 
-    private static final long DELAY_IN_MILLIS = 0;
+    private final Scheduler scheduler;
 
     private DownloadBatchStatus downloadBatchStatus;
-    private Timer timer;
-    private TimerTask timerTask;
-    private final long periodInMillis;
     private DownloadBatchCallback callback;
 
-    CallbackThrottleByTime(long periodInMillis) {
-        this.periodInMillis = periodInMillis;
+    CallbackThrottleByTime(Scheduler scheduler) {
+        this.scheduler = scheduler;
     }
 
     @Override
     public void setCallback(final DownloadBatchCallback callback) {
         this.callback = callback;
-        timerTask = new TimerTask() {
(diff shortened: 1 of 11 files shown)


library/src/main/java/com/novoda/downloadmanager/CallbackThrottleByTime.java
```diff
@@ -1,60 +1,47 @@
...
+    private final Scheduler scheduler;
 
     private DownloadBatchStatus downloadBatchStatus;
-    private Timer timer;
-    private TimerTask timerTask;
     private DownloadBatchCallback callback;
 
-    CallbackThrottleByTime(long period) {
-        this.period = period;
+    CallbackThrottleByTime(Scheduler scheduler) {
```
Hutch4: 👍

library/src/main/java/com/novoda/downloadmanager/Scheduler.java
```diff
@@ -0,0 +1,17 @@
+package com.novoda.downloadmanager;
+
+interface Scheduler {
```
Hutch4: Is Rx used in downloadmanager at all?

I'd avoid the use of Scheduler as an interface name, as it's very common. 
`DownloadActionScheduler` or `ActionScheduler` might be better?

library/src/main/java/com/novoda/downloadmanager/CallbackThrottleCreator.java
```diff
@@ -41,7 +41,7 @@ private CallbackThrottleCreator(Type type, TimeUnit timeUnit, long frequency, Cl
     CallbackThrottle create() {
         switch (type) {
             case THROTTLE_BY_TIME:
-                return new CallbackThrottleByTime(timeUnit.toMillis(frequency));
+                return new CallbackThrottleByTime(SchedulerFactory.createFixedRateTimerScheduler(timeUnit.toMillis(frequency)));
```
Hutch4: i'd make extract out the Scheduler creation into a local var. To make it clear about what is being created.

Mecharyry:
@Hutch4 I've addressed your comments.

PR #255, opened 2017-11-15
Add fixtures

Mecharyry:
## Problem
We want to start testing `DownloadBatch` but it has a number of dependencies, some of which are models. It would be good to have fixtures on these in place before starting testing.

## Solution
Add some fixtures to create default models that can be used in tests.


Mecharyry:
We should take a look at the coupling on `DownloadBatch` and `DownloadFile`, these classes are difficult to test.

PR #258, opened 2017-11-16
Add tests for the DownloadManager

Mecharyry:
Add tests to the `DownloadManager` class. Opening for a head-start on reviews 😄  there will be some tidy up on these tests.  Does anyone have ideas on how to test the synchronized blocks in this class?

### DIFF
--- a/library/src/test/java/com/novoda/downloadmanager/DownloadManagerTest.java
+++ b/library/src/test/java/com/novoda/downloadmanager/DownloadManagerTest.java
@@ -18,15 +18,14 @@ import org.mockito.stubbing.Answer;
 import static com.google.common.truth.Truth.assertThat;
 import static com.novoda.downloadmanager.DownloadBatchIdFixtures.aDownloadBatchId;
 import static com.novoda.downloadmanager.InternalDownloadBatchStatusFixtures.anInternalDownloadsBatchStatus;
-import static org.mockito.BDDMockito.*;
-import static org.mockito.Mockito.any;
-import static org.mockito.Mockito.inOrder;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.verifyZeroInteractions;
+import static org.mockito.BDDMockito.given;
+import static org.mockito.BDDMockito.willAnswer;
+import static org.mockito.Mockito.*;
 
 public class DownloadManagerTest {
 
+    private static final InternalDownloadBatchStatus BATCH_STATUS = anInternalDownloadsBatchStatus().build();
+    private static final InternalDownloadBatchStatus ADDITIONAL_BATCH_STATUS = anInternalDownloadsBatchStatus().build();
     private static final DownloadBatchId DOWNLOAD_BATCH_ID = aDownloadBatchId().withRawDownloadBatchId("id01").build();
     private static final DownloadBatchId ADDITIONAL_DOWNLOAD_BATCH_ID = aDownloadBatchId().withRawDownloadBatchId("id02").build();
     private static final Batch BATCH = new Batch.Builder(DOWNLOAD_BATCH_ID, "title").build();
@@ -42,10 +41,11 @@ public class DownloadManagerTest {
     private final FileOperations fileOperations = mock(FileOperations.class);
     private final DownloadsBatchPersistence downloadsBatchPersistence = mock(DownloadsBatchPersistence.class);
     private final LiteDownloadManagerDownloader downloadManagerDownloader = mock(LiteDownloadManagerDownloader.class);
-    private List<DownloadBatchCallback> downloadBatchCallbacks = new ArrayList<>();
 
     private DownloadManager downloadManager;
-    private Map<DownloadBatchId, DownloadBatch> downloadBatches;
+    private Map<DownloadBatchId, DownloadBatch> downloadBatches = new HashMap<>();
+    private List<DownloadBatchStatus> downloadBatchStatuses = new ArrayList<>();
+    private List<DownloadBatchCallback> downloadBatchCallbacks = new ArrayList<>();
 
     @Before
     public void setUp() {
@@ -73,6 +73,18 @@ public class DownloadManagerTest {
                 return null;
             }
         }).given(downloadsBatchPersistence).loadAsync(any(FileOperations.class), loadBatchesCallbackCaptor.capture());
+
+        final ArgumentCaptor<List> argumentCaptor = ArgumentCaptor.forClass(List.class);
+        willAnswer(new Answer<Void>() {
+            @Override
+            public Void answer(InvocationOnMock invocation) throws Throwable {
+                downloadBatchStatuses = argumentCaptor.getValue();
+                return null;
+            }
+        }).given(allBatchStatusesCallback).onReceived(argumentCaptor.capture());
+
+        given(downloadBatch.status()).willReturn(BATCH_STATUS);
+        given(additionalDownloadBatch.status()).willReturn(ADDITIONAL_BATCH_STATUS);
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


library/src/test/java/com/novoda/downloadmanager/DownloadManagerTest.java
```diff
@@ -0,0 +1,239 @@
...
+                executorService,
+                downloadBatches,
+                downloadBatchCallbacks,
+                fileOperations,
+                downloadsBatchPersistence,
+                downloadManagerDownloader
+        );
+
+        final ArgumentCaptor<DownloadsBatchPersistence.LoadBatchesCallback> loadBatchesCallbackCaptor = ArgumentCaptor.forClass(DownloadsBatchPersistence.LoadBatchesCallback.class);
+        willAnswer(new Answer<Void>() {
```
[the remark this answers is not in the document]
pablisco: And do it when needed rather than on set up

library/src/test/java/com/novoda/downloadmanager/DownloadManagerTest.java
```diff
@@ -0,0 +1,239 @@
...
+    private DownloadManager downloadManager;
+    private Map<DownloadBatchId, DownloadBatch> downloadBatches = new HashMap<>();
+    private List<DownloadBatchStatus> downloadBatchStatuses = new ArrayList<>();
+    private List<DownloadBatchCallback> downloadBatchCallbacks = new ArrayList<>();
+
+    @Before
+    public void setUp() {
+        downloadBatches = new HashMap<>();
+        downloadBatches.put(DOWNLOAD_BATCH_ID, downloadBatch);
+        downloadBatches.put(ADDITIONAL_DOWNLOAD_BATCH_ID, additionalDownloadBatch);
```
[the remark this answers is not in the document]
pablisco: Also `downloadBatch` is also changed later one in a capture

library/src/test/java/com/novoda/downloadmanager/DownloadManagerTest.java
```diff
@@ -0,0 +1,239 @@
...
+        downloadManager.getAllDownloadBatchStatuses(allBatchStatusesCallback);
+    }
+
+    @Test
+    public void getsAllBatchStatuses_whenServiceAlreadyExists() {
+        downloadManager.initialise(mock(DownloadService.class));
+
+        downloadManager.getAllDownloadBatchStatuses(allBatchStatusesCallback);
+
+        assertThat(downloadBatchStatuses).containsAllOf(BATCH_STATUS, ADDITIONAL_BATCH_STATUS);
```
pablisco: Better to use `containsOnly`, it's more specific

pablisco: `containsExactly` sorry, thought it was `assertJ`

library/src/test/java/com/novoda/downloadmanager/DownloadManagerTest.java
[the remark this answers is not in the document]
Mecharyry: @Hutch4 what's the benefit of moving this? 

@pablisco this is the default implementation so most of the tests don't break and then we override in the test that changes it.

library/src/test/java/com/novoda/downloadmanager/DownloadManagerTest.java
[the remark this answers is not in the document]
Mecharyry: I don't see the benefit of moving this @pablisco this is essentially setting up the mock behaviour that will be used in the tests. I can move to the test but then it will make that test much more difficult to read.

library/src/test/java/com/novoda/downloadmanager/DownloadManagerTest.java
```diff
@@ -0,0 +1,246 @@
...
+        verify(downloadBatch).delete();
+    }
+
+    @Test
+    public void addsCallbackToInternalList() {
+        DownloadBatchCallback additionalDownloadBatchCallback = mock(DownloadBatchCallback.class);
+
+        downloadManager.addDownloadBatchCallback(additionalDownloadBatchCallback);
+
+        assertThat(downloadBatchCallbacks).contains(additionalDownloadBatchCallback);
```
pablisco: We could test this by doing an action on `downloadManager` and verify the response is received rather than expose the internals of the class

library/src/test/java/com/novoda/downloadmanager/DownloadManagerTest.java
```diff
@@ -0,0 +1,246 @@
...
+        verify(lock).notifyAll();
+    }
+
+    @Test
+    public void triggersDownloadOfBatches_whenSubmittingAllStoredDownloads() {
+        downloadManager.submitAllStoredDownloads(allStoredDownloadsSubmittedCallback);
+
+        InOrder inOrder = inOrder(downloadManagerDownloader);
+        inOrder.verify(downloadManagerDownloader).download(downloadBatch, downloadBatches);
+        inOrder.verify(downloadManagerDownloader).download(additionalDownloadBatch, downloadBatches);
```
pablisco: We can add `inOrder.verifyNoMoreInteractions();` here

library/src/test/java/com/novoda/downloadmanager/DownloadManagerTest.java
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    @Test
+    public void setDownloadService_whenInitialising() {
+        downloadManager.initialise(downloadService);
+
+        verify(downloadManagerDownloader).setDownloadService(downloadService);
+    }
+
+    @Ignore // How do we unit test the synchronized block?
```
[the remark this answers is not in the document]
Mecharyry: I managed to get it working after looking a little bit more on how synchronized blocks work 😄 

```
        downloadManager.initialise(downloadService);

        synchronized (lock) {
            verify(lock).notifyAll();
        }
```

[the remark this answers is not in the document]
Hutch4: Nice! Good to know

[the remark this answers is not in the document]
Mecharyry: Scrap that mockito has a silent failure, throws warning but seems to pass. Can do it like this instead but a bit more verbose.

```
        final AtomicBoolean triggeredNotify = new AtomicBoolean(false);

        synchronized (lock) {
            EXECUTOR.submit(new Runnable() {
                @Override
                public void run() {
                    downloadManager.initialise(downloadService);
                    triggeredNotify.set(true);
                }
            });

            lock.wait();
        }

        assertThat(triggeredNotify.get()).isTrue();
```

library/src/test/java/com/novoda/downloadmanager/DownloadManagerTest.java
```diff
@@ -0,0 +1,281 @@
...
+
+    @Test
+    public void setDownloadService_whenInitialising() {
+        downloadManager.initialise(downloadService);
+
+        verify(downloadManagerDownloader).setDownloadService(downloadService);
+    }
+
+    @Test(timeout = 500)
+    public void notifyAll_whenInitialising() throws InterruptedException {
```
Mecharyry: Can use `timeout` because `synchronized` with `wait` and `notify` will block and unblock the testing thread. 

We force the testing thread to lock-up with `lock.wait` and at the same time schedule a call that will internally release the thread with `downloadManager.initialise`.

library/src/test/java/com/novoda/downloadmanager/DownloadManagerTest.java
```diff
@@ -0,0 +1,281 @@
...
+
+    @Test
+    public void removesCallbackFromInternalList() {
+        downloadManager.removeDownloadBatchCallback(downloadBatchCallback);
+
+        assertThat(downloadBatchCallbacks).doesNotContain(downloadBatchCallback);
+    }
+
+    @Test(timeout = 500)
+    public void waitsForServiceToExist_whenGettingAllBatchStatuses() throws InterruptedException {
```
Mecharyry: This one was painful. Need to argument captor the runnable to ensure that it runs on the same thread as the test, so it locks up when the internal `wait` is called. We then need to trigger a notify on another thread which will release the testing thread so that it doesn't hit the timeout and fail.

library/src/test/java/com/novoda/downloadmanager/DownloadManagerTest.java
```diff
@@ -0,0 +1,281 @@
...
+        final ArgumentCaptor<Runnable> argumentCaptor = ArgumentCaptor.forClass(Runnable.class);
+        willAnswer(new Answer<Void>() {
+            @Override
+            public Void answer(InvocationOnMock invocation) throws Throwable {
+                argumentCaptor.getValue().run();
+                return null;
+            }
+        }).given(executorService).submit(argumentCaptor.capture());
```
pablisco: do we need a captor? The parameter is passed into he answer

Mecharyry: I never realised this before ❤️

PR #259, opened 2017-11-16
Add tests for CallbackThrottleCreator

Mecharyry:
Just adding tests for the `CallbackThrottleCreator` 😄

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/CallbackThrottleCreator.java
+++ b/library/src/main/java/com/novoda/downloadmanager/CallbackThrottleCreator.java
@@ -47,7 +47,7 @@ class CallbackThrottleCreator {
             case CUSTOM:
                 return createCallbackThrottle();
             default:
-                throw new IllegalStateException("type " + type + " not supported");
+                throw new IllegalStateException("type " + type + " not supported.");
         }
     }
 
--- a/library/src/test/java/com/novoda/downloadmanager/CallbackThrottleCreatorTest.java
+++ b/library/src/test/java/com/novoda/downloadmanager/CallbackThrottleCreatorTest.java
@@ -1,0 +1,80 @@
+package com.novoda.downloadmanager;
+
+import java.util.concurrent.TimeUnit;
+
+import org.junit.Test;
+
+import static com.google.common.truth.Truth.assertThat;
+
+public class CallbackThrottleCreatorTest {
+
+    @Test
+    public void createsTimeThrottle() {
+        CallbackThrottle callbackThrottle = CallbackThrottleCreator.ByTime(TimeUnit.SECONDS, 1)
+                .create();
+
+        assertThat(callbackThrottle).isInstanceOf(CallbackThrottleByTime.class);
+    }
+
+    @Test
+    public void createsProgressThrottle() {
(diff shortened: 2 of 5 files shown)


library/src/main/java/com/novoda/downloadmanager/CallbackThrottleCreator.java
```diff
@@ -47,7 +47,7 @@ CallbackThrottle create() {
             case CUSTOM:
                 return createCallbackThrottle();
             default:
-                throw new IllegalStateException("type " + type + " not supported");
+                throw new IllegalStateException("type " + type + " not supported.");
```
jonreeve: Grammar! 😄 👍

library/src/test/java/com/novoda/downloadmanager/CallbackThrottleCreatorTest.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.concurrent.TimeUnit;
+
+import org.junit.Test;
+
+import static com.google.common.truth.Truth.assertThat;
+
+public class CallbackThrottleCreatorTest {
+
+    @Test
+    public void createsTimeThrottle() {
```
jonreeve: ✨

library/src/test/java/com/novoda/downloadmanager/CallbackThrottleCreatorTest.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void throwsException_whenCustomCallbackIsNotFound() {
+        CallbackThrottleCreator.ByCustomThrottle(TestNotFoundCustomThrottle.class)
+                .create();
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void throwsException_whenCustomCallbackIsNotInstantiable() {
```
jonreeve: Are these things we have code to specifically handle? They sound a bit like language features, as in not our code. If we have cases specifically for them then OK.

Mecharyry: We have cases for these in the code. I still don't fully understand the need to pass a class type and instantiate in the `CallbackThrottleCreator` but it's there with all the associated catches 😄

jonreeve: yeah it does seem odd but :man_shrugging: 

Cool, so long as it's testing our stuff 👍

library/src/test/java/com/novoda/downloadmanager/CallbackThrottleCreatorTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+    @Test(expected = RuntimeException.class)
+    public void throwsException_whenCallbackThrottleDoesNotExist() {
+        CallbackThrottleCreator.ByCustomThrottle(null)
+                .create();
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void throwsException_whenCustomCallbackIsNotPublic() {
+        CallbackThrottleCreator.ByCustomThrottle(TestNonPublicCustomThrottle.class)
```
zegnus: 💯

PR #260, opened 2017-11-16
Wrap OkHttpClient classes

Mecharyry:
## Problem
Wanted to test `NetworkFileDownloader` but it uses `OkHttp` which has a few final classes that can't be mocked / extended to ensure everything works as expected 😢 

## Solution
Create an `HttpClient` interface and wrap `OkHttp` classes so that we can mock / extend and swap these out if needed without breaking the logic that is in `NetworkFileDownloader`. Would rather not wrap these classes so if someone can think of a better solution that would be awesome 😄 

No tests in this PR, will be in a follow-up.


library/src/main/java/com/novoda/downloadmanager/DownloadManagerRequestCreator.java
```diff
@@ -0,0 +1,26 @@
+package com.novoda.downloadmanager;
+
+import java.util.HashMap;
+import java.util.Map;
+
+class DownloadManagerRequestCreator {
```
Mecharyry: This class now holds of the requests that the app does 🤔  at least I'm pretty sure it does, unless I missed any 😆

library/src/main/java/com/novoda/downloadmanager/HttpClient.java
```diff
@@ -0,0 +1,23 @@
+package com.novoda.downloadmanager;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+interface HttpClient {
```
Mecharyry: Could have interfaced the `DownloadManagerRequest` too but it seemed overkill and unnecessary.

library/src/main/java/com/novoda/downloadmanager/NetworkFileDownloader.java
```diff
@@ -14,30 +10,27 @@
 
     private static final int BUFFER_SIZE = 8 * 512;
 
-    private final OkHttpClient httpClient;
+    private final HttpClient httpClient;
```
Mecharyry: Not tied to `OkHttp` anymore 🎉

library/src/main/java/com/novoda/downloadmanager/WrappedOkHttpClient.java
```diff
@@ -0,0 +1,35 @@
...
+class WrappedOkHttpClient implements HttpClient {
+
+    private final OkHttpClient httpClient;
+
+    WrappedOkHttpClient(OkHttpClient httpClient) {
+        this.httpClient = httpClient;
+    }
+
+    @Override
+    public DownloadManagerResponse execute(DownloadManagerRequest request) throws IOException {
```
Mecharyry: This should make testing this class and the `FileDownloader` much easier 😄  `throws` because we still handles these in the calling code.

library/src/main/java/com/novoda/downloadmanager/HttpClient.java
```diff
@@ -0,0 +1,23 @@
...
+
+        boolean isSuccessful();
+
+        String header(String name, String defaultValue);
+
+        InputStream openByteStream() throws IOException;
+
+        void closeByteStream() throws IOException;
+    }
+
```
ataulm: mindful of the project convention re: blank lines - I see a mix in this PR

Mecharyry: I don't know what the convention is in this project 😂

library/src/main/java/com/novoda/downloadmanager/DownloadManagerRequestCreator.java
zegnus: What about renaming this to `NetworkRequestCreator`, as this object only leaves in the specific implementation `NetworkFileDownloader` of the interface `FileDownloader`

library/src/main/java/com/novoda/downloadmanager/DownloadManagerRequest.java
```diff
@@ -0,0 +1,72 @@
+package com.novoda.downloadmanager;
+
+import java.util.Map;
+
+class DownloadManagerRequest {
```
zegnus: Could be rename this to `NetworkRequest`?

library/src/main/java/com/novoda/downloadmanager/DownloadManagerRequestCreator.java
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    DownloadManagerRequest createDownloadRequestWithDownloadedBytesHeader(String url, long currentSize, long totalSize) {
+        Map<String, String> headers = new HashMap<>();
+        String headerValue = String.format(DOWNLOADED_BYTES_VALUE_FORMAT, currentSize, totalSize);
+        headers.put("Range", headerValue);
+
+        return new DownloadManagerRequest(headers, url, DownloadManagerRequest.Method.GET);
+    }
+
```
zegnus: in here you didn't https://github.com/novoda/download-manager/pull/260/files#diff-27dcb7c35d12550580704edd131f729dR39 :)

Mecharyry:
@zegnus renamed all of the network classes to include `Network` and drop `DownloadManager`.

PR #261, opened 2017-11-17
File downloader tests

Mecharyry:
Add tests for the `NetworkFileDownloader`

### DIFF
--- a/library/src/test/java/com/novoda/downloadmanager/NetworkFileDownloaderTest.java
+++ b/library/src/test/java/com/novoda/downloadmanager/NetworkFileDownloaderTest.java
@@ -11,50 +11,38 @@ import org.junit.Test;
 
 import static com.novoda.downloadmanager.InternalFileSizeFixtures.aFileSize;
 import static com.novoda.downloadmanager.NetworkResponseFixtures.aNetworkResponse;
-import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.BDDMockito.given;
 import static org.mockito.Mockito.*;
 
 public class NetworkFileDownloaderTest {
 
     private static final byte[] BYTES_TO_RECEIVE = "s".getBytes();
+    private static final HttpClient.NetworkResponse RESPONSE_WITH_INPUT_STREAM = aNetworkResponse()
+            .withInputStream(new ByteArrayInputStream(BYTES_TO_RECEIVE))
+            .build();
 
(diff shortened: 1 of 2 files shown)


library/src/test/java/com/novoda/downloadmanager/NetworkFileDownloaderTest.java
```diff
@@ -0,0 +1,120 @@
...
+        given(httpClient.execute(any(NetworkRequest.class))).willReturn(
+                aNetworkResponse().withInputStream(new ByteArrayInputStream(BYTES_TO_RECEIVE)).build()
+        );
+    }
+
+    @Test
+    public void createsRequestWithDownloadedBytesHeader_whenKnown() {
+        networkFileDownloader.startDownloading(ANY_RAW_URL, KNOWN_FILE_SIZE, callback);
+
+        verify(requestCreator).createDownloadRequestWithDownloadedBytesHeader(ANY_RAW_URL, KNOWN_FILE_SIZE.currentSize(), KNOWN_FILE_SIZE.totalSize());
```
zegnus: I think that this is testing an interaction with an internal collaborator. If we refactor this class and the way that the internal collaborators work we will then have to come here and change this again. We should only test what the public methods/callbacks return and interaction with external collaborators. What do you think? The idea is to allow internal refactoring without having to change tests, while keeping behaviour tested

Mecharyry: 👍 very true. Updated in 2703344

library/src/test/java/com/novoda/downloadmanager/NetworkFileDownloaderTest.java
```diff
@@ -0,0 +1,108 @@
...
+        networkFileDownloader.startDownloading(ANY_RAW_URL, KNOWN_FILE_SIZE, callback);
+
+        byte[] expectedBytes = new byte[4096];
+        expectedBytes[0] = "s".getBytes()[0];
+
+        verify(callback).onBytesRead(expectedBytes, 1);
+    }
+
+    @Test
+    public void closesByteStream_whenDownloading() throws IOException {
```
ataulm: should this be `closesByteStream_afterDownloaded` or something? beware the test names that describe the implementation over describing the test case

Mecharyry: a02ab82

PR #263, opened 2017-12-04
V1 to V2 migration

Mecharyry:
### Problem
In version two of DM we have a completely different DB structure using a different underlying implementation with Room. We need to be able to migrate existing downloads from v1 to v2 with no visible impact to client applications.

### Solution
Quite a bit of this PR is porting the simple demo from v1 to v2. Classes of note are:
- `MigrationFactory`
- Migrator specifically `VersionOneToVersionTwoMigrator`
- `MigrationExtractor` and the associated test class

Retrieve the `Batch Id` & `Batch Title` which can be used to extract the `originating network address` and the `local storage address`. Using this we create new file locations in the v2 db and then register new batches against the new DM. Once the batches have been transferred we nuke the old database so that subsequent runs do not trigger the migration again.

### Note
This has only be done on the happy path assuming a batch has been completely downloaded.

### Screen Captures
Shows one complete batch and an incomplete batch in v1. When porting over to v2 only the completed batches are migrated.
##### V1 DB
screen shot 2017-12-06 at 15 14 13
screen shot 2017-12-06 at 15 14 22

##### V2 DB
screen shot 2017-12-06 at 15 15 00
screen shot 2017-12-06 at 15 15 08


library/src/test/java/com/novoda/downloadmanager/StubCursor.java
```diff
@@ -0,0 +1,292 @@
...
+import android.os.Bundle;
+import android.util.Log;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+class StubCursor implements Cursor {
```
Mecharyry: Plan to swap this out with a wrapped `Cursor` that we own and is easier to mock / rely on.

Mecharyry:
Originally had an additional demo on version 2 to ensure that we had both the old and new database. Given that DBs are per application I've renamed the demo and ensured that the `application id` matched from the old sample, this ensures that the v1 DB exists when running v2 of the demo so we can check the migration. This means no additional demo needed 🎉

library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
```diff
@@ -0,0 +1,59 @@
...
+        this.database = database;
+    }
+
+    List<Migration> extractMigrations() {
+        Cursor batchesCursor = database.rawQuery(BATCHES_QUERY);
+
+        List<Migration> migrations = new ArrayList<>();
+        while (batchesCursor.moveToNext()) {
+
+            String query = "SELECT uri, _data, total_bytes FROM Downloads WHERE batch_id = ?";
```
zegnus: This can probably be a constant

library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
```diff
@@ -0,0 +1,59 @@
...
+            String batchId = batchesCursor.getString(0);
```
zegnus: We could extract the `0`, `1` as a constants

library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
```diff
@@ -0,0 +1,59 @@
...
+            String batchTitle = batchesCursor.getString(1);
+
+            Cursor uriCursor = database.rawQuery(query, batchId);
+            Batch.Builder newBatchBuilder = new Batch.Builder(DownloadBatchIdCreator.createFrom(batchId), batchTitle);
+
+            List<String> originalFileLocations = new ArrayList<>();
+            List<FileSize> fileSizes = new ArrayList<>();
+
+            while (uriCursor.moveToNext()) {
+                String uri = uriCursor.getString(0);
```
zegnus: same here

library/src/main/java/com/novoda/downloadmanager/MigrationFactory.java
```diff
@@ -0,0 +1,34 @@
...
+
+public final class MigrationFactory {
+
+    private MigrationFactory() {
+        // Uses static methods.
+    }
+
+    public static Migrator createVersionOneToVersionTwoMigrator(Context context, File databasePath, Migrator.Callback callback) {
+        if (!databasePath.exists()) {
+            return new Migrator() {
```
zegnus: We could have a static NO_OP migrator defined in the interface

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
```diff
@@ -0,0 +1,137 @@
...
+    private void migrateV1FilesToV2Location(List<Migration> migrations) {
+        for (Migration migration : migrations) {
+            migrateV1FilesToV2Location(migration);
+        }
+    }
+
+    private void migrateV1FilesToV2Location(Migration migration) {
+        Batch batch = migration.batch();
+
+        for (int i = 0; i < migration.originalFileLocations().size(); i++) {
```
zegnus: You can `migration.originalFileLocations().size()` once

library/src/main/java/com/novoda/downloadmanager/Migrator.java
```diff
@@ -0,0 +1,9 @@
+package com.novoda.downloadmanager;
+
+public interface Migrator {
+    void migrate();
```
zegnus: new line before

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
```diff
@@ -0,0 +1,137 @@
...
+            // initialise the InternalFilePersistence
+            String originalFileLocation = migration.originalFileLocations().get(i);
+            FileSize actualFileSize = migration.fileSizes().get(i);
+            FileName newFileName = LiteFileName.from(batch, batch.getFileUrls().get(i));
+            internalFilePersistence.create(newFileName, actualFileSize);
+
+            FileInputStream inputStream = null;
```
zegnus: From API 26 you could use `Files.move(source, destination, StandardCopyOption.ATOMIC_MOVE);`

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
```diff
@@ -0,0 +1,137 @@
...
+
+    private void migrateV1DataToV2Database(Migration migration) {
+        Batch batch = migration.batch();
+        downloadsPersistence.startTransaction();
+
+        DownloadBatchTitle downloadBatchTitle = new LiteDownloadBatchTitle(batch.getTitle());
+        DownloadsBatchPersisted persistedBatch = new LiteDownloadsBatchPersisted(downloadBatchTitle, batch.getDownloadBatchId(), DownloadBatchStatus.Status.DOWNLOADED);
+        downloadsPersistence.persistBatch(persistedBatch);
+
+        for (int i = 0; i < migration.originalFileLocations().size(); i++) {
```
zegnus: You can `migration.originalFileLocations().size()` once

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
```diff
@@ -0,0 +1,137 @@
...
+            String url = batch.getFileUrls().get(i);
+            FileName fileName = LiteFileName.from(batch, url);
+            FilePath filePath = FilePathCreator.create(fileName.name());
+            DownloadFileId downloadFileId = DownloadFileId.from(batch);
+            DownloadsFilePersisted persistedFile = new LiteDownloadsFilePersisted(
+                    batch.getDownloadBatchId(),
+                    downloadFileId,
+                    fileName,
+                    filePath,
+                    migration.fileSizes().get(i).totalSize(),
```
zegnus: You coud take `migration.fileSizes()` out of the loop

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
```diff
@@ -0,0 +1,137 @@
...
+            downloadsPersistence.persistFile(persistedFile);
+        }
+
+        downloadsPersistence.transactionSuccess();
+        downloadsPersistence.endTransaction();
+    }
+
+    private void deleteFrom(DatabaseWrapper database, List<Migration> migrations) {
+        for (Migration migration : migrations) {
+            String query = "DELETE FROM batches WHERE _id = ?";
```
zegnus: This could be a constant

library/src/main/java/com/novoda/downloadmanager/DatabaseWrapper.java
```diff
@@ -0,0 +1,33 @@
+package com.novoda.downloadmanager;
+
+import android.database.Cursor;
+import android.database.sqlite.SQLiteDatabase;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+class DatabaseWrapper {
```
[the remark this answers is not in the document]
Mecharyry: 69d58e0027da737ab672b26afa9fa359f2e43e2c

library/src/main/java/com/novoda/downloadmanager/Migration.java
```diff
@@ -0,0 +1,74 @@
+package com.novoda.downloadmanager;
+
+import java.util.List;
+
+class Migration {
+
+    private Batch batch;
+    private final List<String> originalFileLocations;
```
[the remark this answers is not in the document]
Mecharyry: 4926db94dffc0666a0349c2eb01f34222b7519a5

library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
Mecharyry: 12bcc53c09b05c0d397085dfc23fddc870faca03

library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
Mecharyry: 93ef127e02c9bc9e18347e9669fb515a92fc5bc2

library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
Mecharyry: 93ef127e02c9bc9e18347e9669fb515a92fc5bc2

library/src/main/java/com/novoda/downloadmanager/MigrationFactory.java
Mecharyry: 2639fbe2a260b7f4511fd44f38d735946c569eff

library/src/main/java/com/novoda/downloadmanager/Migrator.java
Mecharyry: 2639fbe2a260b7f4511fd44f38d735946c569eff

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
Mecharyry: ace556cc2d0b4d63f6356b21eb82459afdaa5d38

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
Mecharyry: I'd rather avoid for now, happy to create a tech-debt for this.

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
Mecharyry: de203b7eae396924aeb6a703a02afcab55862c47

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
Mecharyry: de203b7eae396924aeb6a703a02afcab55862c47

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
Mecharyry: 5d0a8bb8a993077824d74d0ac47c851463a43559

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
zegnus: 👍

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -39,6 +41,25 @@ protected void onCreate(Bundle savedInstanceState) {
...
+                            @Override
+                            public void onMigrationComplete() {
+                                deleteDatabase("downloads.db");
+                            }
+                        });
+                        migrator.migrate();
+                    }
+                }).start();
+            }
+        });
```
jonreeve: I guess we're not too fussed about this code since it's just here to exercise the lib?

library/src/main/java/com/novoda/downloadmanager/LiteFileName.java
```diff
@@ -4,12 +4,12 @@
...
     private final String name;
 
-    public static LiteFileName from(Batch batch, String fileUrl) {
+    static LiteFileName from(Batch batch, String fileUrl) {
         String name = batch + fileUrl + String.valueOf(System.nanoTime());
         return new LiteFileName(String.valueOf(name.hashCode()));
     }
 
-    public static LiteFileName from(String name) {
+    static LiteFileName from(String name) {
```
jonreeve: LiteFileName? Reference to the package name that was changed above, "litedownloadmanager" ?

library/src/main/java/com/novoda/downloadmanager/Migration.java
```diff
@@ -0,0 +1,70 @@
+package com.novoda.downloadmanager;
+
+import java.util.List;
+
+class Migration {
+
+    private final Batch batch;
+    private final List<String> originalFileLocations;
+    private final List<FileSize> fileSizes;
```
jonreeve: These two lists assumed to be same length? i.e. should there be a type to contain a file location and size?

library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.List;
+
+class MigrationExtractor {
+
+    private static final String BATCHES_QUERY = "SELECT batches._id, batches.batch_title FROM batches INNER JOIN DownloadsByBatch ON " +
+            "DownloadsByBatch.batch_id = batches._id WHERE DownloadsByBatch.batch_total_bytes = DownloadsByBatch.batch_current_bytes";
+    private static final String URIS_QUERY = "SELECT uri, _data, total_bytes FROM Downloads WHERE batch_id = ?";
+    private static final int FIRST_COLUMN = 0;
+    private static final int SECOND_COLUMN = 1;
+    private static final int THIRD_COLUMN = 2;
```
jonreeve: FIRST_COLUMN isn't really any more meaningful to me than 0. Maybe like COLUMN_INDEX_URI, or URI_COLUMN, or something (depending on your pref for grouping hierarchical names)? i.e. mention column name

library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
```diff
@@ -0,0 +1,63 @@
...
+                newBatchBuilder.addFile(uri);
+
+                originalFileLocations.add(originalFileName);
+
+                long rawFileSize = uriCursor.getLong(THIRD_COLUMN);
+                FileSize fileSize = new LiteFileSize(rawFileSize, rawFileSize);
+                fileSizes.add(fileSize);
+            }
+
+            uriCursor.close();
```
jonreeve: Should we have these cursor close calls in a finally block?

library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
```diff
@@ -0,0 +1,63 @@
...
+
+            Batch batch = newBatchBuilder.build();
+            migrations.add(new Migration(batch, originalFileLocations, fileSizes));
+        }
+        batchesCursor.close();
```
jonreeve: (I mean this one too ^)

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
joetimmins: Precisely 👍

library/src/main/java/com/novoda/downloadmanager/MigrationFactory.java
```diff
@@ -0,0 +1,29 @@
...
+        // Uses static methods.
+    }
+
+    public static Migrator createVersionOneToVersionTwoMigrator(Context context, File databasePath, Migrator.Callback callback) {
+        if (!databasePath.exists()) {
+            return Migrator.NO_OP;
+        }
+
+        SQLiteDatabase sqLiteDatabase = SQLiteDatabase.openDatabase(databasePath.getAbsolutePath(), null, 0);
+        SqlDatabaseWrapper database = new SqlDatabaseWrapper(sqLiteDatabase);
```
jonreeve: Seems a bit off having logic and DB opening in a factory class. Functionality creeping into a factory

library/src/main/java/com/novoda/downloadmanager/Migrator.java
```diff
@@ -0,0 +1,17 @@
...
+
+public interface Migrator {
+
+    void migrate();
+
+    interface Callback {
+        void onMigrationComplete();
+    }
+
+    Migrator NO_OP = new Migrator() {
```
jonreeve: 👍

library/src/main/java/com/novoda/downloadmanager/Migration.java
joetimmins: Yeah there should be. Then a collection of that type. Think we had this as a `Map` originally which isn't wonderful either. This would also clean up the use of the getters in the `Migrator` implementation as well.

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
```diff
@@ -0,0 +1,143 @@
...
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.util.List;
+
+import static com.novoda.downloadmanager.DownloadBatchStatus.Status;
+
+class VersionOneToVersionTwoMigrator implements Migrator {
+
+    private static final int BUFFER_SIZE = 8 * 512;
```
jonreeve: Maybe we could explain the size choice if there's some reasoning. Multiples of 512... disk blocks?

If we can express that in a short enough name that would be great, if not a comment would be better than nothing I think.

joetimmins: There's no real reason we're going with this buffer size other than "we tried a few a different sizes and this one seemed to work". Any suggestions welcome here!

library/src/main/java/com/novoda/downloadmanager/MigrationFactory.java
joetimmins: I don't think this is logic per se, we just have `SqlDatabaseWrapper` covering the Android class and exposing a few methods on it, for test purposes. (Don't mock types you don't own, etc.) We could perhaps move the `openDatabase` call to the wrapper though? would that be better?

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
```diff
@@ -0,0 +1,143 @@
...
+            FileName newFileName = LiteFileName.from(batch, batch.getFileUrls().get(i));
+            internalFilePersistence.create(newFileName, actualFileSize);
+
+            FileInputStream inputStream = null;
+            try {
+                // open the v1 file
+                inputStream = new FileInputStream(new File(originalFileLocation));
+                byte[] bytes = new byte[BUFFER_SIZE];
+
+                // read the v1 file
```
jonreeve: I wonder if we could extract a general file copier from here? Might reduce length + nesting, tighten scope. The comments seem to back up that something should be extracted.

I do notice that it writes to that internalFilePersistence thing rather than an output stream, which would make that not so simple. Though that might be a prompt to make the API of that more in keeping with typical Java File and Stream IO, return a stream when creating a file or something.

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
```diff
@@ -0,0 +1,143 @@
...
+                    filePath,
+                    fileSizes.get(i).totalSize(),
+                    url,
+                    FilePersistenceType.INTERNAL
+            );
+            downloadsPersistence.persistFile(persistedFile);
+        }
+
+        downloadsPersistence.transactionSuccess();
+        downloadsPersistence.endTransaction();
```
jonreeve: I guess the catch and abort scenario is just not added yet since you mentioned happy case in the description

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    private void deleteFrom(SqlDatabaseWrapper database, List<Migration> migrations) {
+        for (Migration migration : migrations) {
+            Batch batch = migration.batch();
+            database.rawQuery(DELETE_QUERY, batch.getDownloadBatchId().stringValue());
```
jonreeve: `DELETE_BY_ID_QUERY` ?

library/src/test/java/com/novoda/downloadmanager/MigrationExtractorTest.java
```diff
@@ -0,0 +1,80 @@
...
+    @Before
+    public void setUp() {
+        Log.setShowLogs(false);
+        migrationExtractor = new MigrationExtractor(database);
+    }
+
+    @Test
+    public void returnsMigrations_WhenExtracting() {
+        given(database.rawQuery(BATCHES_QUERY)).willReturn(BATCHES_CURSOR);
+        given(database.rawQuery(eq(DOWNLOADS_QUERY), eq("1"))).willReturn(BATCH_ONE_DOWNLOADS_CURSOR);
```
jonreeve: If using eq for both do we need to supply the matchers? I had a feeling that was the default if all args were not matchers. Could be wrong tho

library/src/test/java/com/novoda/downloadmanager/MigrationExtractorTest.java
```diff
@@ -0,0 +1,80 @@
...
+    private MigrationExtractor migrationExtractor;
+
+    @Before
+    public void setUp() {
+        Log.setShowLogs(false);
+        migrationExtractor = new MigrationExtractor(database);
+    }
+
+    @Test
+    public void returnsMigrations_WhenExtracting() {
```
jonreeve: extractsMigrations would be shorter equivalent perhaps? nbd anyway tho

jonreeve: Oh maybe should mention cursor / DB tho, if we wanted to be more clear about what it does just from the name

library/src/main/java/com/novoda/downloadmanager/LiteFileName.java
Frankie-Boy: There are a lot of Classes that are prefixed with 'Lite', we'll tackle them all at a later date 👍

library/src/test/java/com/novoda/downloadmanager/StubCursor.java
jonreeve: Makes sense. There is a heck of a lot on Cursor.

Wish it didn't make sense. It's a shame Android deprecated MockCursor and just said "use Mockito", rather than providing a decent helpful fake.

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
jonreeve: Ah. Then even that I guess, and maybe not multiples of 512. It just lead me to believe it might be super important.

library/src/main/java/com/novoda/downloadmanager/MigrationFactory.java
jonreeve: Sorry I meant the logic as in the branching just above, kinda combined two comments.

As for the DB opening, it seems to me the same as putting too much in constructors. If I call a factory I wouldn't really expect it to fail, and I might be surprised it was opening a DB unless the thing I was constructing directly was a "DB" object.

library/src/main/java/com/novoda/downloadmanager/LiteFileName.java
jonreeve: Yeah tbh I noticed it more as I went on, so understand now 👍

library/src/test/java/com/novoda/downloadmanager/MigrationExtractorTest.java
Mecharyry: Yea I think you are right @jonreeve, I probably did one and then the other and forgot to change them 😄

PR #265, opened 2017-12-08
Port v1 db to v2

Mecharyry:
### Problem
It's a really long process to test the migration from v1 to v2 because we are constantly switching between the `v1` branch and the `v2` branch. We need to be able to create the `v1` database from the `v2` demo!

### Solution
Copy a version of the `v1` database that is stored in the project `assets` folder into the devices `databases` storage location. Access the now `cloned` database retrieve the `download file locations` and copy the download file stored in `assets` to the correct location. 💥  one clone of the v1 data 😄 

##### v1 DB created from the v2 demo
screen shot 2017-12-08 at 17 09 29

##### v2 database created from the v1 cloned data
screen shot 2017-12-08 at 17 10 46


demo-simple/src/main/java/com/novoda/downloadmanager/demo/VersionOneDatabaseCloner.java
```diff
@@ -0,0 +1,137 @@
...
+
+    private static final String DATABASE_NAME = "downloads.db";
+    private static final String DOWNLOAD_FILE_NAME = "file.zip";
+    private static final String ORIGINAL_FILE_LOCATION_QUERY = "SELECT _data FROM Downloads WHERE _data IS NOT NULL";
+    private static final int ORIGINAL_FILE_LOCATION_COLUMN_INDEX = 0;
+
+    private final AssetManager assetManager;
+    private final Executor executor;
+    private final UpdateListener updateListener;
+    private final Handler updateHandler;
```
joetimmins: This is basically the main thread right? Slightly weird having threads represented using both `Handler` and `Executor` but i'm not hugely bothered given the temporary nature of the code here 🤷‍♂️

PR #266, opened 2017-12-11
File size spinner

Mecharyry:
### Problem
We want to be able to change the file size when cloning the v1 database.

### Solution
Add a spinner option to select from a series of download file sizes.

### Screen Capture
device-2017-12-11-113709

### DIFF
--- a/demo-simple/src/main/java/com/novoda/downloadmanager/demo/DatabaseClonerFactory.java
+++ b/demo-simple/src/main/java/com/novoda/downloadmanager/demo/DatabaseClonerFactory.java
@@ -1,0 +1,42 @@
+package com.novoda.downloadmanager.demo;
+
+import android.content.Context;
+import android.content.res.AssetManager;
+import android.os.Handler;
+import android.os.Looper;
+
+import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
+
+final class DatabaseClonerFactory {
+
+    private DatabaseClonerFactory() {
+        // Uses static factory method.
+    }
+
+    static VersionOneDatabaseCloner databaseCloner(Context context, final VersionOneDatabaseCloner.CloneCallback cloneCallback) {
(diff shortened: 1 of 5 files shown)


demo-simple/src/main/java/com/novoda/downloadmanager/demo/DatabaseClonerFactory.java
```diff
@@ -0,0 +1,42 @@
...
+        // Uses static factory method.
+    }
+
+    static VersionOneDatabaseCloner databaseCloner(Context context, final VersionOneDatabaseCloner.CloneCallback cloneCallback) {
+        String originalDatabaseLocation = context.getFilesDir().getParentFile().getAbsolutePath() + "/databases/downloads.db";
+        Executor executor = Executors.newSingleThreadExecutor();
+        AssetManager assets = context.getAssets();
+        final Handler mainThreadHandler = new Handler(Looper.getMainLooper());
+
+        VersionOneDatabaseCloner.CloneCallback mainThreadReportingCloneCallback = new VersionOneDatabaseCloner.CloneCallback() {
```
Mecharyry: Decided to hide this callback on the `main thread` here to prevent unnecessary noise in the `VersionOneDatabaseCloner` class. Also reduces the amount of code in the `activity`.

demo-simple/src/main/res/values/strings.xml
```diff
@@ -1,3 +1,10 @@
 <resources>
   <string name="app_name">Lite Download Manager</string>
+
+  <string-array name="file_sizes">
+    <item>10MB</item>
+    <item>20MB</item>
+    <item>50MB</item>
+    <item>100MB</item>
```
Mecharyry: We can only go up to 100MB in file sizes, GH limits prevents larger files without additional customisation of the repo.

joetimmins: Boooooooo @github

PR #267, opened 2017-12-11
Migration Service

Mecharyry:
### Problem
We want to be able to run the migration process in a service so that it can survive being the app being killed by users. We also want to reschedule this service in the event that the system decides to kill it or the user decides to swipe away the app.

### Solution
Create a `LiteDownloadMigrationService`, implement a naive rescheduling mechanism when swipped away from recents. Pass callbacks to the `migrator` and display updates in the UI. This is still a naive approach and will be iterated upon as we work on the migration story.

### Paired with
@joetimmins and @Frankie-Boy a little at the beginning


demo-simple/build.gradle
```diff
@@ -1,13 +1,13 @@
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 27
-    buildToolsVersion "27.0.0"
+    compileSdkVersion 26
```
Mecharyry: These need to be reverted to the latest version before releasing the library, left at 26 for now so we can have access to the latest docs in the code.

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
```diff
@@ -0,0 +1,124 @@
...
+public class LiteDownloadMigrationService extends Service {
+
+    private Migrator v1ToV2Migrator;
+    private ExecutorService executor;
+    private IBinder binder;
+    private Migrator.Callback migrationCallback;
+    private String updateMessage;
+
+    @Override
+    public int onStartCommand(Intent intent, int flags, int startId) {
```
Mecharyry: We need to handle the case where this is called multiple times in a future PR.

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
```diff
@@ -0,0 +1,124 @@
...
+        super.onStartCommand(intent, flags, startId);
+
+        Log.e(getClass().getSimpleName(), "onStartCommand");
+
+        if (v1ToV2Migrator != null) {
+            executor.execute(new Runnable() {
```
Mecharyry: Moved to `onStartCommand` because this needs to re-execute whenever the service is killed and restarted without binding.

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+public class LiteDownloadMigrationService extends Service {
+
+    private Migrator v1ToV2Migrator;
+    private ExecutorService executor;
+    private IBinder binder;
+    private Migrator.Callback migrationCallback;
+    private String updateMessage;
```
Mecharyry: Need to store the previous value so that when binding we get the latest update message initially.

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
```diff
@@ -0,0 +1,124 @@
...
+                getApplicationContext(),
+                getDatabasePath("downloads.db"),
+                migrationCallback()
+        );
+
+        super.onCreate();
+    }
+
+    private Migrator.Callback migrationCallback() {
+        return new Migrator.Callback() {
```
Mecharyry: Wraps the `Migrator.Callback` so that we only forward when a `bind` has occurred.

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    @Nullable
+    @Override
+    public IBinder onBind(Intent intent) {
+        return binder;
+    }
+
+    @Override
+    public void onTaskRemoved(Intent rootIntent) {
```
Mecharyry: This is executed whenever a user swipes the app from recents. We also need to handle the case when the system kills a service / app.

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
```diff
@@ -0,0 +1,123 @@
...
+
+    class MigrationDownloadServiceBinder extends Binder {
+
+        MigrationDownloadServiceBinder withCallback(Migrator.Callback migrationCallback) {
+            LiteDownloadMigrationService.this.migrationCallback = migrationCallback;
+            return this;
+        }
+
+        void bind() {
+            LiteDownloadMigrationService.this.migrationCallback.onUpdate(LiteDownloadMigrationService.this.updateMessage);
```
Mecharyry: To send the initial message straight to the UI.

library/src/main/java/com/novoda/downloadmanager/MigrationFactory.java
```diff
@@ -11,7 +13,24 @@ private MigrationFactory() {
         // Uses static methods.
     }
 
-    public static Migrator createVersionOneToVersionTwoMigrator(Context context, File databasePath, Migrator.Callback callback) {
+    public static MigrationServiceBinder migrationServiceBinder(Context context, final Migrator.Callback migrationCallback) {
+        final Handler mainThreadHandler = new Handler(Looper.getMainLooper());
+        Migrator.Callback mainThreadReportingMigrationCallback = new Migrator.Callback() {
```
Mecharyry: I figured this was the best place to hide the fact that the `Service` uses a background thread and we need to report back to the `main thread`.

library/src/main/java/com/novoda/downloadmanager/MigrationServiceBinder.java
```diff
@@ -0,0 +1,52 @@
+package com.novoda.downloadmanager;
+
+import android.content.ComponentName;
+import android.content.Context;
+import android.content.Intent;
+import android.content.ServiceConnection;
+import android.os.IBinder;
+
+public class MigrationServiceBinder {
```
Mecharyry: Handles the binding without exposing `ServiceConnection` to the client applications of the library.

Mecharyry: This will be iterated upon to refine the definition of `bound`

library/src/main/java/com/novoda/downloadmanager/Migrator.java
```diff
@@ -5,7 +5,7 @@
     void migrate();
 
     interface Callback {
-        void onMigrationComplete();
```
Mecharyry: `onMigrationComplete` was exposed originally so we could delete the database using the `ContextWrapper`, found out we can do it by just deleting the `file` which doesn't require access to `Context` 💥

library/src/main/java/com/novoda/downloadmanager/SqlDatabaseWrapper.java
```diff
@@ -27,7 +28,14 @@ public Cursor rawQuery(String query, String selectionArgument, String... selecti
         return sqLiteDatabase.rawQuery(query, arguments.toArray(new String[arguments.size()]));
     }
 
-    void close() {
+    public void close() {
         sqLiteDatabase.close();
     }
+
+    public void deleteDatabase() {
```
Mecharyry: No `Context` required 💥

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -60,27 +64,20 @@ public void onClick(View view) {
...
-                }).start();
+                migrationServiceBinder.bind();
+            }
+        });
+
+        View buttonAbortMigration = findViewById(R.id.button_abort_migration);
+        buttonAbortMigration.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View view) {
+                migrationServiceBinder.unbind();
```
joetimmins: Does this actually work? We tried it on a Pixel C and the migration still completed before the service was unbound! Had to power the device off in the end

Mecharyry: `unbinding` doesn't mean killing the service unfortunately. See [here](https://stackoverflow.com/questions/17146822/when-is-a-started-and-bound-service-destroyed). 

I thought that this should work when calling `stopSelf` as the above suggests but it didn't seem to work. I think that perhaps there is a leak and that is why it doesn't work. 🤔  Perhaps we need to do `stopSelf` and terminate the executor. That's assuming we want the process to stop when `unbinding`.

Mecharyry: I think this needs more investigation. I haven't worked on `Service`s for the longest time and I forget the best way to do them.

Mecharyry:
test this please

PR #268, opened 2017-12-11
Delete orphaned data, fix two bugs

joetimmins:
When a previous migration hasn't completed properly, there can be files left in the client app's local storage that have no corresponding database entries and just sit there taking up space.

This pull request adds functionality that deletes any unlinked data in the client app's local storage.

We also found two bugs and fixed them. First bug, each v1 file was being copied to local storage twice, because the migrations were not being calculated correctly. Second bug, we weren't deleting the v1 files from their old locations after they were moved to the v2 location.

Paired with @Frankie-Boy


demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -108,30 +108,13 @@ public void onClick(View v) {
             }
         });
 
-        final TextView downloadedTitles = findViewById(R.id.textview_completed_download_title);
-        View buttonShowCompletedDownloads = findViewById(R.id.button_show_completed_downloads);
-        buttonShowCompletedDownloads.setOnClickListener(new View.OnClickListener() {
+        View buttonLogFilesDir = findViewById(R.id.btn_log_files_dir);
+        buttonLogFilesDir.setOnClickListener(new View.OnClickListener() {
             @Override
-            public void onClick(View view) {
```
joetimmins: This didn't actually do anything when it was clicked. Replaced it with a button that dumps the file names in the local storage directory to Logcat.

library/src/main/java/com/novoda/downloadmanager/AndroidLocalFilesDirectory.java
```diff
@@ -0,0 +1,24 @@
+package com.novoda.downloadmanager;
+
+import android.content.Context;
+
+import java.util.Arrays;
+import java.util.List;
+
+class AndroidLocalFilesDirectory implements LocalFilesDirectory {
```
joetimmins: This is just a simple wrapper around the `context`'s file operations, will be built on in subsequent PRs.

library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
```diff
@@ -2,14 +2,12 @@
...
 
-import com.novoda.notils.logger.simple.Log;
-
 import java.util.ArrayList;
 import java.util.List;
 
 class MigrationExtractor {
 
-    private static final String BATCHES_QUERY = "SELECT batches._id, batches.batch_title FROM batches INNER JOIN DownloadsByBatch ON " +
+    private static final String BATCHES_QUERY = "SELECT DISTINCT batches._id, batches.batch_title FROM batches INNER JOIN DownloadsByBatch ON " +
```
joetimmins: Here's the fix for bug one - it was effectively returning two rows for each batch, hence double the required number of migrations were occurring

library/src/main/java/com/novoda/downloadmanager/RoomAppDatabase.java
```diff
@@ -1,12 +1,31 @@
...
 
 import android.arch.persistence.room.Database;
+import android.arch.persistence.room.Room;
 import android.arch.persistence.room.RoomDatabase;
+import android.content.Context;
 
 @Database(entities = {RoomBatch.class, RoomFile.class}, version = 1)
 abstract class RoomAppDatabase extends RoomDatabase {
 
+    private static RoomAppDatabase SINGLE_INSTANCE;
```
joetimmins: It's possible that we can be using the `RoomAppDatabase` instance in two places at once now - once in the migrator and once in the actual downloads service, if the user decides to initiate a download while the migration process is happening. We will need to test this scenario, but for now, we followed the Room guidelines and made this a singleton.

library/src/main/java/com/novoda/downloadmanager/RoomFileDao.java
```diff
@@ -15,4 +15,8 @@
 
     @Query("SELECT * FROM RoomFile WHERE RoomFile.batch_id = :batchId")
     List<RoomFile> loadAllFilesFor(String batchId);
+
+    // Only to be used from the UnlinkedDataRemover
+    @Query("SELECT * FROM RoomFile")
+    List<RoomFile> loadAllFiles();
```
joetimmins: Added so we can easily get a list of all files v2 properly knows about.

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
```diff
@@ -135,7 +139,11 @@ private void deleteFrom(SqlDatabaseWrapper database, List<Migration> migrations)
         for (Migration migration : migrations) {
             Batch batch = migration.batch();
             database.rawQuery(DELETE_BY_ID_QUERY, batch.getDownloadBatchId().stringValue());
+            for (Migration.FileMetadata metadata : migration.getFileMetadata()) {
```
joetimmins: Fix for bug two, will need further testing and the TODO addressing in subsequent PRs.

demo-simple/src/main/res/layout/activity_main.xml
```diff
@@ -125,17 +125,9 @@
     android:text="Delete all" />
 
   <Button
-    android:id="@+id/button_show_completed_downloads"
+    android:id="@+id/btn_log_files_dir"
```
Mecharyry: Avoid abbreviations whenever possible and please respect the naming convention already in the project, it avoids confusion and unnecessary maintenance down the line.

demo-simple/src/main/res/layout/activity_main.xml
```diff
@@ -125,17 +125,9 @@
...
-    android:text="Show Completed Downloads" />
-
-  <TextView
-    android:id="@+id/textview_completed_download_title"
-    android:layout_width="match_parent"
-    android:layout_height="wrap_content"
-    android:text="TextView"
-    android:textColor="@android:color/black"
-    android:textSize="16sp" />
+    android:text="Log Files Dir" />
```
Mecharyry: `Log File Directory`

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -108,30 +108,13 @@ public void onClick(View v) {
...
-                                    + "\ndownloaded! "
-                                    + "\nbytes: " + status.bytesDownloaded()
-                                    + "\n\n";
-                            displayMessage.append(message);
-                        }
-                        downloadedTitles.setText(displayMessage.toString());
-                    }
-                });
+            public void onClick(View v) {
+                for (String s : getFilesDir().list()) {
```
Mecharyry: You didn't use the wrapped version here? `AndroidLocalFilesDirectory`

library/src/main/java/com/novoda/downloadmanager/RoomFileDao.java
```diff
@@ -15,4 +15,8 @@
 
     @Query("SELECT * FROM RoomFile WHERE RoomFile.batch_id = :batchId")
     List<RoomFile> loadAllFilesFor(String batchId);
+
+    // Only to be used from the UnlinkedDataRemover
```
Mecharyry: I think this comment is unnecessary, `loadAllFilesFor` is also used in one place. It's fine because we don't expose the database to clients.

library/src/main/java/com/novoda/downloadmanager/RoomV2DatabaseFiles.java
```diff
@@ -0,0 +1,23 @@
+package com.novoda.downloadmanager;
+
+import java.util.ArrayList;
+import java.util.List;
+
+class RoomV2DatabaseFiles implements V2DatabaseFiles {
+    private final RoomFileDao roomFileDao;
```
Mecharyry: Spacing above this line.

library/src/main/java/com/novoda/downloadmanager/RoomV2DatabaseFiles.java
```diff
@@ -0,0 +1,23 @@
...
+
+    RoomV2DatabaseFiles(RoomFileDao roomFileDao) {
+        this.roomFileDao = roomFileDao;
+    }
+
+    @Override
+    public List<String> databaseContents() {
```
Mecharyry: I understand the naming here but I think this is wrong. We are leaking that the backing data-structure is a database.

library/src/main/java/com/novoda/downloadmanager/RoomV2DatabaseFiles.java
```diff
@@ -0,0 +1,23 @@
...
+        List<RoomFile> roomFilesList = roomFileDao.loadAllFiles();
+        List<String> stringFilesList = new ArrayList<>();
```
Mecharyry: `downloadFilesAsStrings` or similar.

library/src/main/java/com/novoda/downloadmanager/UnlinkedDataRemover.java
```diff
@@ -0,0 +1,20 @@
+package com.novoda.downloadmanager;
+
+class UnlinkedDataRemover {
+    private final LocalFilesDirectory localFilesDirectory;
```
Mecharyry: Spacing above

library/src/main/java/com/novoda/downloadmanager/UnlinkedDataRemover.java
```diff
@@ -0,0 +1,20 @@
...
+    private final V2DatabaseFiles v2DatabaseFiles;
+
+    UnlinkedDataRemover(LocalFilesDirectory localFilesDirectory, V2DatabaseFiles v2DatabaseFiles) {
+        this.localFilesDirectory = localFilesDirectory;
+        this.v2DatabaseFiles = v2DatabaseFiles;
+    }
+
+    void remove() {
+        for (String filename : localFilesDirectory.contents()) {
+            if (!v2DatabaseFiles.databaseContents().contains(filename)) {
```
Mecharyry: This method kind of represents my point from before. It doesn't need to know or care that the backing data-structure is a database.

`if(!downloadFiles.fileNames().contains(fileName)))`

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
```diff
@@ -17,25 +17,29 @@
     private static final String DELETE_BY_ID_QUERY = "DELETE FROM batches WHERE _id = ?";
 
     private final MigrationExtractor migrationExtractor;
-    private final RoomDownloadsPersistence downloadsPersistence;
```
Mecharyry: Better naming 👍  don't expose that it is backed by `Room`

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
```diff
@@ -135,7 +139,11 @@ private void deleteFrom(SqlDatabaseWrapper database, List<Migration> migrations)
         for (Migration migration : migrations) {
             Batch batch = migration.batch();
             database.rawQuery(DELETE_BY_ID_QUERY, batch.getDownloadBatchId().stringValue());
+            for (Migration.FileMetadata metadata : migration.getFileMetadata()) {
+                // TODO: Handle error cases, pay attention to file.delete() result
```
Mecharyry: Would be better represented as a ticket in GH `Issues` with a link here to the ticket. This way you are forced to provide more information than just `Handle error cases` which someone new on the project would not understand.

library/src/test/java/com/novoda/downloadmanager/UnlinkedDataRemoverTest.java
```diff
@@ -0,0 +1,72 @@
...
+import static com.google.common.truth.Truth.assertThat;
+
+public class UnlinkedDataRemoverTest {
+    private static final String FIRST_FILE = "a filename";
+    private static final String SECOND_FILE = "another filename";
+    private static final String THIRD_FILE = "yet another filename"; // Orphan
+
+    @Test
+    public void givenUnlinkedFilesInLocalStorage_whenRemoving_thenUnlinkedFilesAreDeleted() throws Exception {
+        // Arrange
```
Mecharyry: Would remove comments in favour of extracting more explanatory methods. This project doesn't follow the `given_when_then` structure but prefers `action_whenSomethingOccurs` see [here](https://github.com/novoda/download-manager/blob/epic_download_manager_v2/library/src/test/java/com/novoda/downloadmanager/CallbackThrottleByTimeTest.java)

library/src/test/java/com/novoda/downloadmanager/UnlinkedDataRemoverTest.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import org.junit.Test;
+
+import static com.google.common.truth.Truth.assertThat;
+
+public class UnlinkedDataRemoverTest {
+    private static final String FIRST_FILE = "a filename";
```
Mecharyry: Spacing above

library/src/test/java/com/novoda/downloadmanager/UnlinkedDataRemoverTest.java
```diff
@@ -0,0 +1,72 @@
...
+                    newFileList.add(s);
+                }
+            }
+            fileList = newFileList;
+            return fileWasRemoved;
+        }
+    }
+
+    private static class FakeV2DatabaseFiles implements V2DatabaseFiles {
+        private final List<String> fileList;
```
Mecharyry: Spacing above.

demo-simple/src/main/res/layout/activity_main.xml
```diff
@@ -125,17 +125,9 @@
     android:text="Delete all" />
 
   <Button
-    android:id="@+id/button_show_completed_downloads"
+    android:id="@+id/button_show_files_in_directory"
```
Mecharyry: Naming doesn't match the text or the variable name in `MainActivity`. If you make these consistent it will be easier to find when looking through the codebase.

library/src/main/java/com/novoda/downloadmanager/AndroidLocalFilesDirectory.java
```diff
@@ -0,0 +1,24 @@
+package com.novoda.downloadmanager;
+
+import android.content.Context;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class AndroidLocalFilesDirectory implements LocalFilesDirectory {
+    private final Context context;
```
Mecharyry: Spacing above.

library/src/main/java/com/novoda/downloadmanager/LocalFilesDirectory.java
```diff
@@ -0,0 +1,9 @@
+package com.novoda.downloadmanager;
+
+import java.util.List;
+
+interface LocalFilesDirectory {
+    List<String> contents();
```
Mecharyry: Spacing above.

library/src/main/java/com/novoda/downloadmanager/RoomV2DatabaseFiles.java
```diff
@@ -0,0 +1,24 @@
+package com.novoda.downloadmanager;
+
+import java.util.ArrayList;
+import java.util.List;
+
+class RoomV2DatabaseFiles implements V2DatabaseFiles {
```
Mecharyry: I'm not completely sold on this class TBH. 

1. I think the naming is confusing perhaps it could be `RoomFiles` but then it doesn't contain files, just names.
2. Part of me thinks this could just exist on the `RoomAppDatabase` object where you call `roomFileNames()` and it does the `fileNames()` method there instead. 
3. The interface feels unnecessary but I guess was added for testing? But it seems like you could `mock` the `RoomFileDao` or the whole `Database` instead.

library/src/main/java/com/novoda/downloadmanager/UnlinkedDataRemover.java
```diff
@@ -0,0 +1,21 @@
...
+    private final LocalFilesDirectory localFilesDirectory;
+    private final V2DatabaseFiles downloadFiles;
+
+    UnlinkedDataRemover(LocalFilesDirectory localFilesDirectory, V2DatabaseFiles downloadFiles) {
+        this.localFilesDirectory = localFilesDirectory;
+        this.downloadFiles = downloadFiles;
+    }
+
+    void remove() {
+        for (String fileName : localFilesDirectory.contents()) {
```
Mecharyry: Could do `RoomV2DatabaseFiles.fileNames` here instead of passing through the `V2DatabaseFiles` 🤔

library/src/main/java/com/novoda/downloadmanager/RoomAppDatabase.java
```diff
@@ -1,12 +1,44 @@
...
+
     abstract RoomBatchDao roomBatchDao();
 
     abstract RoomFileDao roomFileDao();
+
+    List<String> fileNames() {
```
Mecharyry: @joetimmins I've moved the logic that was in `V2DatabaseFiles` to here. Perhaps we should consider adding an interface to this `Database` so we don't leak `Room` everywhere in the app.

library/src/test/java/com/novoda/downloadmanager/UnlinkedDataRemoverTest.java
```diff
@@ -0,0 +1,62 @@
...
+    private static final String LINKED_SECOND_FILE = "another filename";
+    private static final String UNLINKED_THIRD_FILE = "yet another filename";
+
+    private final RoomAppDatabase roomAppDatabase = mock(RoomAppDatabase.class);
+    private final LocalFilesDirectory localFilesDirectory = new FakeLocalFilesDirectory(Arrays.asList(LINKED_FIRST_FILE, LINKED_SECOND_FILE, UNLINKED_THIRD_FILE));
+    private final UnlinkedDataRemover unlinkedDataRemover = new UnlinkedDataRemover(roomAppDatabase, localFilesDirectory);
+
+    @Test
+    public void removesFiles_whenFileIsUnlinked() {
+        given(roomAppDatabase.fileNames()).willReturn(Arrays.asList(LINKED_FIRST_FILE, LINKED_SECOND_FILE));
```
Mecharyry: @joetimmins needed to add the `mock` when moving the `fileNames` to the `RoomAppDatabase` class. Would prefer to expose an interface on the `Database` to avoid leaking `Room` and so we can create a nice `Dummy` implementation for tests.

library/src/main/java/com/novoda/downloadmanager/AndroidLocalFilesDirectory.java
```diff
@@ -0,0 +1,25 @@
+package com.novoda.downloadmanager;
+
+import android.content.Context;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class AndroidLocalFilesDirectory implements LocalFilesDirectory {
```
zegnus: You can avoid making this public if this class is constructed by a factory, then only the factory and the interface is exposed and you keep the implementation detail private

library/src/main/java/com/novoda/downloadmanager/RoomAppDatabase.java
```diff
@@ -1,12 +1,44 @@
...
+        List<String> fileNames = new ArrayList<>();
+
+        for (RoomFile roomFile : roomFiles) {
+            fileNames.add(roomFile.name);
+        }
+        return fileNames;
+    }
+
+    static RoomAppDatabase obtainInstance(Context context) {
+        if (SINGLE_INSTANCE == null) {
```
zegnus: this is not thread safe, if you need it to be use the lazy init method

library/src/main/java/com/novoda/downloadmanager/RoomAppDatabase.java
zegnus: mmm the app doesn't currently know that there is room under-the-hood because it only talks to `DownloadsPersistence`

library/src/main/java/com/novoda/downloadmanager/UnlinkedDataRemover.java
```diff
@@ -0,0 +1,25 @@
+package com.novoda.downloadmanager;
+
+import java.util.List;
+
+class UnlinkedDataRemover {
+
+    private final RoomAppDatabase database;
```
zegnus: This **must** be database agnostic, the rest of the code only talks to the interface `DownloadsPersistence`

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
```diff
@@ -17,25 +17,29 @@
     private static final String DELETE_BY_ID_QUERY = "DELETE FROM batches WHERE _id = ?";
 
     private final MigrationExtractor migrationExtractor;
-    private final RoomDownloadsPersistence downloadsPersistence;
+    private final DownloadsPersistence downloadsPersistence;
```
zegnus: 👍

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
```diff
@@ -17,25 +17,29 @@
...
     private final InternalFilePersistence internalFilePersistence;
     private final SqlDatabaseWrapper database;
     private final Callback migrationCallback;
+    private final UnlinkedDataRemover unlinkedDataRemover;
 
     VersionOneToVersionTwoMigrator(MigrationExtractor migrationExtractor,
                                    RoomDownloadsPersistence downloadsPersistence,
```
zegnus: This **must** be `DownloadsPersistence` instead

library/src/main/java/com/novoda/downloadmanager/UnlinkedDataRemover.java
Mecharyry: `DownloadsPersistence` is not database agnostic using this logic. If one place needs to do it then all should be expected to do it by using an interface. That was kind of my point.

Mecharyry: Ah! I didn't realise that `DownloadsPersistence` was the interface I was looking for 😂

library/src/main/java/com/novoda/downloadmanager/RoomAppDatabase.java
joetimmins: Yeah this is really the wrong place for this code to live. The class that extends RoomDatabase should basically just return `Dao`s, the `obtainInstance` method just prevents us having another class to act as a singleton for this one.

The idea of having a `V2DatabaseFiles` interface was precisely to stop the rest of the app having a concrete dependency on a Room class. Then, the `RoomV2DatabaseFiles` class was the Room-specific implementation of that interface. Had this been a proper app and not a library, that Room-specific implementation would have lived in a very small package and set to package-private, so as not to leak implementation details, but that isn't possible in a library.

zegnus: I also think that this shouldn't be here

library/src/main/java/com/novoda/downloadmanager/AndroidLocalFilesDirectory.java
joetimmins: We use it in the activity so it has to be public.

library/src/main/java/com/novoda/downloadmanager/RoomAppDatabase.java
```diff
@@ -1,12 +1,31 @@
...
 
 import android.arch.persistence.room.Database;
+import android.arch.persistence.room.Room;
 import android.arch.persistence.room.RoomDatabase;
+import android.content.Context;
 
 @Database(entities = {RoomBatch.class, RoomFile.class}, version = 1)
 abstract class RoomAppDatabase extends RoomDatabase {
 
+    private static volatile RoomAppDatabase SINGLE_INSTANCE;
```
joetimmins: @zegnus couldn't use the inner "holder" class to make this thread safe as it depends on `Context`. This should do the job though.

library/src/test/java/com/novoda/downloadmanager/FakeDownloadsPersistence.java
```diff
@@ -0,0 +1,67 @@
+package com.novoda.downloadmanager;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+class FakeDownloadsPersistence implements DownloadsPersistence {
```
joetimmins: Hopefully this will be built upon and reused in future work.

library/src/main/java/com/novoda/downloadmanager/AndroidLocalFilesDirectory.java
zegnus: Only a factory and the interface needs to be public, the specific implementation can be private. If you go through a factory+interface then you can evolve and change the specific implementation and the public API keeps being unchanged along with all the public classes.

`LocalFilesDirectory  localFilesDirectory = LocalFilesDirectoryFactory.create(...)` should be enough

or maybe I'm missing something?

Mecharyry: As far as I am aware you are completely correct @zegnus. You can expose a factory in the download manager that will return the interface `LocalFilesDirectory` hiding the concrete implementation. 💯

library/src/test/java/com/novoda/downloadmanager/UnlinkedDataRemoverTest.java
```diff
@@ -0,0 +1,76 @@
...
+        public List<String> contents() {
+            return fileList;
+        }
+
+        @Override
+        public boolean deleteFile(String filename) {
+            boolean fileWasRemoved = false;
+            List<String> newFileList = new ArrayList<>();
+            for (String s : fileList) {
+                if (filename.equals(s)) {
```
Mecharyry: I don't really agree with having logic in this sort of class. For me it should just take a list or a map if you want to ensure that you return true or false for a given file.

library/src/main/java/com/novoda/downloadmanager/UnlinkedDataRemover.java
```diff
@@ -0,0 +1,32 @@
...
+    private final LocalFilesDirectory localFilesDirectory;
+
+    UnlinkedDataRemover(DownloadsPersistence downloadsPersistence, LocalFilesDirectory localFilesDirectory) {
+        this.downloadsPersistence = downloadsPersistence;
+        this.localFilesDirectory = localFilesDirectory;
+    }
+
+    void remove() {
+        List<String> filesInV2Database = new ArrayList<>();
+        for (DownloadsBatchPersisted batch : downloadsPersistence.loadBatches()) {
```
Mecharyry: Surprised you didn't leverage an SQL query to do most of the heavy lifting here, any particular reason? 

"SELECT * FROM RoomFile"

joetimmins: @zegnus said in Slack that we shouldn't add any more methods to the `DownloadsPersistence` interface.

library/src/test/java/com/novoda/downloadmanager/UnlinkedDataRemoverTest.java
joetimmins: This is just a workaround for Java's infamous `ConcurrentModificationException`.

library/src/main/java/com/novoda/downloadmanager/UnlinkedDataRemover.java
Mecharyry: Really @zegnus ? Can't we add this?

library/src/test/java/com/novoda/downloadmanager/UnlinkedDataRemoverTest.java
Mecharyry: 😬

Mecharyry:
@joetimmins for me I think the outstanding stuff is:

- `AndroidLocalFilesDirectory` should not be exposed, we should just use the interface, hide it behind a factory.
- If @zegnus is ok with it, add the query to the `DownloadsPersistence`.
- Use `DownloadsPersistence` instead of `RoomDownloadsPersistence`.

library/src/main/java/com/novoda/downloadmanager/AndroidLocalFilesDirectory.java
```diff
@@ -0,0 +1,25 @@
...
+import android.content.Context;
+
+import java.util.Arrays;
+import java.util.List;
+
+class AndroidLocalFilesDirectory implements LocalFilesDirectory {
+
+    private final Context context;
+
+    public AndroidLocalFilesDirectory(Context context) {
```
Mecharyry: Constructor can be package now.

PR #269, opened 2017-12-12
Service improvements

Mecharyry:
### Problem
There are some outstanding issues with the `MigrationService` including:

* `bind` can be called multiple times and it will spawn off multiple executors to perform the migration.
* `unbind` can be called multiple times and we don't guard against `nulls` in the `binder` 😬 

### Solution
* Make `Migrator` single instance and expose an `isRunning` state to ensure that `onStartCommand` is only executed once until the `Migrator` completes.
* Expose `isBound` method on `Service` so we avoid `unbinding` from a `Service` which has already been unbound.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
+++ b/library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
@@ -17,8 +17,9 @@ import java.util.concurrent.Executors;
 public class LiteDownloadMigrationService extends Service {
 
     private static final String TAG = "MigrationService";
+    private static ExecutorService executor;
+
     private Migrator v1ToV2Migrator;
-    private ExecutorService executor;
     private IBinder binder;
     private Migrator.Callback migrationCallback;
     private String updateMessage;
@@ -32,18 +33,15 @@ public class LiteDownloadMigrationService extends Service {
     @Override
     public int onStartCommand(Intent intent, int flags, int startId) {
         super.onStartCommand(intent, flags, startId);
-
         Log.d(TAG, "onStartCommand");
 
-        if (v1ToV2Migrator != null && !v1ToV2Migrator.isRunning()) {
-            executor.execute(new Runnable() {
-                @Override
-                public void run() {
-                    Log.d(TAG, "Begin Migration");
-                    v1ToV2Migrator.migrate();
-                }
-            });
-        }
+        executor.execute(new Runnable() {
+            @Override
+            public void run() {
+                Log.d(TAG, "Begin Migration");
+                v1ToV2Migrator.migrate();
+            }
+        });
 
         return START_STICKY;
     }
@@ -51,7 +49,10 @@ public class LiteDownloadMigrationService extends Service {
     @Override
     public void onCreate() {
         Log.d(TAG, "onCreate");
-        executor = Executors.newSingleThreadExecutor();
+        if (executor == null) {
+            executor = Executors.newSingleThreadExecutor();
+        }
+
         binder = new MigrationDownloadServiceBinder();
         v1ToV2Migrator = MigrationFactory.createVersionOneToVersionTwoMigrator(
                 getApplicationContext(),
--- a/library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
+++ b/library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
@@ -16,6 +16,7 @@ import java.util.concurrent.Executors;
 
 public class LiteDownloadMigrationService extends Service {
 
+    private static final String TAG = "MigrationService";
     private Migrator v1ToV2Migrator;
     private ExecutorService executor;
     private IBinder binder;
@@ -26,13 +27,13 @@ public class LiteDownloadMigrationService extends Service {
     public int onStartCommand(Intent intent, int flags, int startId) {
         super.onStartCommand(intent, flags, startId);
 
-        Log.d(getClass().getSimpleName(), "onStartCommand");
+        Log.d(TAG, "onStartCommand");
 
-        if (v1ToV2Migrator != null) {
+        if (v1ToV2Migrator != null && !v1ToV2Migrator.isRunning()) {
             executor.execute(new Runnable() {
                 @Override
                 public void run() {
-                    Log.d(getClass().getSimpleName(), "Begin Migration");
+                    Log.d(TAG, "Begin Migration");
                     v1ToV2Migrator.migrate();
                 }
             });
@@ -43,7 +44,7 @@ public class LiteDownloadMigrationService extends Service {
 
     @Override
     public void onCreate() {
-        Log.d(getClass().getSimpleName(), "onCreate");
+        Log.d(TAG, "onCreate");
         executor = Executors.newSingleThreadExecutor();
         binder = new MigrationDownloadServiceBinder();
         v1ToV2Migrator = MigrationFactory.createVersionOneToVersionTwoMigrator(
@@ -75,9 +76,9 @@ public class LiteDownloadMigrationService extends Service {
 
     @Override
     public void onTaskRemoved(Intent rootIntent) {
-        Log.d(getClass().getSimpleName(), "onTaskRemoved");
+        Log.d(TAG, "onTaskRemoved");
         rescheduleMigration();
-        Log.d(getClass().getSimpleName(), "Rescheduling");
+        Log.d(TAG, "Rescheduling");
         super.onTaskRemoved(rootIntent);
     }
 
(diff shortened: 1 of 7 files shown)


Mecharyry:
test this please

library/src/main/java/com/novoda/downloadmanager/MigrationFactory.java
```diff
@@ -35,14 +35,19 @@ static Migrator createVersionOneToVersionTwoMigrator(Context context, File datab
...
 
-        SQLiteDatabase sqLiteDatabase = SQLiteDatabase.openDatabase(databasePath.getAbsolutePath(), null, 0);
-        SqlDatabaseWrapper database = new SqlDatabaseWrapper(sqLiteDatabase);
-
-        MigrationExtractor migrationExtractor = new MigrationExtractor(database);
-        RoomDownloadsPersistence downloadsPersistence = RoomDownloadsPersistence.newInstance(context);
-        InternalFilePersistence internalFilePersistence = new InternalFilePersistence();
-        internalFilePersistence.initialiseWith(context);
-        return new VersionOneToVersionTwoMigrator(migrationExtractor, downloadsPersistence, internalFilePersistence, database, migrationCallback);
+        if (INSTANCE == null) {
```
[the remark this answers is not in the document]
Mecharyry: I figured it was better to have the one return and just modify the `INSTANCE` rather than having two 🤔  now I am unsure 😂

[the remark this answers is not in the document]
takecare: always struggle with multiple return points vs nesting as well. if you considered that it's fine. happy to merge or wait for another review.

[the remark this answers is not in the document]
Mecharyry: I'd like @joetimmins or @zegnus to take a look too @takecare. Thanks for reviewing 😃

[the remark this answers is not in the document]
takecare: _nudge_

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
```diff
@@ -16,23 +16,30 @@
 
 public class LiteDownloadMigrationService extends Service {
 
+    private static final String TAG = "MigrationService";
     private Migrator v1ToV2Migrator;
     private ExecutorService executor;
```
zegnus: If you make this executor to be static and you only create one when there is no instance, then you can avoid the Migrator to be a singleton. Might we worth to double check that my assumption is correct and that if there is a task running in the executor and you kill the service, if you start the service again then the executor instance is the same...

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
```diff
@@ -16,23 +16,30 @@
...
+
     @Override
     public int onStartCommand(Intent intent, int flags, int startId) {
         super.onStartCommand(intent, flags, startId);
 
-        Log.d(getClass().getSimpleName(), "onStartCommand");
+        Log.d(TAG, "onStartCommand");
 
-        if (v1ToV2Migrator != null) {
+        if (v1ToV2Migrator != null && !v1ToV2Migrator.isRunning()) {
```
zegnus: can `v1ToV2Migrator` be null?

`.isRunning()` can be kept private and removed from the interface, you can call the migrator and if its running will just do nothing. But if you use a static executor, as only allows one task at a time, you don't even need to check for this.

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
```diff
@@ -16,23 +16,30 @@
...
             executor.execute(new Runnable() {
                 @Override
                 public void run() {
-                    Log.d(getClass().getSimpleName(), "Begin Migration");
+                    Log.d(TAG, "Begin Migration");
                     v1ToV2Migrator.migrate();
```
zegnus: you could annotate this method as `workingThread`

library/src/main/java/com/novoda/downloadmanager/Migrator.java
```diff
@@ -13,5 +15,10 @@
         public void migrate() {
             // no-op.
         }
+
+        @Override
+        public boolean isRunning() {
```
zegnus: What happens if you actually return the NO_OP? I think that is not processed anywhere

library/src/main/java/com/novoda/downloadmanager/MigrationServiceBinder.java
```diff
@@ -29,9 +29,11 @@ public void bind() {
     }
 
     public void unbind() {
-        context.unbindService(serviceConnection);
-        context.stopService(new Intent(context, LiteDownloadMigrationService.class));
-        serviceConnection = null;
+        if (LiteDownloadMigrationService.isBound() && migrationCallback != null) {
```
zegnus: When you bind here, you can hold a reference to the service so that this can be non-static. Also I'm not sure all the static access to the service is needed (regarding `MigrationDownloadServiceBinder`), you can check how is done here https://github.com/novoda/download-manager/blob/epic_download_manager_v2/library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java#L166

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
Mecharyry: Good thinking! Reduce the codz 😄

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
Mecharyry: No, it can't be `null` really 😄  I'll remove both of these and go for the `static` executor.

library/src/main/java/com/novoda/downloadmanager/Migrator.java
Mecharyry: It executes the `Migrator` in the same way, it just doesn't do anything.

library/src/main/java/com/novoda/downloadmanager/MigrationServiceBinder.java
Mecharyry: It seems that we never `unbind` the `DownloadService` can you explain the reasoning? As far as I am aware we should always `unbind` where a corresponding `bind` exists. 

I'll change the above to introduce an interface anyway and use that instead of having the `isBound` etc.

library/src/main/java/com/novoda/downloadmanager/MigrationServiceBinder.java
```diff
@@ -19,19 +19,22 @@
     }
 
     public void bind() {
-        if (serviceConnection == null) {
-            serviceConnection = new MigrationServiceConnection();
+        if (serviceConnection != null) {
```
Mecharyry: We are assuming that this is only being executed on an `Application` level and therefore will only be allowed to be bound once.

library/src/main/java/com/novoda/downloadmanager/SqlDatabaseWrapper.java
```diff
@@ -16,10 +16,6 @@ public SqlDatabaseWrapper(SQLiteDatabase sqLiteDatabase) {
         this.sqLiteDatabase = sqLiteDatabase;
     }
 
-    public boolean isOpen() {
```
Mecharyry: @zegnus don't need this after moving the creation point of the `Migrator`.

library/src/main/java/com/novoda/downloadmanager/MigrationFactory.java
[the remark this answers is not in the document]
joetimmins: I think what you've done with the `INSTANCE` here is fine but it could help readability to pull out another method for lines 39-45. So you end up with something like 
```
        if (INSTANCE == null) {
            INSTANCE = buildMigrator(context, databasePath, migrationCallback);
        }
        return INSTANCE;
```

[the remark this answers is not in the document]
joetimmins: Or even 
```
return INSTANCE == null ? createInstance(context, databasePath, migrationCallback) : INSTANCE;
```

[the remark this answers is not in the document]
zegnus: Because the singleton is the executor, this is no longer needed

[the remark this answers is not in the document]
Mecharyry: @zegnus if you don't have this then you are always stuck with the `V1Migrator` even when it is not being used, it's allocating more memory than the `no-op` which is why I kept this.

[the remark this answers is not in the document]
Mecharyry: Yes, you are right @zegnus. We use the executor and the `ServiceBinder` to drive the single instance nature.

Mecharyry:
Removed the Single Instance, it's not required anymore. The Service and the Executor should be single instance, preventing any issues.

PR #271, opened 2017-12-13
Batch by batch migration

Mecharyry:
### Problem
The migration from the v1 database to the v2 database can be stopped at any point if the system runs out of resources. This can currently leave the migration process in an inconsistent state because we process in steps e.g. move files and then move the data.

### Solution
Perform the migration in batches, where the migration of a batch is performed in a transaction. If the transaction never completes the changes will be rolled back leaving the databases in a consistent state.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
+++ b/library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
@@ -14,7 +14,8 @@ class VersionOneToVersionTwoMigrator implements Migrator {
     private static final String TAG = "V1 to V2 migrator";
 
     private static final int RANDOMLY_CHOSEN_BUFFER_SIZE_THAT_SEEMS_TO_WORK = 4096;
-    private static final String DELETE_BY_ID_QUERY = "DELETE FROM batches WHERE _id = ?";
+    private static final String TABLE_BATCHES = "batches";
+    private static final String WHERE_CLAUSE_ID = "_id = ?";
 
     private final MigrationExtractor migrationExtractor;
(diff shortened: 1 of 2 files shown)


library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
```diff
@@ -47,103 +48,103 @@ public void migrate() {
...
+            );
+            downloadsPersistence.persistFile(persistedFile);
         }
     }
 
     // TODO: See https://github.com/novoda/download-manager/issues/270
-    private void deleteFrom(SqlDatabaseWrapper database, List<Migration> migrations) {
-        for (Migration migration : migrations) {
-            Batch batch = migration.batch();
-            database.rawQuery(DELETE_BY_ID_QUERY, batch.getDownloadBatchId().stringValue());
```
Mecharyry: This wasn't working before but it didn't matter because we just dropped the whole db anyway 😂

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
```diff
@@ -47,103 +48,103 @@ public void migrate() {
...
+            database.startTransaction();
+
+            migrateV1FilesToV2Location(migration);
+            migrateV1DataToV2Database(migration);
+            deleteFrom(database, migration);
+
+            downloadsPersistence.transactionSuccess();
+            downloadsPersistence.endTransaction();
+            database.setTransactionSuccessful();
+            database.endTransaction();
```
Mecharyry: End both transactions at the same time just in case the process is terminated during method execution.

PR #272, opened 2017-12-14
Android Oreo Notification Channel

Mecharyry:
### Problem
`download-manager` v2 crashes 😱  under the following scenario:

1. Start the demo application
2. Start a download
3. Application crash on API 27, check logs
4. Application does not show a notification on API 26, there is a warning on the usage of the channel

### Solution
Add a Notification Channel for the standard download notifications that are sent by the `download-manager`.

### DIFF
--- a/demo-simple/src/main/java/com/novoda/downloadmanager/demo/CustomNotificationCreator.java
+++ b/demo-simple/src/main/java/com/novoda/downloadmanager/demo/CustomNotificationCreator.java
@@ -4,14 +4,16 @@ import android.app.Notification;
 import android.content.Context;
 import android.support.v4.app.NotificationCompat;
 
-import com.novoda.notils.logger.simple.Log;
 import com.novoda.downloadmanager.DownloadBatchTitle;
 import com.novoda.downloadmanager.NotificationCreator;
 import com.novoda.downloadmanager.NotificationInformation;
+import com.novoda.notils.logger.simple.Log;
 
 public class CustomNotificationCreator implements NotificationCreator {
+
     private static final int ID = 1;
     private static final boolean NOT_INDETERMINATE = false;
+    private static final String CHANNEL = "download_manager";
 
     private final Context context;
     private final int iconDrawable;
@@ -31,7 +33,7 @@ public class CustomNotificationCreator implements NotificationCreator {
 
         Log.v("Create notification for " + title + ", " + content);
 
-        Notification notification = new NotificationCompat.Builder(context)
+        Notification notification = new NotificationCompat.Builder(context, CHANNEL)
                 .setProgress(bytesFileSize, bytesDownloaded, NOT_INDETERMINATE)
                 .setSmallIcon(iconDrawable)
                 .setContentTitle(title)
--- a/demo-simple/src/main/java/com/novoda/downloadmanager/demo/CustomNotificationCreator.java
+++ b/demo-simple/src/main/java/com/novoda/downloadmanager/demo/CustomNotificationCreator.java
@@ -5,6 +5,7 @@ import android.content.Context;
 import android.support.v4.app.NotificationCompat;
 
 import com.novoda.downloadmanager.DownloadBatchTitle;
+import com.novoda.downloadmanager.NotificationChannelCreator;
 import com.novoda.downloadmanager.NotificationCreator;
 import com.novoda.downloadmanager.NotificationInformation;
 import com.novoda.notils.logger.simple.Log;
@@ -13,7 +14,6 @@ public class CustomNotificationCreator implements NotificationCreator {
 
     private static final int ID = 1;
     private static final boolean NOT_INDETERMINATE = false;
-    private static final String CHANNEL = "download_manager";
 
     private final Context context;
     private final int iconDrawable;
@@ -33,12 +33,15 @@ public class CustomNotificationCreator implements NotificationCreator {
 
         Log.v("Create notification for " + title + ", " + content);
 
-        Notification notification = new NotificationCompat.Builder(context, CHANNEL)
+        String notificationChannel = NotificationChannelCreator.createDownloadNotificationChannel(context);
+
+        Notification notification = new NotificationCompat.Builder(context, notificationChannel)
                 .setProgress(bytesFileSize, bytesDownloaded, NOT_INDETERMINATE)
                 .setSmallIcon(iconDrawable)
                 .setContentTitle(title)
                 .setContentText(content)
                 .build();
+
         return new CustomNotificationInformation(ID, notification);
     }
 
--- a/demo-simple/src/main/java/com/novoda/downloadmanager/demo/CustomNotificationCreator.java
+++ b/demo-simple/src/main/java/com/novoda/downloadmanager/demo/CustomNotificationCreator.java
@@ -5,7 +5,7 @@ import android.content.Context;
 import android.support.v4.app.NotificationCompat;
 
 import com.novoda.downloadmanager.DownloadBatchTitle;
-import com.novoda.downloadmanager.NotificationChannelCreator;
+import com.novoda.downloadmanager.DownloadNotificationChannelCreator;
 import com.novoda.downloadmanager.NotificationCreator;
 import com.novoda.downloadmanager.NotificationInformation;
 import com.novoda.notils.logger.simple.Log;
@@ -33,7 +33,7 @@ public class CustomNotificationCreator implements NotificationCreator {
 
         Log.v("Create notification for " + title + ", " + content);
 
-        String notificationChannel = NotificationChannelCreator.createDownloadNotificationChannel(context);
+        String notificationChannel = DownloadNotificationChannelCreator.createDownloadNotificationChannel(context);
 
         Notification notification = new NotificationCompat.Builder(context, notificationChannel)
                 .setProgress(bytesFileSize, bytesDownloaded, NOT_INDETERMINATE)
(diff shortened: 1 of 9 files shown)


library/src/main/java/com/novoda/downloadmanager/NotificationChannelCreator.java
```diff
@@ -0,0 +1,30 @@
+package com.novoda.downloadmanager;
+
+import android.app.NotificationChannel;
+import android.app.NotificationManager;
+import android.content.Context;
+import android.os.Build;
+
+public class NotificationChannelCreator {
```
Mecharyry: @zegnus I've exposed this class so that the `custom` implementations can still use a notification channel without having to essentially duplicate the code here. Can hide behind an interface if you would prefer, that way clients can decide the priority etc of the channel.

library/src/main/java/com/novoda/downloadmanager/NotificationChannelCreator.java
```diff
@@ -0,0 +1,30 @@
...
+
+    public static String createDownloadNotificationChannel(Context context) {
+        if (CHANNEL_ID == null) {
+            CHANNEL_ID = "download-manager";
+            createNotificationChannelForAndroidOreo(context);
```
Mecharyry: Using a `static` to ensure that the setting of the channel is only done once per instance of the application, we always want to supply the basic channel identifier because it is used to create the notification.

library/src/main/java/com/novoda/downloadmanager/DownloadNotificationChannelCreator.java
```diff
@@ -0,0 +1,36 @@
...
+
+class DownloadNotificationChannelCreator implements NotificationChannelCreator {
+
+    private final NotificationManager notificationManager;
+
+    DownloadNotificationChannelCreator(NotificationManager notificationManager) {
+        this.notificationManager = notificationManager;
+    }
+
+    private static String CHANNEL_ID;
```
Mecharyry: Using a `static` to ensure that the `NotificationManager.createNotificationChannel` is called once per instance of the application, rather than every time that a notification is created.

library/src/main/java/com/novoda/downloadmanager/NotificationChannelCreator.java
```diff
@@ -0,0 +1,9 @@
+package com.novoda.downloadmanager;
+
+import android.content.Context;
+
+public interface NotificationChannelCreator {
```
Mecharyry: Created an interface that can be used by the `CustomNotificationBuilder`s.

library/src/main/java/com/novoda/downloadmanager/DownloadBatchNotification.java
```diff
@@ -26,7 +25,9 @@ public NotificationInformation createNotification(DownloadBatchTitle downloadBat
         String title = downloadBatchTitle.asString();
         String content = percentageDownloaded + "% downloaded";
 
-        Notification notification = new NotificationCompat.Builder(context, CHANNEL_ID)
+        String notificationChannel = NotificationChannelCreator.createDownloadNotificationChannel(context);
```
zegnus: I think that `NotificationChannelCreator` should be private. The reason is that this `DownloadBatchNotification` is an implementation of the public `NotificationCreator`.

A client can use just the default notification (and so this default channel), or if they want to create their own notification they should handle it all and decide if they want a channel or not, and with which features (showBadge? lights? etc...)

On the other hand, if you make it public then it belongs to the API contract and we'll have to maintain it and be more strict and careful with the changes, even changes on the class name.

Mecharyry:
@zegnus Sorry that you had to write that long message 😂  When I was writing you the last thing I decided to introduce the interface for the reasons you put 😄

zegnus:
I have a suggestion based on the following; the user of the library should have to do as minimum work as possible, being as trivial as possible. In the current proposal we have a custom implementation (notificationChannelCreator) inside a custom implementation (notificationCreator) and inside the last one it needs to know how to orchestrate the other one. What if we could orchestrate that ourselves?

I've came up with this:

```
public interface NotificationChannelCreator {

    String getNotificationChannelName();

    Optional<NotificationChannel> createNotificationChannel();
}
```

Notice the new first param
```
public interface NotificationCreator {

    NotificationInformation createNotification(String notificationChannelName,
                                               DownloadBatchTitle downloadBatchTitle,
                                               int percentageDownloaded,
                                               int bytesFileSize,
                                               int bytesDownloaded);
}
```

The class `LiteDownloadManagerDownloader` now accepts a `NotificationChannelCreator` as a constructor param and it used method:
```
    private void updateNotification(DownloadBatchStatus liteDownloadBatchStatus, DownloadService downloadService) {
        NotificationInformation notificationInformation = notificationCreator.createNotification(
                notificationChannelCreator.getNotificationChannelName(),
                liteDownloadBatchStatus.getDownloadBatchTitle(),
                liteDownloadBatchStatus.percentageDownloaded(),
                (int) liteDownloadBatchStatus.bytesTotalSize(),
                (int) liteDownloadBatchStatus.bytesDownloaded()
        );
```

Finally, in the class `DownloadManagerBuilder` before `LiteDownloadManagerDownloader downloader = new LiteDownloadManagerDownloader(` we can now create the channel, and it will be done just once so we don't need to hold a static instance of the channel

```
        Optional<NotificationChannel> notificationChannel = notificationChannelCreator.createNotificationChannel();
        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O && notificationChannel.isPresent()) {
            NotificationManager notificationManager = (NotificationManager) context.getSystemService(Context.NOTIFICATION_SERVICE);
            notificationManager.createNotificationChannel(notificationChannel.get());
        }
```

Now, also the `DownloadManagerBuilder` can have an optional `withCustomNotificationChannelCreator(...)`, we have done all the orchestration and the creators now are super simple to use.

library/src/main/java/com/novoda/downloadmanager/DownloadNotificationChannelCreator.java
```diff
@@ -0,0 +1,36 @@
...
+            CHANNEL_ID = "download-manager";
+            createNotificationChannelForAndroidOreo(context);
+        }
+
+        return CHANNEL_ID;
+    }
+
+    private void createNotificationChannelForAndroidOreo(Context context) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            String channelName = "Download Manager Notification Service";
```
takecare: is the plan to have this hardcoded? no resources?

Mecharyry: Good point @takecare, I'll add this to `resources`

library/src/main/java/com/novoda/downloadmanager/Optional.java
```diff
@@ -0,0 +1,62 @@
+package com.novoda.downloadmanager;
+
+public final class Optional<T> {
```
Mecharyry: Had to introduce an `Optional` for some of the recommended changes. Now we can remove `@Nullable` from the codebase.

demo-simple/src/main/java/com/novoda/downloadmanager/demo/CustomNotificationCreator.java
```diff
@@ -4,25 +4,29 @@
...
-    public CustomNotificationCreator(Context context, int iconDrawable) {
+    public CustomNotificationCreator(Context context, int iconDrawable, NotificationChannelCreator notificationChannelCreator) {
         this.context = context;
         this.iconDrawable = iconDrawable;
+        this.notificationChannelCreator = notificationChannelCreator;
     }
 
     @Override
-    public NotificationInformation createNotification(DownloadBatchTitle downloadBatchTitle,
+    public NotificationInformation createNotification(String notificationChannelName, DownloadBatchTitle downloadBatchTitle,
```
joetimmins: Sure we don't want to wrap this `String` in an object? 😅

Mecharyry: I can't tell if you are serious 😆

library/src/main/java/com/novoda/downloadmanager/NotificationChannelCreator.java
```diff
@@ -0,0 +1,11 @@
+package com.novoda.downloadmanager;
+
+import android.app.NotificationChannel;
+
+public interface NotificationChannelCreator {
+
+    Optional<NotificationChannel> createNotificationChannel();
```
joetimmins: If this is gonna be part of the public API of the library can we have some javadoc or something explaining why it's `Optional`?

joetimmins: Sure

Mecharyry: https://novoda.atlassian.net/browse/ALL-4581

demo-simple/src/main/java/com/novoda/downloadmanager/demo/CustomNotificationCreator.java
zegnus: second parameter in a new line ;)

demo-simple/src/main/java/com/novoda/downloadmanager/demo/CustomNotificationCreator.java
```diff
@@ -4,25 +4,29 @@
...
+
     private static final int ID = 1;
     private static final boolean NOT_INDETERMINATE = false;
 
     private final Context context;
     private final int iconDrawable;
+    private final NotificationChannelCreator notificationChannelCreator;
 
-    public CustomNotificationCreator(Context context, int iconDrawable) {
+    public CustomNotificationCreator(Context context, int iconDrawable, NotificationChannelCreator notificationChannelCreator) {
```
zegnus: `NotificationChannelCreator` not needed

Mecharyry: It will be if we want to use this as an example and the custom one not crash 😄

PR #273, opened 2017-12-15
Foreground Migration Service

Mecharyry:
### Problem
We have a `MigrationService` but it does not supply notifications to the user! 😱 

### Solution
Implement a `foreground` Service and forward a `MigrationStatus` that will be mapped to a `Notification` and displayed. Decided to open a PR now because I've made some, potentially controversial changes to the `NotificationCreator` interface 😬  I plan to tidy up how we pass information between the `Service` and the `Migrator` in a follow-up PR, possibly by exposing a `MigrationService` interface that can be used to send notifications on `Migrator` progress updates. This would be similar to how the other `Service` works in the library.   

### Screen Capture
migration

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
+++ b/library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
@@ -1,11 +1,14 @@
 package com.novoda.downloadmanager;
 
 import android.app.AlarmManager;
+import android.app.NotificationChannel;
+import android.app.NotificationManager;
 import android.app.PendingIntent;
 import android.app.Service;
 import android.content.Context;
 import android.content.Intent;
 import android.os.Binder;
+import android.os.Build;
 import android.os.IBinder;
 import android.os.SystemClock;
 import android.support.annotation.Nullable;
@@ -22,6 +25,9 @@ public class LiteDownloadMigrationService extends Service {
     private IBinder binder;
     private Migrator.Callback migrationCallback;
     private String updateMessage;
+    private NotificationCreator<MigrationStatus> notificationCreator;
+    private NotificationChannelCreator channelCreator;
+    private NotificationManager notificationManager;
 
     @Override
     public int onStartCommand(Intent intent, int flags, int startId) {
@@ -52,6 +58,15 @@ public class LiteDownloadMigrationService extends Service {
         }
 
         binder = new MigrationDownloadServiceBinder();
+        channelCreator = new MigrationNotificationChannelCreator(getResources());
+        notificationManager = (NotificationManager) getSystemService(Context.NOTIFICATION_SERVICE);
+
+        Optional<NotificationChannel> notificationChannel = channelCreator.createNotificationChannel();
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O && notificationChannel.isPresent()) {
+            notificationManager.createNotificationChannel(notificationChannel.get());
+        }
+
+        notificationCreator = new MigrationNotification(getApplicationContext(), android.R.drawable.ic_dialog_alert);
 
         super.onCreate();
     }
@@ -62,10 +77,22 @@ public class LiteDownloadMigrationService extends Service {
             public void onUpdate(String message) {
                 updateMessage = message;
                 if (migrationCallback != null) {
+                    MigrationStatus migrationStatus = new VersionOneToVersionTwoMigrationStatus(message);
+                    updateNotification(migrationStatus);
                     migrationCallback.onUpdate(message);
                 }
             }
         };
+    }
+
+    private void updateNotification(MigrationStatus migrationStatus) {
+        NotificationInformation notification = notificationCreator.createNotification(channelCreator.getNotificationChannelName(), migrationStatus);
+        startForeground(notification.getId(), notification.getNotification());
+
+        if (migrationStatus.message().equals("Migration Complete")) {
+            stopForeground(true);
+            notificationManager.notify(notification.getId(), notification.getNotification());
+        }
     }
 
     @Nullable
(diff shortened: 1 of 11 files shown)


library/src/main/java/com/novoda/downloadmanager/NotificationCreator.java
```diff
@@ -1,10 +1,7 @@
 package com.novoda.downloadmanager;
 
-public interface NotificationCreator {
+public interface NotificationCreator<PAYLOAD> {
```
Mecharyry: We need a `NotificationCreator` for **Migration** and **Downloads**, the client applications should decide how these are displayed. To be able to reuse this interface I have introduced the generic `PAYLOAD` which represents the data need to construct the `Notification` under a specific scenario.

Internally we use `DownloadBatchStatus` for standard downloads, so we type to that. We'll do the same with the migration when we pass that through the builders.

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrationStatus.java
```diff
@@ -0,0 +1,15 @@
+package com.novoda.downloadmanager;
+
+class VersionOneToVersionTwoMigrationStatus implements MigrationStatus {
```
Mecharyry: Just a simple wrapper for now, this will be built upon in a future PR when we decide what we want to surface to the user for the migration notifications.

library/src/main/java/com/novoda/downloadmanager/MigrationNotificationChannelCreator.java
```diff
@@ -0,0 +1,39 @@
+package com.novoda.downloadmanager;
+
+import android.app.NotificationChannel;
+import android.app.NotificationManager;
+import android.content.res.Resources;
+import android.os.Build;
+
+class MigrationNotificationChannelCreator implements NotificationChannelCreator {
```
Mecharyry: A Migration will likely require a separate channel, some users / client applications may not want to surface notifications to users.

zegnus:
Closing, looks like we need to support `all4:///`

library/src/main/java/com/novoda/downloadmanager/NotificationCreator.java
joetimmins: This is a good idea, but I'd go with the more usual `T` instead of `PAYLOAD`

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
```diff
@@ -52,6 +58,15 @@ public void onCreate() {
         }
 
         binder = new MigrationDownloadServiceBinder();
+        channelCreator = new MigrationNotificationChannelCreator(getResources());
```
zegnus: Neither the `MigrationNotificationChannelCreator ` or the `MigrationNotification` can be injected by the client, that means that the client has no choice but to show in the way encoded here the migrations notifications right? If we allow the client to specify its own, then this library just needs to provide a basic good-enough implementation.

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
```diff
@@ -62,12 +77,24 @@ public void onCreate() {
...
                 }
             }
         };
     }
 
+    private void updateNotification(MigrationStatus migrationStatus) {
+        NotificationInformation notification = notificationCreator.createNotification(channelCreator.getNotificationChannelName(), migrationStatus);
+        startForeground(notification.getId(), notification.getNotification());
+
+        if (migrationStatus.message().equals("Migration Complete")) {
```
zegnus: Can we transform the strings on this migration status object by an enum `Status`?

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
```diff
@@ -110,6 +137,7 @@ MigrationDownloadServiceBinder withCallback(Migrator.Callback migrationCallback)
 
         void bind() {
             LiteDownloadMigrationService.this.migrationCallback.onUpdate(LiteDownloadMigrationService.this.updateMessage);
```
zegnus: Do you want to align this statics with the way that is done in the download manager service?

library/src/main/java/com/novoda/downloadmanager/NotificationCreator.java
Mecharyry: Yea I figured we wouldn't like this 👅 will change it.

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
Mecharyry: I will be adding a way for the client to inject this, sorry I should have mentioned. This will be done in a follow-up PR to keep things small and contained 😄

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
Mecharyry: I've made an enum `Status` on the follow-up branch to this one. I just wanted to start introducing the foreground service in this PR, following it up with improvements to the status and notifications 😄

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
Mecharyry: I think I have also done this in the follow-up branch 😄 I will ensure that this `Service` behaves, as much as possible, like the other Service in the library 😄

library/src/main/java/com/novoda/downloadmanager/NotificationCreator.java
Mecharyry: 85b736f

PR #274, opened 2017-12-19
Avoid leaking ServiceConnection

Mecharyry:
### Problem
If a client uses any `Context` except the `Application Context` then the app will likely crash because we are leaking the `ServiceConnection`. This occurs because currently the `DLM` does not unbind it 😬 

### Solution
Ensure that we use only the `Application Context` by calling `context.getApplicationContext` inside the `DLM`.

### Additional
We should decide how we are going to surface the need to `bind` and `unbind` to the client applications.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
@@ -66,7 +66,6 @@ public final class DownloadManagerBuilder {
         FileSizeRequester fileSizeRequester = new NetworkFileSizeRequester(httpClient, requestCreator);
         FileDownloader fileDownloader = new NetworkFileDownloader(httpClient, requestCreator);
 
-        NotificationManager notificationManager = (NotificationManager) applicationContext.getSystemService(Context.NOTIFICATION_SERVICE);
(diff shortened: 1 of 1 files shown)


library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
```diff
@@ -223,7 +223,7 @@ public void onServiceDisconnected(ComponentName name) {
 
         Optional<NotificationChannel> notificationChannel = notificationChannelCreator.createNotificationChannel();
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O && notificationChannel.isPresent()) {
```
joetimmins: I guess it's not really for this pull request, but it looks like we're doing this check twice - once inside the `createNotificationChannel` method and once here.

Mecharyry: We need this check twice 😢 the one inside `createNotificationChannel` can be defined by the client, not by the library, so we need this safety 😬

joetimmins: I don't think that means we need the check twice. I'm going to do some digging / refactoring / exploration

PR #275, opened 2017-12-19
MigrationStatus progress notifications

Mecharyry:
### Problem
Created a foreground migration service in #273 but the notifications do not sure progress updates etc. 

### Solution
Create a `MigrationStatus` that encapsulates all of the information necessary to be able to create notifications with progress updates.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/InternalMigrationStatus.java
+++ b/library/src/main/java/com/novoda/downloadmanager/InternalMigrationStatus.java
@@ -1,0 +1,15 @@
+package com.novoda.downloadmanager;
+
+interface InternalMigrationStatus extends MigrationStatus {
+
+    void update(int currentBatch, int numberOfBatches);
+
+    void markAsExtracting();
+
+    void markAsMigrating();
+
+    void markAsDeleting();
+
+    void markAsComplete();
+
+}
--- a/library/src/main/java/com/novoda/downloadmanager/MigrationStatus.java
+++ b/library/src/main/java/com/novoda/downloadmanager/MigrationStatus.java
@@ -2,6 +2,21 @@ package com.novoda.downloadmanager;
 
 interface MigrationStatus {
 
-    String message();
+    enum Status {
+
+        EXTRACTING,
+        MIGRATING_FILES,
+        DELETING_V1_DATABASE,
+        COMPLETE;
+
+        public String toRawValue() {
+            return this.name();
+        }
+
+    }
+
+    int percentageMigrated();
+
+    Status status();
 
 }
--- a/library/src/main/java/com/novoda/downloadmanager/MigrationStatus.java
+++ b/library/src/main/java/com/novoda/downloadmanager/MigrationStatus.java
@@ -1,6 +1,6 @@
 package com.novoda.downloadmanager;
 
-interface MigrationStatus {
+public interface MigrationStatus {
 
     enum Status {
 
@@ -14,6 +14,10 @@ interface MigrationStatus {
         }
 
     }
+
+    int numberOfBatches();
+
+    int totalNumberOfBatches();
 
     int percentageMigrated();
 
--- a/library/src/main/java/com/novoda/downloadmanager/MigrationStatus.java
+++ b/library/src/main/java/com/novoda/downloadmanager/MigrationStatus.java
@@ -15,9 +15,9 @@ public interface MigrationStatus {
 
     }
 
-    int numberOfBatches();
+    int numberOfMigratedBatches();
 
-    int totalNumberOfBatches();
+    int totalNumberOfBatchesToMigrate();
 
     int percentageMigrated();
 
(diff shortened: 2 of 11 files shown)


library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
```diff
@@ -65,12 +81,28 @@ public void migrate() {
...
+        notifyOfProgress(internalMigrationStatus);
+    }
+
+    private void notifyOfProgress(MigrationStatus migrationStatus) {
+        String channelName = channelCreator.getNotificationChannelName();
+        NotificationInformation notification = notificationCreator.createNotification(channelName, migrationStatus);
+
+        migrationService.updateMessage(migrationStatus);
+
+        if (migrationStatus.status() == MigrationStatus.Status.COMPLETE) {
```
Mecharyry: I've been torn as to whether just pass generic `Notification` back to the `activity` or to pass back `MigrationStatus` and let the `Service` decide which type of notification to use based on that. What do you think?

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
```diff
@@ -21,35 +21,51 @@
...
+        internalMigrationStatus.markAsExtracting();
+        notifyOfProgress(internalMigrationStatus);
+
         Log.d(TAG, "about to extract migrations, time is " + System.nanoTime());
         List<Migration> migrations = migrationExtractor.extractMigrations();
         Log.d(TAG, "migrations are all EXTRACTED, time is " + System.nanoTime());
 
-        migrationCallback.onUpdate("Migrating Files");
+        internalMigrationStatus.markAsMigrating();
+        notifyOfProgress(internalMigrationStatus);
```
Mecharyry: I'm not sure if `marking` and then `notifying` adds to much noise in this class but it is super explicit that we are mutating an object (to prevent unnecessary memory churn) and then forwarding that for the `Service`.

library/src/main/java/com/novoda/downloadmanager/InternalMigrationStatus.java
```diff
@@ -0,0 +1,15 @@
+package com.novoda.downloadmanager;
+
+interface InternalMigrationStatus extends MigrationStatus {
```
zegnus: Are you 100% you need an internal interface with one implementation?

library/src/main/java/com/novoda/downloadmanager/MigrationStatus.java
```diff
@@ -1,7 +1,26 @@
...
+        DELETING_V1_DATABASE,
+        COMPLETE;
+
+        public String toRawValue() {
+            return this.name();
+        }
+
+    }
+
+    int numberOfBatches();
```
zegnus: is this the number of migrated batches?

library/src/main/java/com/novoda/downloadmanager/MigrationStatus.java
```diff
@@ -1,7 +1,26 @@
...
+
+    int totalNumberOfBatches();
```
zegnus: is this the absolute number of batches that needs to be migrated?

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
zegnus: I might be torn to use the most obvious solution for the client (Activity)

library/src/main/java/com/novoda/downloadmanager/InternalMigrationStatus.java
Mecharyry: Yes, I can't imagine us requiring another.

library/src/main/java/com/novoda/downloadmanager/MigrationStatus.java
Mecharyry: Yes 😄

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
Mecharyry: I don't understand your comment @zegnus 😄

library/src/main/java/com/novoda/downloadmanager/MigrationStatus.java
Mecharyry: e2b956c1a199d8766e391b136a63c9b265ff03ba

library/src/main/java/com/novoda/downloadmanager/MigrationStatus.java
Mecharyry: e2b956c1a199d8766e391b136a63c9b265ff03ba

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
Mecharyry: I think I put `Activity` in the comment by accident earlier 😄 the `Service` is the one that creates and displays the notification. We have an optional callback for the activity to receive `NotificationStatus` and act on that.

PR #276, opened 2017-12-20
Migration service binder

Mecharyry:
### Problem
The `DownloadMigrationService` is capable of restarting a migration even if terminated abruptly when it is rebound. This means we can remove the code associated with `onStartCommand` etc, this `Service` does not need to restart itself when low on resources, we can continue once again when a client rebinds.

### Solution
Remove `onStartCommand` and `onTaskRemoved`. Have the `binder` return a reference to the `DownloadMigrationService` interface which we can use to set the dependencies, this will bring it in line with the other `Service` in the library.


library/src/main/java/com/novoda/downloadmanager/DownloadMigrationService.java
```diff
@@ -0,0 +1,12 @@
+package com.novoda.downloadmanager;
+
+interface DownloadMigrationService {
```
Mecharyry: Interface that we expose rather than the `android` version of `Service`.

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
```diff
@@ -81,65 +35,91 @@ public void onCreate() {
...
+        this.notificationCreator = notificationCreator;
     }
 
-    class MigrationDownloadServiceBinder extends Binder {
-
-        MigrationDownloadServiceBinder withCallback(Migrator.Callback migrationCallback) {
-            LiteDownloadMigrationService.this.migrationCallback = migrationCallback;
-            return this;
+    @Override
+    public void startMigration() {
```
Mecharyry: Mainly reordering below this point.

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
```diff
@@ -81,65 +35,91 @@ public void onCreate() {
...
+    }
+
+    @TargetApi(Build.VERSION_CODES.O)
+    private boolean notificationChannelDoesNotExist(String channelName) {
+        return notificationManager.getNotificationChannel(channelName) == null;
+    }
+
+    private final Migrator.Callback migrationStatusCallback = new Migrator.Callback() {
+        @Override
+        public void onUpdate(MigrationStatus migrationStatus) {
```
Mecharyry: We don't pass the `Service` to the `Migrator` anymore. Instead we pass a callback. When updates occur in the `Migrator` we callback here with just the `MigrationStatus` and the `Service` decides how and what to display.

library/src/main/java/com/novoda/downloadmanager/MigrationFactory.java
```diff
@@ -2,39 +2,18 @@
 
 import android.content.Context;
 import android.database.sqlite.SQLiteDatabase;
-import android.os.Handler;
-import android.os.Looper;
 
 import java.io.File;
 
-public final class MigrationFactory {
+final class MigrationFactory {
```
Mecharyry: This shouldn't need to be exposed to clients of the library...UNLESS we are planning on providing different `Migrators` which is unlikely.

library/src/main/java/com/novoda/downloadmanager/MigrationServiceBinder.java
```diff
@@ -8,31 +8,42 @@
...
 
-    private final Context context;
-    private final Migrator.Callback migrationCallback;
+    private final Context applicationContext;
+    private final Callback migrationCallback;
+    private final NotificationChannelCreator notificationChannelCreator;
+    private final NotificationCreator<MigrationStatus> notificationCreator;
 
-    private MigrationServiceConnection serviceConnection;
+    public interface Callback {
```
Mecharyry: ❓ QUESTION ❓ 

Had to expose a `Callback` that the clients use to receive update events from the `Migrator`. Could do this another way by exposing the `Migrator` to the client. Or we could expose just the listener. I was unsure, what do you think?

library/src/main/java/com/novoda/downloadmanager/MigrationServiceBinder.java
```diff
@@ -8,31 +8,42 @@
 
 public class MigrationServiceBinder {
 
-    private final Context context;
-    private final Migrator.Callback migrationCallback;
+    private final Context applicationContext;
```
Mecharyry: ❓ QUESTION ❓ 

Anyone know if there is an annotation to enforce `ApplicationContext`?

library/src/main/java/com/novoda/downloadmanager/MigrationServiceBinderBuilder.java
```diff
@@ -0,0 +1,70 @@
+package com.novoda.downloadmanager;
+
+import android.content.Context;
+import android.os.Handler;
+
+public class MigrationServiceBinderBuilder {
```
Mecharyry: Provides defaults for the `Binder` and wraps the callback on a main thread handler, or any handler that the client specifies.

library/src/main/java/com/novoda/downloadmanager/Migrator.java
```diff
@@ -2,7 +2,7 @@
 
 import android.support.annotation.WorkerThread;
 
-public interface Migrator {
+interface Migrator {
```
Mecharyry: I'm under the impression that this should be an internal concept, clients should interact through the binder only.

library/src/main/java/com/novoda/downloadmanager/VersionOneToVersionTwoMigrator.java
```diff
@@ -81,28 +75,14 @@ public void migrate() {
...
+        migratorUpdateCallback.onUpdate(migrationStatus);
         Log.d(TAG, "all traces of v1 are ERASED, time is " + System.nanoTime());
         database.close();
 
         database.deleteDatabase();
-        internalMigrationStatus.markAsComplete();
-        notifyOfProgress(internalMigrationStatus);
-    }
-
-    private void notifyOfProgress(MigrationStatus migrationStatus) {
```
Mecharyry: I thought this just added noise to this class and the `Service` or some other object can be responsible for deciding how to display a notification based on a status.

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
```diff
@@ -81,65 +35,91 @@ public void onCreate() {
...
-        MigrationDownloadServiceBinder withCallback(Migrator.Callback migrationCallback) {
-            LiteDownloadMigrationService.this.migrationCallback = migrationCallback;
-            return this;
+    @Override
+    public void startMigration() {
+        Optional<NotificationChannel> notificationChannel = notificationChannelCreator.createNotificationChannel();
+        String channelName = notificationChannelCreator.getNotificationChannelName();
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O && notificationChannel.isPresent() && notificationChannelDoesNotExist(channelName)) {
+            notificationManager.createNotificationChannel(notificationChannel.get());
         }
```
pablisco: For clarity I think we can move the version and channel existence to a separate method. Feels like these 3 lines are doing too much.
```
if(notificationChannel.isPresent()) {
  createNotificationChannel(notificationChannel.get())
}
```
It could even be part of a "compat" utils module/library

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
```diff
@@ -81,65 +35,91 @@ public void onCreate() {
...
 
-        void bind() {
-            if (migrationStatus != null && migrationCallback != null) {
-                LiteDownloadMigrationService.this.migrationCallback.onUpdate(migrationStatus);
+        executor.execute(new Runnable() {
```
pablisco: If we use an `IntentService` there is no need to use an `Executor` right? I'm not 100% sure

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
```diff
@@ -81,65 +35,91 @@ public void onCreate() {
...
+        notificationChannelCreator = new MigrationNotificationChannelCreator(getResources());
         notificationCreator = new MigrationNotification(getApplicationContext(), android.R.drawable.ic_dialog_alert);
+        notificationManager = (NotificationManager) getSystemService(Context.NOTIFICATION_SERVICE);
 
         super.onCreate();
     }
 
+    class MigrationDownloadServiceBinder extends Binder {
+        DownloadMigrationService getService() {
+            return LiteDownloadMigrationService.this;
```
zegnus: 💯

library/src/main/java/com/novoda/downloadmanager/MigrationServiceBinder.java
```diff
@@ -41,9 +52,11 @@ public void unbind() {
 
         @Override
         public void onServiceConnected(ComponentName componentName, IBinder iBinder) {
-            ((LiteDownloadMigrationService.MigrationDownloadServiceBinder) iBinder)
-                    .withCallback(migrationCallback)
-                    .bind();
+            DownloadMigrationService migrationService = ((LiteDownloadMigrationService.MigrationDownloadServiceBinder) iBinder).getService();
```
zegnus: 💯 yeah, now is super clean

library/src/main/java/com/novoda/downloadmanager/MigrationServiceBinderBuilder.java
```diff
@@ -0,0 +1,70 @@
...
+
+    private final Context applicationContext;
+    private final Handler callbackHandler;
+    private MigrationServiceBinder.Callback migrationCallback;
+    private NotificationChannelCreator notificationChannelCreator;
+    private NotificationCreator<MigrationStatus> notificationCreator;
+
+    public static MigrationServiceBinderBuilder newInstance(Context context, Handler callbackHandler) {
+        Context applicationContext = context.getApplicationContext();
+        MigrationServiceBinder.Callback callback = new MigrationServiceBinder.Callback() {
```
zegnus: Do you need a static `NO_OP`?

library/src/main/java/com/novoda/downloadmanager/MigrationServiceBinder.java
pablisco: The only defensive strategy is to eagerly get the application context from the provided context. If the provided context is already an application context it just returns itself

library/src/main/java/com/novoda/downloadmanager/MigrationServiceBinder.java
pablisco: Maybe if it's a top level class called `MigrationCallback` or spoilers ahead for my review... using `Promises`

library/src/main/java/com/novoda/downloadmanager/MigrationServiceBinderBuilder.java
```diff
@@ -0,0 +1,70 @@
...
+                    @Override
+                    public void run() {
+                        migrationCallback.onUpdate(migrationStatus);
+                    }
+                });
+            }
+        };
+
+        return new MigrationServiceBinder(applicationContext, handlerReportingMigrationCallback, notificationChannelCreator, notificationCreator);
+    }
```
pablisco: With the suggestions I made earlier there may not be a need for this builder ❓

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -17,10 +18,9 @@
 import com.novoda.downloadmanager.LiteDownloadManagerCommands;
 import com.novoda.downloadmanager.LocalFilesDirectory;
 import com.novoda.downloadmanager.LocalFilesDirectoryFactory;
-import com.novoda.downloadmanager.MigrationFactory;
 import com.novoda.downloadmanager.MigrationServiceBinder;
```
zegnus: Can this be an interface instead? Also, what if we just rename these classes dropping the `binder` suffix?

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -46,7 +46,9 @@ protected void onCreate(Bundle savedInstanceState) {
 
         Log.setShowLogs(true);
 
-        migrationServiceBinder = MigrationFactory.migrationServiceBinder(this, migrationCallback);
+        migrationServiceBinder = MigrationServiceBinderBuilder.newInstance(getApplicationContext(), new Handler(getMainLooper()))
```
zegnus: Just `MigrationServiceBuilder`?

library/src/main/java/com/novoda/downloadmanager/MigrationServiceBinder.java
zegnus: No as far as I know, as is a `Context` also

zegnus:
By the way, in `DownloadManagerBuilder.build()` can you please just return `LiteDownloadManagerCommands` ? :)

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
joetimmins: Suggested the same rename yesterday 😅

joetimmins: Not even any real need for `Service` to be in there either. How about `DownloadMigrator`?

library/src/main/java/com/novoda/downloadmanager/DownloadMigrationService.java
```diff
@@ -0,0 +1,12 @@
...
+
+    void setMigrationCallback(MigrationServiceBinder.Callback migrationCallback);
```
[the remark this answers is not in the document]
joetimmins: We're not using Java 8 syntax in this lib, can we still do a `Promise`?

library/src/main/java/com/novoda/downloadmanager/MigrationFactory.java
joetimmins: Does that mean we can remove the `Migrator` interface? No real need for it if there's only ever going to be one implementation of it

pablisco:
@joetimmins We pair on it, I think it simplifies it a bit :) `Promise` is just another name for an observable future.

library/src/main/java/com/novoda/downloadmanager/DownloadMigrationService.java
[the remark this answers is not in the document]
Mecharyry: We can still apply the basic pattern behind promises using interfaces

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -46,7 +47,7 @@ protected void onCreate(Bundle savedInstanceState) {
 
         Log.setShowLogs(true);
 
-        migrationServiceBinder = MigrationFactory.migrationServiceBinder(this, migrationCallback);
+        downloadMigrationService = new MigrationService(this);
```
Mecharyry: No longer passing `Notification` specific classes as constructor params, these are now given through the `startMigration` method.

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -72,15 +73,8 @@ public void onClick(View view) {
...
-                migrationServiceBinder.bind();
-            }
-        });
-
-        View buttonAbortMigration = findViewById(R.id.button_abort_migration);
-        buttonAbortMigration.setOnClickListener(new View.OnClickListener() {
-            @Override
-            public void onClick(View view) {
-                migrationServiceBinder.unbind();
+                downloadMigrationService.startMigration()
```
Mecharyry: We have an override to which a client can pass `Notification` customisations to `startMigration`.

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -72,15 +73,8 @@ public void onClick(View view) {
...
+                        .observe(migrationCallback);
```
Mecharyry: The clients specifically `observe` the changes on the `MigrationService` now.

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -137,17 +131,23 @@ public void onReceived(List<DownloadBatchStatus> downloadBatchStatuses) {
...
             databaseCloningUpdates.setText(updateMessage);
         }
     };
 
-    Migrator.Callback migrationCallback = new Migrator.Callback() {
+    private final DownloadMigrationService.MigrationCallback migrationCallback = new DownloadMigrationService.MigrationCallback() {
         @Override
-        public void onUpdate(MigrationStatus migrationStatus) {
-            databaseMigrationUpdates.setText(migrationStatus.status().toRawValue());
+        public void onUpdate(final MigrationStatus migrationStatus) {
```
Mecharyry: `onUpdate` is marked with `@WorkerThread` to highlight to clients that they need to handle the thread swap should they want to display anything when an update is received.

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
pablisco: NIT: overload 🙃

demo-simple/src/main/res/layout/activity_main.xml
```diff
@@ -49,12 +49,6 @@
       android:layout_height="wrap_content"
       android:text="Migrate" />
 
-    <Button
```
Mecharyry: Removing this from the UI we are implicitly unbinding now. If wanting to test this can be done via `adb` or through the device directly by terminating a service.

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
Mecharyry: I have no idea what you are talking about :trollface:

joetimmins: I *really* have no idea what you guys are talking about.

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
zegnus: ❌ the updates should be received in the thread where the tool is being used, same as in the downloadmanager service callbacks

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -137,17 +131,23 @@ public void onReceived(List<DownloadBatchStatus> downloadBatchStatuses) {
         bindViews();
     }
 
-    VersionOneDatabaseCloner.CloneCallback cloneCallback = new VersionOneDatabaseCloner.CloneCallback() {
+    private final VersionOneDatabaseCloner.CloneCallback cloneCallback = new VersionOneDatabaseCloner.CloneCallback() {
```
zegnus: What's this callback for? Isn't the migrator callback enough?

library/src/main/java/com/novoda/downloadmanager/DownloadMigrationService.java
```diff
@@ -0,0 +1,27 @@
+package com.novoda.downloadmanager;
+
+import android.support.annotation.NonNull;
+import android.support.annotation.WorkerThread;
+
+public interface DownloadMigrationService {
+
+    MigrationFuture startMigration(NotificationChannelCreator notificationChannelCreator, NotificationCreator<MigrationStatus> notificationCreator);
```
zegnus: The builder allowed to optionally inject custom implementations, now you are forcing the user to pass both when he might actually just want to use a custom notificationChannelCreator, not both. I'm not sure this change is for better.

library/src/main/java/com/novoda/downloadmanager/DownloadMigrationService.java
```diff
@@ -0,0 +1,27 @@
...
+
+    interface MigrationFuture {
```
zegnus: You can remove this interface if you just pass the callback in the `migrate` method as it was before. Also, the `.observe()` is mandatory now because the execution now is delayed. What's the advantage from the previous implementation?

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
joetimmins: This is for some different functionality - the demo app has the ability to create a v1 database for testing purposes, this relates to that

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
joetimmins: I agree with @zegnus - the clients of the library shouldn't need to care about managing threads like this.

library/src/main/java/com/novoda/downloadmanager/ManagedDownloadMigrationService.java
```diff
@@ -0,0 +1,5 @@
+package com.novoda.downloadmanager;
+
+public interface ManagedDownloadMigrationService extends DownloadMigrationService {
```
joetimmins: I don't think this should be a separate interface - we need to keep the public API as small as possible.

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
```diff
@@ -1,76 +1,69 @@
...
 
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.util.List;
 
-import static com.novoda.downloadmanager.DownloadBatchStatus.Status;
-
-class VersionOneToVersionTwoMigrator implements Migrator {
+class MigrationJob extends MigrationFutureWithCallbacks implements Runnable, DownloadMigrationService.MigrationFuture {
```
joetimmins: I would avoid using inheritance wherever possible. You can use composition to achieve the same result here quite easily - I think if you have a `MigrationFutureWithCallbacks` as a field on this class you can call `onUpdate()` on it instead of using the inherited implementation.

pablisco: It's a mixin. I agree that composition is better. We could easily have it as a delegate instead.

library/src/main/java/com/novoda/downloadmanager/ManagedDownloadMigrationService.java
pablisco: I thought this was remuved @Mecharyry

library/src/main/java/com/novoda/downloadmanager/DownloadMigrationService.java
Mecharyry: This is only used by the library. The clients interact with the `DownloadMigrator` (not an interface).

library/src/main/java/com/novoda/downloadmanager/DownloadMigrator.java
```diff
@@ -0,0 +1,60 @@
+package com.novoda.downloadmanager;
+
+import android.content.ComponentName;
+import android.content.Context;
+import android.content.Intent;
+import android.content.ServiceConnection;
+import android.os.Handler;
+import android.os.IBinder;
+
+public class DownloadMigrator {
```
Mecharyry: The clients only interact with this class and the associated `builder` now.

library/src/main/java/com/novoda/downloadmanager/DownloadMigrator.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public void startMigration(final MigrationCallback migrationCallback) {
+        ServiceConnection serviceConnection = new ServiceConnection() {
+            @Override
+            public void onServiceConnected(ComponentName name, IBinder binder) {
+                DownloadMigrationService migrationService = ((LiteDownloadMigrationService.MigrationDownloadServiceBinder) binder).getService();
+                migrationService.setNotificationChannelCreator(notificationChannelCreator);
+                migrationService.setNotificationCreator(notificationCreator);
+
+                MigrationCallback mainThreadReportingMigrationCallback = new MigrationCallback() {
```
Mecharyry: We make sure that the callbacks occur on the `main thread` again, rather than leaving it to the client.

library/src/main/java/com/novoda/downloadmanager/DownloadMigratorBuilder.java
```diff
@@ -0,0 +1,45 @@
+package com.novoda.downloadmanager;
+
+import android.content.Context;
+import android.os.Handler;
+import android.os.Looper;
+
+public class DownloadMigratorBuilder {
```
Mecharyry: This is one of the two classes that the client now interacts with. `DownloadMigrator` being the other.

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
```diff
@@ -1,97 +1,78 @@
...
-    public void onCreate() {
-        Log.d(TAG, "onCreate");
-        if (executor == null) {
-            executor = Executors.newSingleThreadExecutor();
-        }
+    public void startMigration(MigrationCallback migrationCallback) {
+        createNotificationChannel();
+        MigrationJob migrationJob = new MigrationJob(getApplicationContext(), getDatabasePath("downloads.db"));
+        migrationJob.addCallback(migrationCallback);
+        migrationJob.addCallback(notificationMigrationCallback);
```
Mecharyry: `MigrationJob` takes several `callbacks` so that we can easily distinguish what comes from the client and what we do internally. I felt this was better than wrapping them unnecessarily.

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
```diff
@@ -1,76 +1,75 @@
...
+        this.databasePath = databasePath;
+    }
+
+    void addCallback(MigrationCallback callback) {
+        migrationCallbacks.add(callback);
     }
 
-    @Override
-    public void migrate() {
+    public void run() {
```
Mecharyry: Each job is now self contained and creates all of it's dependencies when running. We now return a `DB_NOT_PRESENT` if the file doesn't exist rather than just terminating 😬

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -137,17 +131,23 @@ public void onReceived(List<DownloadBatchStatus> downloadBatchStatuses) {
...
         }
     };
 
-    Migrator.Callback migrationCallback = new Migrator.Callback() {
+    private final MigrationCallback migrationCallback = new MigrationCallback() {
         @Override
-        public void onUpdate(MigrationStatus migrationStatus) {
-            databaseMigrationUpdates.setText(migrationStatus.status().toRawValue());
+        public void onUpdate(final MigrationStatus migrationStatus) {
+            Handler handler = new Handler(getMainLooper());
```
joetimmins: Do we still need this? We wrap the callback in a main thread handler in `DownloadMigrator`

Mecharyry: Oh snap 😂

Mecharyry: You saw nothing 😂 b354991

library/src/main/java/com/novoda/downloadmanager/DownloadMigrationService.java
joetimmins: Is there any need for it if it's not gonna be public? We already know that `LiteDownloadMigrationService` implements this. Every time it's used we downcast a `binder.getService()` result to LiteDownloadMigrationService` and then upcast that to `DownloadManagerService`. Feels a bit pointless. What do you think @zegnus?

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -137,16 +130,16 @@ public void onReceived(List<DownloadBatchStatus> downloadBatchStatuses) {
...
         public void onUpdate(String updateMessage) {
             databaseCloningUpdates.setText(updateMessage);
         }
     };
 
-    Migrator.Callback migrationCallback = new Migrator.Callback() {
+    private final MigrationCallback migrationCallback = new MigrationCallback() {
         @Override
-        public void onUpdate(MigrationStatus migrationStatus) {
+        public void onUpdate(final MigrationStatus migrationStatus) {
```
zegnus: No need for `final`

library/src/main/java/com/novoda/downloadmanager/DownloadMigrator.java
zegnus: Shall we expose then a public interface? ;)

Mecharyry:
@zegnus I wouldn't introduce another interface, I wouldn't expect us to introduce another implementation and would be worried about introducing too many.

PR #277, opened 2017-12-21
Reduce public API around custom notifications as much as possible

joetimmins:
## Problem
The public API presented by the library for customising notifications was large and confusing. It also needed the same string to be used in three different places - twice on the `NotificationChannelCreator` interface and once on the `NotificationCreator` interface - otherwise notifications would not appear at all.

## Solution
Reduce the amount of public facing classes and interfaces needed to customise notifications, by creating a new `NotificationConfig` class and a `NotificationCustomiser` interface for the client to supply an implementation for. We then used the new classes and interfaces internally to provide default behaviour.

### Paired with 
@Frankie-Boy


demo-simple/src/main/java/com/novoda/downloadmanager/demo/CustomNotificationConfigFactory.java
```diff
@@ -0,0 +1,50 @@
...
+import android.support.v4.app.NotificationCompat;
+import android.support.v4.app.NotificationManagerCompat;
+
+import com.novoda.downloadmanager.DownloadBatchStatus;
+import com.novoda.downloadmanager.DownloadBatchTitle;
+import com.novoda.downloadmanager.NotificationConfig;
+import com.novoda.downloadmanager.NotificationCustomiser;
+import com.novoda.notils.logger.simple.Log;
+
+public final class CustomNotificationConfigFactory {
```
joetimmins: This class is unused, it replaces the functionality from the now-deleted `CustomNotificationCreator`.

library/src/main/java/com/novoda/downloadmanager/DownloadBatchNotification.java
```diff
@@ -1,58 +0,0 @@
...
-    private final int iconDrawable;
-
-    DownloadBatchNotification(Context context, @DrawableRes int iconDrawable) {
-        this.context = context.getApplicationContext();
-        this.iconDrawable = iconDrawable;
-    }
-
-    @Override
-    public NotificationInformation createNotification(String notificationChannelName, DownloadBatchStatus downloadBatchStatus) {
-        DownloadBatchTitle downloadBatchTitle = downloadBatchStatus.getDownloadBatchTitle();
```
joetimmins: This code is now in an anonymous class inside `DownloadManagerBuilder` at 🏈

library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
```diff
@@ -66,9 +68,35 @@ public static DownloadManagerBuilder newInstance(Context context, Handler callba
...
 
-        NotificationChannelCreator notificationChannelCreator = new DownloadNotificationChannelCreator(context.getResources());
+        NotificationCustomiser<DownloadBatchStatus> notificationCustomiser = new NotificationCustomiser<DownloadBatchStatus>() {
 
-        NotificationCreator<DownloadBatchStatus> downloadBatchNotification = new DownloadBatchNotification(context, notificationIcon);
+            private static final boolean NOT_INDETERMINATE = false;
+
+            @Override
+            public Notification customNotificationFrom(NotificationCompat.Builder builder, DownloadBatchStatus payload) {
+                DownloadBatchTitle downloadBatchTitle = payload.getDownloadBatchTitle();
```
joetimmins: 🏈

Happy to move this wherever we think is best.

library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
```diff
@@ -234,8 +259,8 @@ public void onServiceDisconnected(ComponentName name) {
                 fileOperations,
                 downloadsBatchPersistence,
                 downloadsFilePersistence,
-                notificationChannelCreator,
-                notificationCreator,
+                notificationConfig,
+                notificationConfig,
```
joetimmins: `NotificationConfig` implements both of the old interfaces so can be passed like this. Reason for doing this is just to keep the pull request size a bit smaller.

library/src/main/java/com/novoda/downloadmanager/NotificationConfig.java
```diff
@@ -0,0 +1,53 @@
+package com.novoda.downloadmanager;
+
+import android.app.Notification;
+import android.app.NotificationChannel;
+import android.content.Context;
+import android.os.Build;
+import android.support.v4.app.NotificationCompat;
+
+public class NotificationConfig<T> implements NotificationChannelCreator, NotificationCreator<T> {
```
joetimmins: This class replaces the old implementations of `NotificationChannelCreator` and `NotificationCreator`. Passing this many constructor parameters ensures that the clients never have to specify IDs or descriptions more than once, removing a potential source of bugs.

joetimmins: Happy to rename / create a builder / anything else here.

library/src/main/java/com/novoda/downloadmanager/NotificationConfig.java
```diff
@@ -0,0 +1,53 @@
...
+        return new NotificationInformation() {
+            @Override
+            public int getId() {
+                return notificationPayload.hashCode();
+            }
+
+            @Override
+            public Notification getNotification() {
+                NotificationCompat.Builder builder = new NotificationCompat.Builder(applicationContext, channelId);
+                return notificationCustomiser.customNotificationFrom(builder, notificationPayload);
```
joetimmins: The `notificationCustomiser` field is passed in by the client. They will need to use the supplied `builder` in that implementation as it will have been pre-configured with the correct channel name. We can write Javadoc explaining that when we come to that task.

library/src/main/java/com/novoda/downloadmanager/NotificationInformation.java
```diff
@@ -2,7 +2,7 @@
 
 import android.app.Notification;
 
-public interface NotificationInformation {
+interface NotificationInformation {
```
joetimmins: This class can be converted to a data class quite easily now but the pull request was already big enough.

demo-simple/src/main/java/com/novoda/downloadmanager/demo/CustomNotificationConfigFactory.java
```diff
@@ -0,0 +1,50 @@
...
+import com.novoda.downloadmanager.DownloadBatchTitle;
+import com.novoda.downloadmanager.NotificationConfig;
+import com.novoda.downloadmanager.NotificationCustomizer;
+import com.novoda.notils.logger.simple.Log;
+
+public final class CustomNotificationConfigFactory {
+
+    private static final boolean NOT_INDETERMINATE = false;
+
+    public static NotificationConfig<DownloadBatchStatus> createDownloadNotificationConfig(Context context, String channelId, String channelDescription, final int iconDrawable) {
```
Mecharyry: If we are planning to have lint etc to run on this later then we will need to include `private` constructors on our factories. Can you add one here please 😃

Mecharyry: I normally add a comment saying "Uses static factory methods."

library/src/main/java/com/novoda/downloadmanager/DownloadBatchNotification.java
Mecharyry: Your usage of icons for locating code ❤️

library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
```diff
@@ -66,9 +68,35 @@ public static DownloadManagerBuilder newInstance(Context context, Handler callba
...
+                        .setProgress(bytesFileSize, bytesDownloaded, NOT_INDETERMINATE)
+                        .setSmallIcon(notificationIcon)
+                        .setContentTitle(title)
+                        .setContentText(content)
+                        .build();
+
+            }
+        };
+        NotificationConfig<DownloadBatchStatus> notificationConfig = new NotificationConfig<>(
+                context,
```
Mecharyry: I'm assuming that this should match other areas in here? If so it should use the `applicationContext` defined above 😸

library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
Mecharyry: Will this change in a follow-up PR?

library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
joetimmins: `NotificationConfig` calls `getApplicationContext` on whichever context gets passed in so it doesn't matter

library/src/main/java/com/novoda/downloadmanager/NotificationConfig.java
zegnus: Now there is only one implementation for both interfaces, what about if we unify them under one interface, make it public and then you make this class private, only a factory could be visible (we do that in other parts of the app) so that only the minimum is visible.

Also I have a question, now the client is forced to create a channel if they only want to provide a custom notification. Can we make it more flexible?

library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
Mecharyry: I think I would prefer these as anonymous inner classes. I think we are at risk of making these constructors super hard to follow. If we pull them out into anonymous classes at least then they are self contained. What do you think?

library/src/main/java/com/novoda/downloadmanager/DownloadMigratorBuilder.java
```diff
@@ -1,45 +1,66 @@
...
+    private NotificationConfig<MigrationStatus> notificationConfig;
 
     public static DownloadMigratorBuilder newInstance(Context context) {
         Context applicationContext = context.getApplicationContext();
-        NotificationChannelCreator notificationChannelCreator = new MigrationNotificationChannelCreator(applicationContext.getResources());
-        NotificationCreator<MigrationStatus> notificationCreator = new MigrationNotification(applicationContext, android.R.drawable.ic_menu_gallery);
+
+        String channelId = context.getResources().getString(R.string.migration_notification_channel_name);
+        String channelDescription = context.getResources().getString(R.string.migration_notification_channel_description);
+        NotificationCustomizer<MigrationStatus> customiser = new NotificationCustomizer<MigrationStatus>() {
```
Mecharyry: > I think I would prefer these as anonymous inner classes. I think we are at risk of making these constructors super hard to follow. If we pull them out into anonymous classes at least then they are self contained. What do you think?

Could we do this here too?

library/src/main/java/com/novoda/downloadmanager/NotificationConfig.java
Mecharyry: The `Config` bothers me a bit because it's a shortcut 😂 but tbh it's just me being a nit 😄

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
```diff
@@ -89,8 +89,7 @@ private boolean notificationChannelDoesNotExist(String channelName) {
     private final MigrationCallback notificationMigrationCallback = new MigrationCallback() {
         @Override
         public void onUpdate(MigrationStatus migrationStatus) {
-            String channelName = notificationChannelCreator.getNotificationChannelName();
-            NotificationInformation notification = notificationCreator.createNotification(channelName, migrationStatus);
+            NotificationInformation notification = notificationCreator.createNotification(migrationStatus);
```
zegnus: not used right?

joetimmins: Everything here is used! 😅

joetimmins:
@zegnus 
> now the client is forced to create a channel if they only want to provide a custom notification. Can we make it more flexible?

They absolutely have to provide a channel. If they're using this lib it means they're targeting API 26 at a minimum and all notifications in API 26 and above require a channel. If they're not targeting 26, this lib doesn't even compile.

joetimmins:
Also @zegnus 
> Now there is only one implementation for both interfaces, what about if we unify them under one interface, make it public and then you make this class private, only a factory could be visible (we do that in other parts of the app) so that only the minimum is visible.

We'll do that on the next PR, this one is already big.

library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
joetimmins: Yes, planning to change it in the next PR or the one after.

library/src/main/java/com/novoda/downloadmanager/NotificationConfig.java
```diff
@@ -0,0 +1,53 @@
...
+
+public class NotificationConfig<T> implements NotificationChannelCreator, NotificationCreator<T> {
+
+    private final Context applicationContext;
+    private final String channelId;
+    private final String userFacingChannelDescription;
+    private final NotificationCustomizer<T> notificationCustomizer;
+    private final int importance;
+
+    public NotificationConfig(Context context, String channelId, String userFacingChannelDescription, NotificationCustomizer<T> customizer, int importance) {
```
zegnus: Do you know if its possible to annotate the `int importance` param so that we know the valid range of values?

joetimmins: Yes we can do that!

PR #278, opened 2017-12-22
Incomplete downloads

Mecharyry:
### Problem
The new version of the `download-manager` is capable of migrating **complete** downloads from v1 to v2. We also need to be able to migrate **incomplete** downloads, those that do not have a local file path or have a local file path but where the current bytes do not match the total bytes for a batch.

### Solution
Create a `PartialDownloadMigrationExtractor` that is responsible for extracting a `List<Migration>` that represents just the partial downloads. For partials we check whether the whole batch is downloaded. If the whole batch is not downloaded then we mark it with **queueing** which will automatically trigger a re-download of the files. For the partial downloads that had a local file we delete rather than carrying on from the current byte range. 

If we decide to carry on from a given byte range this would be more effort and greater risk for the migration process.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/Migration.java
+++ b/library/src/main/java/com/novoda/downloadmanager/Migration.java
@@ -21,6 +21,15 @@ class Migration {
         return fileMetadata;
     }
 
+    boolean hasDownloadedBatch() {
+        for (FileMetadata fileMetadatum : fileMetadata) {
+            if (fileMetadatum.fileSize().currentSize() != fileMetadatum.fileSize().totalSize()) {
+                return false;
+            }
+        }
+        return true;
+    }
+
     @Override
     public boolean equals(Object o) {
         if (this == o) {
--- a/library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
+++ b/library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
@@ -56,7 +56,6 @@ class MigrationJob implements Runnable {
         onUpdate(migrationStatus);
 
         Log.d(TAG, "about to extract migrations, time is " + System.nanoTime());
-        Log.d(TAG, "partial migrations are all EXTRACTED, time is " + System.nanoTime());
 
         migratePartialDownloads(database, partialDownloadMigrationExtractor, downloadsPersistence);
         migrateCompleteDownloads(migrationStatus, database, migrationExtractor, downloadsPersistence, internalFilePersistence);
@@ -83,6 +82,7 @@ class MigrationJob implements Runnable {
             database.setTransactionSuccessful();
             database.endTransaction();
         }
+        Log.d(TAG, "partial migrations are all EXTRACTED, time is " + System.nanoTime());
     }
 
     private void migrateV1DataToV2Database(DownloadsPersistence downloadsPersistence, Migration migration) {
--- a/library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
+++ b/library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
@@ -131,7 +131,11 @@ class MigrationJob implements Runnable {
         return metadata.originalFileLocation() != null && !metadata.originalFileLocation().isEmpty();
     }
 
-    private void migrateCompleteDownloads(InternalMigrationStatus migrationStatus, SqlDatabaseWrapper database, MigrationExtractor migrationExtractor, DownloadsPersistence downloadsPersistence, InternalFilePersistence internalFilePersistence) {
+    private void migrateCompleteDownloads(InternalMigrationStatus migrationStatus,
+                                          SqlDatabaseWrapper database,
+                                          MigrationExtractor migrationExtractor,
+                                          DownloadsPersistence downloadsPersistence,
+                                          InternalFilePersistence internalFilePersistence) {
         List<Migration> migrations = migrationExtractor.extractMigrations();
         Log.d(TAG, "migrations are all EXTRACTED, time is " + System.nanoTime());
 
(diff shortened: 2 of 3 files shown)


library/src/main/java/com/novoda/downloadmanager/Migration.java
```diff
@@ -21,6 +21,15 @@ Batch batch() {
         return fileMetadata;
     }
 
+    boolean hasDownloadedBatch() {
```
Mecharyry: Might be able to switch this with a better query that checks the batch size from the v1 database. I'd rather merge this and make improvements in a follow-up. We know this works after all 😄

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
```diff
@@ -53,6 +56,86 @@ public void run() {
...
+        migrateCompleteDownloads(migrationStatus, database, migrationExtractor, downloadsPersistence, internalFilePersistence);
+    }
+
+    private void onUpdate(MigrationStatus migrationStatus) {
+        for (MigrationCallback migrationCallback : migrationCallbacks) {
+            migrationCallback.onUpdate(migrationStatus);
+        }
+    }
+
+    private void migratePartialDownloads(SqlDatabaseWrapper database, PartialDownloadMigrationExtractor partialDownloadMigrationExtractor, DownloadsPersistence downloadsPersistence) {
```
Mecharyry: ❓ QUESTION TIME! ❓ 

I'm wondering if we should have several jobs at this point and extract out some collaborators for these different migration steps.

We could have:
`PartialDownloadMigrationJob`
`CompleteDownloadMigrationJob`
`DeleteV1DatabaseJob`

My thinking is this might make the whole process easier to test. What do you think?

library/src/main/java/com/novoda/downloadmanager/PartialDownloadMigrationExtractor.java
```diff
@@ -0,0 +1,64 @@
+package com.novoda.downloadmanager;
+
+import android.database.Cursor;
+
+import java.util.ArrayList;
+import java.util.List;
+
+class PartialDownloadMigrationExtractor {
+
+    private static final String BATCHES_QUERY = "SELECT batches._id, batches.batch_title FROM batches INNER JOIN DownloadsByBatch ON " +
```
Mecharyry: I think we can make this query a bit more intelligent. Would rather merge and make improvements in a follow-up PR today.

library/src/main/java/com/novoda/downloadmanager/Migration.java
```diff
@@ -21,6 +21,15 @@ Batch batch() {
...
+        for (FileMetadata fileMetadatum : fileMetadata) {
```
joetimmins: nice use of the singular `datum` here 👍

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
joetimmins: We can try and tests for what we have first and let the pain points guide us towards a better more testable solution.

library/src/main/java/com/novoda/downloadmanager/Migration.java
Mecharyry: I wish I could take credit here 😄  this was ASide 😄

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
Mecharyry: In that case, I'll start testing these `MigrationJob`s now in a separate branch to see where that gets me 😄

library/src/main/java/com/novoda/downloadmanager/PartialDownloadMigrationExtractor.java
joetimmins: do we not need `SELECT DISTINCT` here?

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
```diff
@@ -53,6 +56,86 @@ public void run() {
         onUpdate(migrationStatus);
 
         Log.d(TAG, "about to extract migrations, time is " + System.nanoTime());
+        Log.d(TAG, "partial migrations are all EXTRACTED, time is " + System.nanoTime());
```
Frankie-Boy: Shouldn't this be logged out after `migratePartialDownloads()` (just a couple lines down 😛  )

library/src/main/java/com/novoda/downloadmanager/PartialDownloadMigrationExtractor.java
Mecharyry: No, you can do the same with a `GROUP_BY` 😄

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
```diff
@@ -53,6 +56,86 @@ public void run() {
...
+
+    private void onUpdate(MigrationStatus migrationStatus) {
+        for (MigrationCallback migrationCallback : migrationCallbacks) {
+            migrationCallback.onUpdate(migrationStatus);
+        }
+    }
+
+    private void migratePartialDownloads(SqlDatabaseWrapper database, PartialDownloadMigrationExtractor partialDownloadMigrationExtractor, DownloadsPersistence downloadsPersistence) {
+        List<Migration> partialMigrations = partialDownloadMigrationExtractor.extractMigrations();
+        for (int i = 0; i < partialMigrations.size(); i++) {
```
Frankie-Boy: Could use a For Each here instead I think 👍

Mecharyry: Ah true, I'm not using the index like in the other one.

PR #279, opened 2017-12-22
Use @RequiresApi annotation for added runtime safety

joetimmins:
Android API levels 26 and above (so Android 8 Oreo 🍪 and any future releases) need to use `NotificationChannel` objects in order to create user notifications. `NotificationChannel` objects cannot be created on devices running Android OSs lower than API level 26.

Previously, we were guarding against code being run the wrong devices by returning `Optional<NotificationChannel>` and then doing a runtime OS level check before calling `optional.get()`. We can simplify this by annotating the method returning `NotificationChannel` with `@RequiresApi(OREO)`. This annotation means that calling this method is a compile error, unless you have first checked you are running on Oreo or above.

This change means we don't need the `Optional` implementation at all in this library.


library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
```diff
@@ -73,19 +72,12 @@ public void startMigration(MigrationCallback migrationCallback) {
...
 
     private void createNotificationChannel() {
-        Optional<NotificationChannel> notificationChannel = notificationChannelCreator.createNotificationChannel();
-        String channelName = notificationChannelCreator.getNotificationChannelId();
-
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O && notificationChannel.isPresent() && notificationChannelDoesNotExist(channelName)) {
-            notificationManager.createNotificationChannel(notificationChannel.get());
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            NotificationChannel notificationChannel = notificationChannelCreator.createNotificationChannel();
+            notificationManager.createNotificationChannel(notificationChannel);
```
joetimmins: We also don't need to check if the channel already exists before creating it. Calling `createNotificationChannel` multiple times with identical `NotificationChannel` arguments still results in only one channel being created.

library/src/main/java/com/novoda/downloadmanager/NotificationChannelCreator.java
```diff
@@ -1,11 +1,10 @@
...
 
 import android.app.NotificationChannel;
+import android.os.Build;
+import android.support.annotation.RequiresApi;
 
 interface NotificationChannelCreator {
-
-    Optional<NotificationChannel> createNotificationChannel();
-
-    String getNotificationChannelId();
```
joetimmins: 🚨

library/src/main/java/com/novoda/downloadmanager/NotificationConfig.java
```diff
@@ -43,17 +44,10 @@ public Notification getNotification() {
...
-            return Optional.of(new NotificationChannel(channelId, userFacingChannelDescription, importance));
-        }
-        return Optional.absent();
-    }
-
-    @Override
-    public String getNotificationChannelId() {
-        return channelId;
+    public NotificationChannel createNotificationChannel() {
+        return new NotificationChannel(channelId, userFacingChannelDescription, importance);
```
joetimmins: The annotation means we don't need the runtime check in here either.

library/src/main/java/com/novoda/downloadmanager/NotificationChannelCreator.java
```diff
@@ -1,11 +1,10 @@
...
-
+    @RequiresApi(Build.VERSION_CODES.O)
```
Mecharyry: Can this annotation live on the interface rather than the method? Just thinking that it applies to everything in here atm. Not sure if that is just being a nit 😄

joetimmins: We actually need it in both places - the callers of this method only know about the interface so they need it on the interface method. Unfortunately method annotations aren't inherited from interfaces so unless we have it on the actual implementation as well, we would need the runtime `Build.VERSION_CODES` check.

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
tasomaniac: Yeah, in the examples too. Google suggests to create the channel without checking anything in the constructor of the class responsible for notifications.

PR #280, opened 2017-12-22
Clean up notification config

Frankie-Boy:
## Problem
`NotificationCreator` & `NotificationChannelCreator` interfaces could be unified into just one new interface.

## Solution
Created a new interface which combined the two old interfaces and update all usages around the library and app. We added a Factory on the interface to create an implementation of NotificationCreator. We did not do this as a builder because all of the parameter passed are mandatory.


library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerDownloader.java
```diff
@@ -39,7 +37,6 @@
         this.fileOperations = fileOperations;
         this.downloadsBatchPersistence = downloadsBatchPersistence;
         this.downloadsFilePersistence = downloadsFilePersistence;
-        this.notificationChannelCreator = notificationChannelCreator;
```
Frankie-Boy: This wasn't being used.

zegnus:
Looks like the `NotificationInformation` must be public. Try to inject a custom notification in the `DemoApplication` download builder

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -61,7 +61,7 @@ protected void onCreate(Bundle savedInstanceState) {
         );
 
         downloadMigrator = DownloadMigratorBuilder.newInstance(this)
-                .withNotificationConfig(notificationConfig)
+                .withNotification(notificationCreator)
```
zegnus: So now the client has two options.

1. Just use what's requested:
```
downloadMigrator = DownloadMigratorBuilder.newInstance(this)
                .withNotification(new NotificationCreator<MigrationStatus>() {
                    @Override
                    public NotificationInformation createNotification(MigrationStatus notificationPayload) {
                        return null;
                    }

                    @Override
                    public NotificationChannel createNotificationChannel() {
                        return null;
                    }
                })
                .build();
```

2. Use the builder:
```
NotificationCreator<MigrationStatus> notificationCreator = NotificationCreator.Factory.build(
                this,
                "chocolate",
                "Migration notifications",
                notificationCustomizer,
                NotificationManagerCompat.IMPORTANCE_DEFAULT
        );
```

is that ok, expected?

joetimmins: Option one is disallowed. The client should only get an instance of `NotificationCreator` by using the factory. We can explain this in Javadoc when we do that task. This is also why `NotificationInformation` is not public.

In fact `NotificationInformation` does not need to be an interface, it can just be a data POJO with getters.

zegnus: let's have a chat on that because it is allowed currently, preventing in through javadocs is not good enough...

zegnus: We have agreed to only expose just one one way of interacting with the notification creator

PR #282, opened 2018-01-03
Add static analysis

Mecharyry:
:x: ~Blocked until #283 is merged~ ❌ 

## Problem

We're not doing static analysis on this project...

## Solution

Add the standard static analysis setup, with rules. Includes the fixes from #283. Main fixes on this branch are in the `team-props` directory, which is the standard setup for static analysis.

### Paired with 

Nobody

PR #283, opened 2018-01-03
Fix issues associated with Static Analysis

Mecharyry:
## Problem
We now have static analysis as of #282 but it throws up a lot of issues 😢 

## Solution
Attempt to fix the issues thrown up by the static analysis.

### DIFF
--- a/demo-simple/build.gradle
+++ b/demo-simple/build.gradle
@@ -28,11 +28,13 @@ android {
 }
 
 dependencies {
-    compile fileTree(dir: 'libs', include: ['*.jar'])
-    androidTestCompile('com.android.support.test.espresso:espresso-core:2.2.2', {
+    implementation fileTree(dir: 'libs', include: ['*.jar'])
+    androidTestImplementation('com.android.support.test.espresso:espresso-core:2.2.2', {
         exclude group: 'com.android.support', module: 'support-annotations'
     })
-    compile 'com.android.support:appcompat-v7:27.0.2'
-    testCompile 'junit:junit:4.12'
-    compile project(path: ':library')
+    implementation 'com.novoda:notils:2.2.16'
+    implementation 'com.facebook.stetho:stetho:1.5.0'
+    implementation 'com.android.support:appcompat-v7:27.0.2'
+    testImplementation 'junit:junit:4.12'
+    implementation project(path: ':library')
 }
--- a/demo-simple/build.gradle
+++ b/demo-simple/build.gradle
@@ -29,12 +29,12 @@ android {
 
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
-    androidTestImplementation('com.android.support.test.espresso:espresso-core:2.2.2', {
-        exclude group: 'com.android.support', module: 'support-annotations'
-    })
     implementation 'com.novoda:notils:2.2.16'
     implementation 'com.facebook.stetho:stetho:1.5.0'
     implementation 'com.android.support:appcompat-v7:27.0.2'
-    testImplementation 'junit:junit:4.12'
     implementation project(path: ':library')
+    testImplementation 'junit:junit:4.12'
+    androidTestImplementation('com.android.support.test.espresso:espresso-core:2.2.2', {
+        exclude group: 'com.android.support', module: 'support-annotations'
+    })
 }
--- a/demo-simple/build.gradle
+++ b/demo-simple/build.gradle
@@ -14,8 +14,9 @@ android {
     }
 
     lintOptions {
-        abortOnError false
-        lintConfig file("$rootProject.projectDir/team-props/lint-rules/config.xml")
+        lintConfig teamPropsFile('static-analysis/lint-config.xml')
+        abortOnError true
+        warningsAsErrors true
     }
 
     buildTypes {
--- a/demo-simple/build.gradle
+++ b/demo-simple/build.gradle
@@ -1,13 +1,13 @@
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 26
-    buildToolsVersion "26.0.2"
+    compileSdkVersion 27
+    buildToolsVersion "27.0.3"
 
     defaultConfig {
         applicationId "com.novoda.downloadmanager.demo.simple"
         minSdkVersion 14
-        targetSdkVersion 26
+        targetSdkVersion 27
         versionCode 1
         versionName "1.0"
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
@@ -32,7 +32,7 @@ dependencies {
     androidTestCompile('com.android.support.test.espresso:espresso-core:2.2.2', {
         exclude group: 'com.android.support', module: 'support-annotations'
     })
-    compile 'com.android.support:appcompat-v7:26.1.0'
+    compile 'com.android.support:appcompat-v7:27.0.2'
     testCompile 'junit:junit:4.12'
     compile project(path: ':library')
 }
--- a/demo-simple/src/main/AndroidManifest.xml
+++ b/demo-simple/src/main/AndroidManifest.xml
@@ -3,13 +3,13 @@
   package="com.novoda.downloadmanager.demo">
 
   <application
-    android:allowBackup="true"
+    android:name="com.novoda.downloadmanager.demo.DemoApplication"
+    android:allowBackup="false"
     android:icon="@mipmap/ic_launcher"
     android:label="@string/app_name"
     android:roundIcon="@mipmap/ic_launcher_round"
     android:supportsRtl="true"
-    android:theme="@style/AppTheme"
-    android:name="com.novoda.downloadmanager.demo.DemoApplication">
+    android:theme="@style/AppTheme">
 
     <activity android:name="com.novoda.downloadmanager.demo.MainActivity">
       <intent-filter>
--- a/demo-simple/src/main/java/com/novoda/downloadmanager/demo/CustomFileDownloader.java
+++ b/demo-simple/src/main/java/com/novoda/downloadmanager/demo/CustomFileDownloader.java
@@ -10,7 +10,7 @@ class CustomFileDownloader implements FileDownloader {
     private static final int BUFFER_SIZE = 5000;
     private static final int SLEEP_IN_MILLIS = 200;
 
-    private byte[] buffer = new byte[BUFFER_SIZE];
+    private final byte[] buffer = new byte[BUFFER_SIZE];
 
     private boolean canDownload;
 
@@ -24,7 +24,7 @@ class CustomFileDownloader implements FileDownloader {
             try {
                 Thread.sleep(SLEEP_IN_MILLIS);
             } catch (InterruptedException e) {
-                e.printStackTrace();
+                Log.e("CustomFileDownloader Thread interrupted.", e);
             }
             callback.onBytesRead(buffer, BYTES_READ);
         }
--- a/demo-simple/src/main/java/com/novoda/downloadmanager/demo/CustomFileDownloader.java
+++ b/demo-simple/src/main/java/com/novoda/downloadmanager/demo/CustomFileDownloader.java
@@ -1,14 +1,16 @@
 package com.novoda.downloadmanager.demo;
 
-import com.novoda.notils.logger.simple.Log;
 import com.novoda.downloadmanager.FileDownloader;
 import com.novoda.downloadmanager.FileSize;
+import com.novoda.notils.logger.simple.Log;
 
 class CustomFileDownloader implements FileDownloader {
 
     private static final int BYTES_READ = 50;
+    private static final int BUFFER_SIZE = 5000;
+    private static final int SLEEP_IN_MILLIS = 200;
 
-    private byte[] buffer = new byte[5000];
+    private byte[] buffer = new byte[BUFFER_SIZE];
 
     private boolean canDownload;
 
@@ -20,7 +22,7 @@ class CustomFileDownloader implements FileDownloader {
 
         while (canDownload && fileSize.currentSize() < fileSize.totalSize()) {
             try {
-                Thread.sleep(200);
+                Thread.sleep(SLEEP_IN_MILLIS);
             } catch (InterruptedException e) {
                 e.printStackTrace();
             }
--- a/demo-simple/src/main/java/com/novoda/downloadmanager/demo/VersionOneDatabaseCloner.java
+++ b/demo-simple/src/main/java/com/novoda/downloadmanager/demo/VersionOneDatabaseCloner.java
@@ -68,8 +68,7 @@ class VersionOneDatabaseCloner {
         boolean parentPathDoesNotExist = !outputFile.getParentFile().exists();
         if (parentPathDoesNotExist) {
             Log.w(String.format("path: %s doesn't exist, creating parent directories...", outputFile.getAbsolutePath()));
-            outputFile.getParentFile().mkdirs();
-            parentPathDoesNotExist = !outputFile.getParentFile().exists();
+            parentPathDoesNotExist = !outputFile.getParentFile().mkdirs();
         }
 
         if (parentPathDoesNotExist) {
@@ -78,8 +77,8 @@ class VersionOneDatabaseCloner {
     }
 
     private void copyAssetToFile(String assetName, File outputFile) {
-        InputStream inputStream;
-        OutputStream myOutput;
+        InputStream inputStream = null;
+        OutputStream myOutput = null;
         int length;
         try {
             inputStream = assetManager.open(assetName);
@@ -87,14 +86,24 @@ class VersionOneDatabaseCloner {
             while ((length = inputStream.read(BUFFER)) > 0) {
                 myOutput.write(BUFFER, 0, length);
             }
-            myOutput.close();
-            myOutput.flush();
-            inputStream.close();
-
             Log.d(getClass().getSimpleName(), "Copied asset: " + assetName);
         } catch (IOException e) {
             e.printStackTrace();
             Log.e("Failed to copy asset: " + assetName, e);
+        } finally {
+            try {
+                if (myOutput != null) {
+                    myOutput.close();
+                    myOutput.flush();
+                }
+
+                if (inputStream != null) {
+                    inputStream.close();
+                }
+            } catch (IOException e) {
+                e.printStackTrace();
+                Log.e("Failed to close streams.", e);
+            }
         }
     }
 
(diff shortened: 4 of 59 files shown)


demo-simple/build.gradle
```diff
@@ -28,11 +28,13 @@ android {
 }
 
 dependencies {
-    compile fileTree(dir: 'libs', include: ['*.jar'])
-    androidTestCompile('com.android.support.test.espresso:espresso-core:2.2.2', {
+    implementation fileTree(dir: 'libs', include: ['*.jar'])
+    implementation 'com.novoda:notils:2.2.16'
+    implementation 'com.facebook.stetho:stetho:1.5.0'
```
Mecharyry: The two above are cloned from `library build.gradle` didn't expose with `api` because we probably don't want to expose it to clients of the library. We use these for debugging only.

demo-simple/src/main/AndroidManifest.xml
```diff
@@ -3,13 +3,13 @@
   package="com.novoda.downloadmanager.demo">
 
   <application
-    android:allowBackup="true"
+    android:name="com.novoda.downloadmanager.demo.DemoApplication"
+    android:allowBackup="false"
```
Mecharyry: @zegnus not sure why we made this `allowBackup` for the demo, put to false as it seems unnecessary.

demo-simple/src/main/java/com/novoda/downloadmanager/demo/CustomFileDownloader.java
```diff
@@ -20,9 +22,9 @@ public void startDownloading(String url, FileSize fileSize, Callback callback) {
 
         while (canDownload && fileSize.currentSize() < fileSize.totalSize()) {
             try {
-                Thread.sleep(200);
+                Thread.sleep(SLEEP_IN_MILLIS);
             } catch (InterruptedException e) {
-                e.printStackTrace();
+                Log.e("CustomFileDownloader Thread interrupted.", e);
```
Mecharyry: Replaced all occurrences of `e.printStackTrace` with a `Logger`.

demo-simple/src/main/java/com/novoda/downloadmanager/demo/VersionOneDatabaseCloner.java
```diff
@@ -77,23 +77,31 @@ private void createFileIfDoesNotExist(File outputFile) {
...
-            myOutput.flush();
-            inputStream.close();
-
             Log.d(getClass().getSimpleName(), "Copied asset: " + assetName);
         } catch (IOException e) {
-            e.printStackTrace();
             Log.e("Failed to copy asset: " + assetName, e);
+        } finally {
+            try {
+                if (myOutput != null) {
```
Mecharyry: We weren't correctly handling the closing down of the input and output streams before, exposing us to potentially unhandled `IOException`

library/src/main/java/com/novoda/downloadmanager/AndroidLocalFilesDirectory.java
```diff
@@ -15,7 +17,15 @@
 
     @Override
     public List<String> contents() {
-        return Arrays.asList(context.getFilesDir().list());
+        File filesDir = context.getFilesDir();
+        if (filesDir != null) {
+            String[] directoryAsStrings = filesDir.list();
+            if (directoryAsStrings != null) {
+                return Arrays.asList(directoryAsStrings);
```
Mecharyry: Possible null pointer dereference when applying just `Arrays.asList(filesDir.list())` as `filesDir.list()` can potentially return a null and `Arrays.asList` expects `non-null` values.

https://sbforge.org/sonar/rules/show/findbugs:NP_NULL_ON_SOME_PATH?layout=false

library/src/main/java/com/novoda/downloadmanager/Batch.java
```diff
@@ -9,7 +9,7 @@
     private final String title;
     private final List<String> fileUrls;
 
-    private Batch(DownloadBatchId downloadBatchId, String title, List<String> fileUrls) {
+    Batch(DownloadBatchId downloadBatchId, String title, List<String> fileUrls) {
```
Mecharyry: We use a `Builder` to create this normally. We should make this package as it is only accessed internally anyway. The `Builder` is just a utility to make this easier.

library/src/main/java/com/novoda/downloadmanager/CallbackThrottleCreator.java
```diff
@@ -19,15 +19,15 @@
     private final long frequency;
     private final Class<? extends CallbackThrottle> customCallbackThrottle;
 
-    static CallbackThrottleCreator ByTime(TimeUnit timeUnit, long quantity) {
```
Mecharyry: `ByTime` 😂

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -3,8 +3,15 @@
...
+import static com.novoda.downloadmanager.DownloadBatchStatus.Status.PAUSED;
+import static com.novoda.downloadmanager.DownloadBatchStatus.Status.DELETION;
+import static com.novoda.downloadmanager.DownloadBatchStatus.Status.DOWNLOADED;
+import static com.novoda.downloadmanager.DownloadBatchStatus.Status.ERROR;
+import static com.novoda.downloadmanager.DownloadBatchStatus.Status.QUEUED;
+import static com.novoda.downloadmanager.DownloadBatchStatus.Status.DOWNLOADING;
+
+// This model knows how to interact with low level components.
+@SuppressWarnings({"PMD.CyclomaticComplexity", "PMD.StdCyclomaticComplexity", "PMD.ModifiedCyclomaticComplexity"})
 class DownloadBatch {
```
Mecharyry: I've added these suppressions as this is a complex class that has knowledge of the low level components in the `download-manager`.

library/src/main/java/com/novoda/downloadmanager/LiteDownloadService.java
```diff
@@ -84,7 +85,7 @@ private void acquireCpuWakeLock() {
         PowerManager powerManager = (PowerManager) getSystemService(POWER_SERVICE);
         if (powerManager != null) {
             wakeLock = powerManager.newWakeLock(PowerManager.PARTIAL_WAKE_LOCK, WAKELOCK_TAG);
-            wakeLock.acquire();
+            wakeLock.acquire(TEN_MINUTES_IN_MILLIS);
```
Mecharyry: @zegnus ASide and lint flagged this as an issue, are there any drawbacks that you can see with adding this timeout? I think it is fine but would be good to hear your thoughts.

library/src/main/java/com/novoda/downloadmanager/LiteFileName.java
```diff
@@ -5,7 +5,7 @@
     private final String name;
 
     static LiteFileName from(Batch batch, String fileUrl) {
-        String name = batch + fileUrl + String.valueOf(System.nanoTime());
```
Mecharyry: Don't need `String.valueOf` if you are already concatenating `String`s.

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
```diff
@@ -121,7 +123,9 @@ private void deleteFrom(SqlDatabaseWrapper database, Migration migration) {
         for (Migration.FileMetadata metadata : migration.getFileMetadata()) {
             if (hasValidFileLocation(metadata)) {
                 File file = new File(metadata.originalFileLocation());
-                file.delete();
+                boolean deleted = file.delete();
```
Mecharyry: ATM, as far as I know, we hope that we are deleting files correctly. Perhaps we should raise an issue to deal with failures?

library/src/main/java/com/novoda/downloadmanager/NetworkFileDownloader.java
```diff
@@ -60,6 +46,28 @@ public void startDownloading(String url, FileSize fileSize, Callback callback) {
         callback.onDownloadFinished();
     }
 
+    private void processResponse(Callback callback, HttpClient.NetworkResponse response, int responseCode) throws IOException {
```
Mecharyry: I wasn't sure how to break this code down more but it is hitting the cyclomatic complexity limit 😬

library/src/main/java/com/novoda/downloadmanager/RoomAppDatabase.java
```diff
@@ -8,24 +8,33 @@
...
+    private static final Object LOCK = new Object();
 
     abstract RoomBatchDao roomBatchDao();
 
     abstract RoomFileDao roomFileDao();
 
     static RoomAppDatabase obtainInstance(Context context) {
-        if (SINGLE_INSTANCE == null) {
-            SINGLE_INSTANCE = newInstance(context);
+        RoomAppDatabase database = singleInstance;
```
Mecharyry: This post gives a rather nice explanation for thread safe singletons 😄  https://www.journaldev.com/1377/java-singleton-design-pattern-best-practices-examples#thread-safe-singleton

library/src/main/java/com/novoda/downloadmanager/RoomBatch.java
```diff
@@ -6,7 +6,8 @@
 import android.arch.persistence.room.PrimaryKey;
 import android.support.annotation.NonNull;
 
-@Entity(indices={@Index("batch_id")})
+@SuppressWarnings("checkstyle:visibilitymodifier") // Accessors will add a lot of boilerplate code.
```
Mecharyry: Can add the accessors but it seems unnecessary to me given that this is only used internally and it is super easy to read atm.

team-props/static-analysis.gradle
```diff
@@ -3,7 +3,8 @@ apply plugin: 'com.novoda.static-analysis'
 staticAnalysis {
     penalty {
         maxErrors = 0
-        maxWarnings = 0
+//      Need to address GH Issue #284 before reducing.
```
Mecharyry: See https://github.com/novoda/download-manager/issues/284

library/src/main/java/com/novoda/downloadmanager/LiteDownloadService.java
zegnus: Yeah we had to add it 👍

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
zegnus: yeah I know that one, its usually ignored as not much can be done though

Mecharyry: Yea that's why I found the warning kind of annoying. Short of surfacing to a user there isn't much we can do 😂

library/src/main/java/com/novoda/downloadmanager/DownloadsNetworkRecoveryCreator.java
```diff
@@ -2,23 +2,27 @@
...
 
-class DownloadsNetworkRecoveryCreator {
+final class DownloadsNetworkRecoveryCreator {
 
-    private static DownloadsNetworkRecovery INSTANCE;
+    private DownloadsNetworkRecoveryCreator() {
+        // Uses static factory methods.
+    }
+
+    private static DownloadsNetworkRecovery singleInstance;
```
zegnus: shouldn't the constant be upper case?

library/src/main/java/com/novoda/downloadmanager/DownloadsNetworkRecoveryCreator.java
```diff
@@ -2,23 +2,27 @@
 
 import android.content.Context;
 
-class DownloadsNetworkRecoveryCreator {
+final class DownloadsNetworkRecoveryCreator {
 
-    private static DownloadsNetworkRecovery INSTANCE;
+    private DownloadsNetworkRecoveryCreator() {
```
zegnus: should the constructor be after all class fields?

library/src/main/java/com/novoda/downloadmanager/FilePathCreator.java
```diff
@@ -1,9 +1,13 @@
 package com.novoda.downloadmanager;
 
-public class FilePathCreator {
+public final class FilePathCreator {
 
     private static final String UNKNOWN = "unknown";
 
+    private FilePathCreator() {
```
zegnus: should the constructor be after all class fields?

library/src/main/java/com/novoda/downloadmanager/RoomAppDatabase.java
```diff
@@ -8,24 +8,33 @@
 @Database(entities = {RoomBatch.class, RoomFile.class}, version = 1)
 abstract class RoomAppDatabase extends RoomDatabase {
 
-    private static volatile RoomAppDatabase SINGLE_INSTANCE;
+    private static volatile RoomAppDatabase singleInstance;
```
zegnus: constant in upper case?

library/src/main/java/com/novoda/downloadmanager/RoomAppDatabase.java
zegnus: Yup

library/src/main/java/com/novoda/downloadmanager/DownloadsNetworkRecoveryCreator.java
Mecharyry: It's not a constant. It's only a constant when it is `static` and `final`. This is still mutable, just at the class level.

library/src/main/java/com/novoda/downloadmanager/DownloadsNetworkRecoveryCreator.java
Mecharyry: 4be4c57

library/src/main/java/com/novoda/downloadmanager/FilePathCreator.java
Mecharyry: feb0b80

library/src/main/java/com/novoda/downloadmanager/RoomAppDatabase.java
joetimmins: isn't final

library/src/main/java/com/novoda/downloadmanager/RoomAppDatabase.java
```diff
@@ -8,24 +8,33 @@
...
     static RoomAppDatabase obtainInstance(Context context) {
-        if (SINGLE_INSTANCE == null) {
-            SINGLE_INSTANCE = newInstance(context);
+        RoomAppDatabase database = singleInstance;
+        if (database == null) {
+            synchronized (LOCK) {
+                database = singleInstance;
+                if (database == null) {
+                    singleInstance = newInstance(context);
+                    database = singleInstance;
```
joetimmins: Lines 19, 22 and 25 are the same. The previous solution was functionally the same as this but clearer to read and didn't need a lock.

Mecharyry: @joetimmins I linked this https://www.journaldev.com/1377/java-singleton-design-pattern-best-practices-examples#thread-safe-singleton which suggests that both solutions are very different. I'm happy to change it back if you can explain why this is the same?

Mecharyry: Static analysis complains about `synchronized` methods as opposed to static blocks because they are a performance hit.

Mecharyry: Saying that I can find a number of other articles that show the `double checked locking` mechanism is stupid too 😂  just can't win, whatever we do 😄

xrigau: was it functionally the same? with the new code only 1 instance of the database will be created but with the old one it'd be possible that 2 threads could create 2 instances because the synchronisation was only on the `newInstance` method, so the 2 threads could execute the null check `if (SINGLE_INSTANCE == null)` and then 1 of them runs `newInstance` while the other waits, once that completes the other thread executes `newInstance` again - correct me if I'm wrong but I think the new solution while less readable it is more correct (in terms of muti threading safety)

joetimmins: @xrigau - doesn't the `volatile` keyword prevent this?

zegnus: This is not a hyperthreading program and there is not a major issue if the migration and the downloader at some point have two different instances of the database, specially when the database has transactions.

Mecharyry: Yea that was my understanding too @xrigau 

I think though that @joetimmins is right about the readability and we should be able to achieve the same with:

```
     if (singleInstance == null) {
            synchronized (RoomAppDatabase.class) {
                if (singleInstance == null) {
                    singleInstance = newInstance(context);
                }
            }
        }
        return singleInstance;
```

zegnus: @joetimmins it solves in a different thread execution but not when two threads enters the `newInstance` in parallel. One thread will be blocked and the other will start, when it finishes, the blocked thread will enter and create the instance again; the double null check prevents that scenario and also only that method is blocked and not the entire class due to the synchronised method.

joetimmins: @Mecharyry the blog post you linked says that having the method containing the null check marked as `synchronized` is a performance hit, because it is called every time we obtain an instance. In the previous solution here, only the object instantiation was `synchronized`, the null check was not. This was because the single instance was marked as `volatile`, meaning it has to be retrieved from main memory every time it is read, and written to main memory whenever it is written, preventing other threads from having out of date copies of it. Which I think mitigates @xrigau 's concern too.

If this is wrong, please tell me what is actually happening here instead 😁

xrigau: this wikipedia article explains the differences between the solution in this PR and the code in your comment https://en.wikipedia.org/wiki/Double-checked_locking#Usage_in_Java but they're actually slightly different, regarding the `volatile` keyword I'm actually not sure I thought it only affected the order of reads and writes but not the synchronisation between threads and shared memory

Mecharyry: The updated code seems to match the solution you linked now @xrigau 😄  I think this is the neatest solution so far but I don't know if I have missed anything 😄

xrigau:
👍

PR #285, opened 2018-01-05
Get local file path from a Network address

Mecharyry:
### Problem
When trying to implement `download-manager` v2 into a client application we discovered that there is not an easy way to retrieve a specific download. 😬 

### Solution
Retrieve a given `download` from its network address. Extracted the `waitForDownloadService` into an easy to read runnable rather than duplicating the code in the `download-manager` for each new query.

### DIFF
--- a/demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
+++ b/demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
@@ -17,8 +17,10 @@ import com.novoda.downloadmanager.DownloadBatchCallback;
 import com.novoda.downloadmanager.DownloadBatchId;
 import com.novoda.downloadmanager.DownloadBatchIdCreator;
 import com.novoda.downloadmanager.DownloadBatchStatus;
+import com.novoda.downloadmanager.DownloadFilePathCallback;
 import com.novoda.downloadmanager.DownloadMigrator;
 import com.novoda.downloadmanager.DownloadMigratorBuilder;
+import com.novoda.downloadmanager.FilePath;
 import com.novoda.downloadmanager.LiteDownloadManagerCommands;
 import com.novoda.downloadmanager.LocalFilesDirectory;
 import com.novoda.downloadmanager.LocalFilesDirectoryFactory;
@@ -36,7 +38,9 @@ public class MainActivity extends AppCompatActivity {
 
     private static final DownloadBatchId BATCH_ID_1 = DownloadBatchIdCreator.createFrom("batch_id_1");
     private static final DownloadBatchId BATCH_ID_2 = DownloadBatchIdCreator.createFrom("batch_id_2");
-    private static final String FILE_URL = "http://ipv4.download.thinkbroadband.com/10MB.zip";
+    private static final String FIVE_MB_FILE_URL = "http://ipv4.download.thinkbroadband.com/5MB.zip";
+    private static final String TEN_MB_FILE_URL = "http://ipv4.download.thinkbroadband.com/10MB.zip";
+    private static final String TWENTY_FILE_URL = "http://ipv4.download.thinkbroadband.com/20MB.zip";
 
     private TextView databaseCloningUpdates;
     private TextView databaseMigrationUpdates;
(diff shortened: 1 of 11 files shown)


library/src/main/java/com/novoda/downloadmanager/WaitForLockRunnable.java
```diff
@@ -0,0 +1,44 @@
...
+
+import com.novoda.notils.logger.simple.Log;
+
+class WaitForLockRunnable {
+
+    interface Action {
+        void performAction();
+    }
+
+    static WaitForLockAndThenPerformActionRunnable waitFor(Object lock) {
```
Mecharyry: Introduced a nested builder so that we can do 

`WaitForLockRunnable.waitFor(downloadService).thenPerform(action)`

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -69,48 +76,22 @@ protected void onCreate(Bundle savedInstanceState) {
...
 
-        final Spinner spinner = findViewById(R.id.database_download_file_size);
+        downloadFileSizeSpinner = findViewById(R.id.database_download_file_size);
         ArrayAdapter<CharSequence> adapter = ArrayAdapter.createFromResource(this, R.array.file_sizes, android.R.layout.simple_spinner_item);
-        spinner.setAdapter(adapter);
+        downloadFileSizeSpinner.setAdapter(adapter);
 
         databaseCloningUpdates = findViewById(R.id.database_cloning_updates);
         View buttonCreateDB = findViewById(R.id.button_create_v1_db);
-        buttonCreateDB.setOnClickListener(new View.OnClickListener() {
```
Mecharyry: Moved to fields to remove the warning for long methods here.

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -32,18 +33,24 @@
 
 import static com.novoda.downloadmanager.DownloadBatchStatus.Status.ERROR;
 
+// Need to extract collaborators for this demo to reduce complexity. GH Issue #286
+@SuppressWarnings({"PMD.CyclomaticComplexity", "PMD.StdCyclomaticComplexity", "PMD.ModifiedCyclomaticComplexity"})
```
Mecharyry: This needs to be dealt with before release but would make this PR way larger and obscure the important parts. GH issue here #286

library/src/main/java/com/novoda/downloadmanager/DownloadManager.java
```diff
@@ -159,4 +145,32 @@ private void executeGetAllDownloadBatchStatuses(AllBatchStatusesCallback callbac
 
         callback.onReceived(downloadBatchStatuses);
     }
+
+    @Override
+    public void getFirstLocalPathForDownloadWithMatching(final String networkUri, final DownloadFilePathCallback callback) {
+        if (downloadService == null) {
+            executor.submit(
```
Mecharyry: Extracted a `Runnable` that knows to wait for a `lock` before performing a given action.

library/src/main/java/com/novoda/downloadmanager/DownloadManager.java
```diff
@@ -159,4 +145,32 @@ private void executeGetAllDownloadBatchStatuses(AllBatchStatusesCallback callbac
...
+                    WaitForLockRunnable.waitFor(waitForDownloadService)
+                            .thenPerform(new WaitForLockRunnable.Action() {
+                                @Override
+                                public void performAction() {
+                                    executeFirstLocalPathForDownloadWithMatching(networkUri, callback);
```
zegnus: shall we execute this back in the main thread?

library/src/main/java/com/novoda/downloadmanager/DownloadManager.java
```diff
@@ -122,34 +120,22 @@ public void removeDownloadBatchCallback(DownloadBatchCallback downloadBatchCallb
...
-    public void getAllDownloadBatchStatuses(AllBatchStatusesCallback callback) {
+    public void getAllDownloadBatchStatuses(final AllBatchStatusesCallback callback) {
         if (downloadService == null) {
-            ensureDownloadServiceExistsAndGetAllDownloadBatchStatuses(callback);
+            executor.submit(
+                    WaitForLockRunnable.waitFor(waitForDownloadService)
+                            .thenPerform(new WaitForLockRunnable.Action() {
+                                @Override
+                                public void performAction() {
+                                    executeGetAllDownloadBatchStatuses(callback);
```
zegnus: shall we execute this back in the main thread?

Mecharyry: Errrrrr....this wasn't done before? 😂

zegnus:
We'll create a separate task in order to put back the callbacks to the main thread as is done in the standard update/notification download callbacks

PR #288, opened 2018-01-05
Fix total file size

danybony:
### Problem
When initialising a new `DownloadFile` we use the `UNKNOWN` InternalFileSize. The problem is that _all_ new DownloadFiles used the same instance of InternalFileSize, which was updated with the total size of the first downloaded file, resulting in an inconsistent total batch size.

### Solution
Use a copy if the constant instance when initialising a new `InternalFileSize`

### Bonus
Show the `%` symbol in the demo activity logs because it was not clear to me that they were referring to the download percentage, at the beginning.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/InternalFileSizeCreator.java
+++ b/library/src/main/java/com/novoda/downloadmanager/InternalFileSizeCreator.java
@@ -10,7 +10,7 @@ final class InternalFileSizeCreator {
     }
 
     static InternalFileSize unknownFileSize() {
-        return UNKNOWN;
+        return UNKNOWN.copy();
     }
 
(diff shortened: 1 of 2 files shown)


library/src/main/java/com/novoda/downloadmanager/InternalFileSizeCreator.java
```diff
@@ -10,7 +10,7 @@ private InternalFileSizeCreator() {
     }
 
     static InternalFileSize unknownFileSize() {
-        return UNKNOWN;
+        return UNKNOWN.copy();
```
joetimmins: 👍

zegnus: wow, good one!! 👍

PR #289, opened 2018-01-05
Use java 8

Mecharyry:
## Problem
Too much boiler plate.

## Solution 
Use Java 8 and leverage lambdas.

## Paired with
@danybony

### DIFF
--- a/demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
+++ b/demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
@@ -16,10 +16,8 @@ import com.novoda.downloadmanager.DownloadBatchCallback;
 import com.novoda.downloadmanager.DownloadBatchId;
 import com.novoda.downloadmanager.DownloadBatchIdCreator;
 import com.novoda.downloadmanager.DownloadBatchStatus;
-import com.novoda.downloadmanager.DownloadFilePathCallback;
 import com.novoda.downloadmanager.DownloadMigrator;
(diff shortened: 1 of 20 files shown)


Mecharyry:
Checking with Java 7 demo

Mecharyry:
Turns out the demo or other client applications won't work unless they also support Java 8. Personally I don't think this should be an issue, it's the most recent version that is supported by ASide. Java 8 is backwards compatible so if client applications migrate to the latest they should have any problems running their applications.

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -116,144 +106,94 @@ public void onClick(View v) {
...
-    };
+    }
 
-    private final AllBatchStatusesCallback batchStatusesCallback = new AllBatchStatusesCallback() {
-        @Override
-        public void onReceived(List<DownloadBatchStatus> downloadBatchStatuses) {
-            for (DownloadBatchStatus downloadBatchStatus : downloadBatchStatuses) {
-                callback.onUpdate(downloadBatchStatus);
-            }
+    private final AllBatchStatusesCallback batchStatusesCallback = downloadBatchStatuses -> {
```
xrigau, line 177: I think it'd be nicer if this was a method and then use a method reference instead

xrigau: there's many places that are using fields instead of methods, so I'm fine either way if you don't want to make that many change

Issue #287, opened 2018-01-05
DownloadManager handler


Mecharyry:
## Problem
The `DownloadManager` class does not use a handler to post back to the `main thread` when performing any action 😬 

## Solution
Pass the `callbackHandler` from the `DownloadManagerBuilder` to the `DownloadManager` and ensure that all callbacks go through this `Handler`.

MechaRyRy:
💥 addressed.

PR #290, opened 2018-01-05
V2 download manager handler

Mecharyry:
## Problem 
As per issue #287 

> The DownloadManager class does not use a handler to post back to the main thread when performing any action

## Solution
Lambda it up 😄  make sure we post all callbacks to the main thread 😄 

## Paired with
@danybony


library/src/main/java/com/novoda/downloadmanager/DownloadManager.java
```diff
@@ -52,7 +59,7 @@ public void submitAllStoredDownloads(AllStoredDownloadsSubmittedCallback callbac
                 downloader.download(downloadBatch, downloadBatchMap);
             }
 
-            callback.onAllDownloadsSubmitted();
+            callbackHandler.post(callback::onAllDownloadsSubmitted);
```
zegnus, line 62: so simple now :)

MechaRyRy: Yea, it's awesome 😄

PR #291, opened 2018-01-05
Public Download File Status

MechaRyRy:
~DO NOT MERGE this is branched from #290~

## Problem
Tried importing v2 again today, turns out we need a tad more information to create a clients version of a `Download` 😬 

## Solution
Expose a publicly accessible version of `DownloadFileStatus` similar to how we expose a `DownloadBatchStatus`. We have created an `InternalDownloadFileStatus` and a `DownloadFileStatus` to hide the `download-manager` accessible methods and those that are for clients. 

## Paired with
@danybony


zegnus (approved):

Hutch4 (approved):
LGTM 👍

PR #293, opened 2018-01-09
Expose getters for batch

MechaRyRy:
### Problem
Sometimes a client needs to retrieve information on a `Batch`. Currently we do not expose the internals of the `Batch` class even though the class is public 🤔 

### Solution
Add getters for the `Batch` class.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/Batch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/Batch.java
@@ -15,15 +15,15 @@ public final class Batch {
         this.fileUrls = fileUrls;
     }
 
-    DownloadBatchId getDownloadBatchId() {
+    public DownloadBatchId downloadBatchId() {
         return downloadBatchId;
     }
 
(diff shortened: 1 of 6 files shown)


library/src/main/java/com/novoda/downloadmanager/Batch.java
```diff
@@ -15,15 +15,15 @@
         this.fileUrls = fileUrls;
     }
 
-    DownloadBatchId getDownloadBatchId() {
+    public DownloadBatchId downloadBatchId() {
```
MechaRyRy, line 18: @zegnus would you rather this class is interfaced out like the others and exposed only through a `Creator`? I found it a bit misleading that I could create this class but I can't use or access any information on it except to pass it to the `download-manager`.

danybony (approved):

zegnus:
After talking with @Mecharyry we see no need for exposing those fields, closing for now

PR #294, opened 2018-01-09
Expose download date

MechaRyRy:
### Problem
In v1 we were supporting the clients to know the `download-date`, we don't anymore 😱 

### Solution
Migrate `last_modified_timestamp` from v1 to v2 and expose in v2 for the clients.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
+++ b/library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
@@ -7,12 +7,14 @@ import java.util.List;
 
 class MigrationExtractor {
 
-    private static final String BATCHES_QUERY = "SELECT DISTINCT batches._id, batches.batch_title FROM batches INNER JOIN DownloadsByBatch ON "
-            + "DownloadsByBatch.batch_id = batches._id WHERE DownloadsByBatch.batch_total_bytes = DownloadsByBatch.batch_current_bytes";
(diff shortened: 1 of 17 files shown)


danybony (approved):

library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
```diff
@@ -7,14 +7,16 @@
 
 class MigrationExtractor {
 
-    private static final String BATCHES_QUERY = "SELECT DISTINCT batches._id, batches.batch_title FROM batches INNER JOIN DownloadsByBatch ON "
-            + "DownloadsByBatch.batch_id = batches._id WHERE DownloadsByBatch.batch_total_bytes = DownloadsByBatch.batch_current_bytes";
-    private static final String URIS_QUERY = "SELECT uri, _data, total_bytes FROM Downloads WHERE batch_id = ?";
+    private static final String BATCHES_QUERY = "SELECT batches._id, batches.batch_title FROM batches INNER JOIN DownloadsByBatch ON "
```
danybony, line 10: don't we need the `DISTINCT` anymore?

MechaRyRy: "GROUP BY batches._id" achieves the same thing and a little more obvious 😄

zegnus:
looks like its more useful for our current clients to use the timestamp from the table `batches` so that we can have just a creation date for the batch, instead of one per file. Also in the code you create one for the batch and reuse it for the files so this might be the way to go...

MechaRyRy:
Going to use batches instead.

PR #297, opened 2018-01-11
Make downloadBatchId and FileId consistent

MechaRyRy:
## Problem
I forget why but we thought that introducing a hash version of the id during the creation might be better 🤔  in retrospect it wasn't because what ends up happening is, you use this creator in multiple places but each time it creates a completely different object because you are hashing an already hashed object 😂 

## Solution
Just use a plan value, `String` was used for the `BatchId` and I like to be able to read the db when I look at it.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/Batch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/Batch.java
@@ -76,8 +76,8 @@ public final class Batch {
         }
 
         public Builder addFile(String fileUrl) {
-            String id = downloadBatchId.rawId() + fileUrl;
-            DownloadFileId downloadFileId = DownloadFileIdCreator.createFrom(id);
+            String rawId = downloadBatchId.rawId() + fileUrl;
+            DownloadFileId downloadFileId = DownloadFileIdCreator.createFrom(rawId);
             fileUrls.put(downloadFileId, fileUrl);
             return this;
         }
--- a/library/src/main/java/com/novoda/downloadmanager/Batch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/Batch.java
@@ -76,7 +76,7 @@ public final class Batch {
         }
 
         public Builder addFile(String fileUrl) {
-            String id = downloadBatchId.stringValue() + fileUrl;
+            String id = downloadBatchId.rawId() + fileUrl;
             DownloadFileId downloadFileId = DownloadFileIdCreator.createFrom(id);
             fileUrls.put(downloadFileId, fileUrl);
             return this;
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadFileId.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadFileId.java
@@ -2,5 +2,5 @@ package com.novoda.downloadmanager;
 
 public interface DownloadFileId {
 
-    String toRawId();
+    String rawId();
 }
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadFileIdCreator.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadFileIdCreator.java
@@ -7,6 +7,6 @@ public final class DownloadFileIdCreator {
     }
 
     public static DownloadFileId createFrom(String id) {
-        return new LiteDownloadFileId(id.hashCode());
+        return new LiteDownloadFileId(id);
     }
 }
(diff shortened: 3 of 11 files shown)


library/src/main/java/com/novoda/downloadmanager/LiteDownloadFileId.java
```diff
@@ -2,41 +2,40 @@
...
-    @Override
-    public String toRawId() {
-        return String.valueOf(id);
+        return id != null ? id.equals(that.id) : that.id == null;
     }
 
     @Override
     public int hashCode() {
-        return id;
+        return id != null ? id.hashCode() : 0;
```
MechaRyRy, line 32: This comes from outside of the library so it is completely feasible that this could be null.

library/src/main/java/com/novoda/downloadmanager/DownloadFileId.java
```diff
@@ -2,5 +2,5 @@
 
 public interface DownloadFileId {
 
-    String toRawId();
+    String rawId();
```
MechaRyRy, line 5: For consistency with the `DownloadBatchId`.

library/src/main/java/com/novoda/downloadmanager/LiteDownloadFileId.java
```diff
@@ -2,41 +2,40 @@
...
+    private final String id;
 
-    LiteDownloadFileId(int id) {
+    LiteDownloadFileId(String id) {
         this.id = id;
     }
 
+    @Override
+    public String rawId() {
+        return String.valueOf(id);
```
danybony: This is not needed anymore... now you can just return `id`

danybony (changes requested):
Good update, just one comment

library/src/main/java/com/novoda/downloadmanager/LiteDownloadFileId.java
MechaRyRy: c3e7df3

library/src/main/java/com/novoda/downloadmanager/LiteDownloadBatchId.java
```diff
@@ -18,19 +18,18 @@ public boolean equals(Object o) {
...
 
-        return id.equals(that.id);
-
+        return id != null ? id.equals(that.id) : that.id == null;
     }
 
     @Override
     public int hashCode() {
-        return id.hashCode();
+        return id != null ? id.hashCode() : 0;
```
MechaRyRy, line 32: This can also be created from outside and so can be `null`.

danybony (approved):

library/src/main/java/com/novoda/downloadmanager/DownloadFileIdCreator.java
```diff
@@ -7,6 +7,6 @@ private DownloadFileIdCreator() {
     }
 
     public static DownloadFileId createFrom(String id) {
-        return new LiteDownloadFileId(id.hashCode());
+        return new LiteDownloadFileId(id);
```
joetimmins: Don't these need to be unique? Does this change enforce that?

MechaRyRy: I don't believe these need to be unique for the `download-manager`, we don't really deal with Ids. The only time this is useful is when retrieving a `DownloadFileStatus`. If a client duplicates the ids then they will only receive the first matching `DownloadFileStatus`. Hashcode wouldn't have enforced uniqueness and broke the app because we were using the creator in multiple places for an id that was meant to remain constant.

library/src/main/java/com/novoda/downloadmanager/Batch.java
```diff
@@ -76,7 +76,7 @@ public Builder(DownloadBatchId downloadBatchId, String title) {
         }
 
         public Builder addFile(String fileUrl) {
-            String id = downloadBatchId.stringValue() + fileUrl;
+            String id = downloadBatchId.rawId() + fileUrl;
```
joetimmins: can we change this variable name to `rawId`? because you're creating an `id` from a variable called `id` right now. It would be clearer to create an `id` from a `rawId` imo

library/src/main/java/com/novoda/downloadmanager/DownloadFileIdCreator.java
joetimmins: You're right, I was thinking of the `filename`

library/src/main/java/com/novoda/downloadmanager/Batch.java
MechaRyRy: 8612e94

PR #298, opened 2018-01-12
Worker thread methods

MechaRyRy:
## Problem
All callbacks from the `DownloadManager` class callback on a `main-thread` handler 😬  this is affecting how some clients use the library and we should provide blocking methods where the client is responsible for defining a `worker-thread`.

## Solution
Introduce `WorkerThread` methods. Using a generic on the `WaitForDownloadServiceThenPerform` so it can be used by both the `worker-thread` methods and the original async ones that callback on the `main-thread`. The `WaitForDownloadServiceThenPerform` tidies the `DownloadManager` class up a little bit but at the caveat that we have removed the following from the async calls:

```
if (downloadService == null) {
            executor.submit(
                    WaitForLockRunnable.waitFor(waitForDownloadService)
                            .thenPerform(() -> executeFirstLocalPathForDownloadWithMatching(downloadFileId, callback))
            );
        } else {
            executeFirstLocalPathForDownloadWithMatching(downloadFileId, callback);
        }
```

with 

```
 executor.submit(() -> WaitForDownloadServiceThenPerform.<Runnable>waitFor(downloadService, waitForDownloadService)
                .thenPerform(() -> {
                    DownloadFileStatus downloadFileStatus = executeGetDownloadStatusWithMatching(downloadFileId);
                    callbackHandler.post(() -> callback.onReceived(downloadFileStatus));
                    return null;
                }));
```
This means if the service is not null we will NOT bypass the call to the executor. I opted for this rather than maintaining the check for the service, it's consistent and makes the code much easier to read IMHO.

## Tests
Updated the `DownloadManagerTests` to test the new `WorkerThread` methods and to ensure that the new `WaitForDownloadServiceThenPerform` doesn't break the previous functionality.


library/src/main/java/com/novoda/downloadmanager/DownloadManager.java
```diff
@@ -123,48 +125,60 @@ public void removeDownloadBatchCallback(DownloadBatchCallback downloadBatchCallb
...
+            if (downloadFileStatus != null) {
+                return downloadFileStatus;
             }
         }
+        return null;
+    }
+
+    @Override
+    public void getDownloadStatusWithMatching(DownloadFileId downloadFileId, DownloadFileStatusCallback callback) {
+        executor.submit((Runnable) () -> WaitForDownloadServiceThenPerform.<Void>waitFor(downloadService, waitForDownloadService)
```
MechaRyRy, line 176: I can use `Callable` which removes the `Runnable` cast but then I need to remove a warning for an exceptional return value that is not used 😬  so ANNOYING!

library/src/main/java/com/novoda/downloadmanager/WaitForDownloadServiceThenPerform.java
```diff
@@ -0,0 +1,52 @@
+package com.novoda.downloadmanager;
+
+import android.support.annotation.Nullable;
+
+import com.novoda.notils.logger.simple.Log;
+
+final class WaitForDownloadServiceThenPerform {
```
MechaRyRy, line 7: Should be able to use this in other places like the `LiteDownloadManagerDownloader` which should improve the readability and maintainability a little bit!

library/src/main/java/com/novoda/downloadmanager/WaitForDownloadServiceThenPerform.java
```diff
@@ -0,0 +1,52 @@
...
+
+        private final DownloadService downloadService;
+        private final Object downloadServiceLock;
+
+        WaitForDownloadServiceThenPerformAction(DownloadService downloadService, Object downloadServiceLock) {
+            this.downloadService = downloadService;
+            this.downloadServiceLock = downloadServiceLock;
+        }
+
+        T thenPerform(final Action<T> action) {
```
danybony: does it need to be final?

library/src/main/java/com/novoda/downloadmanager/WaitForDownloadServiceThenPerform.java
```diff
@@ -0,0 +1,52 @@
+package com.novoda.downloadmanager;
+
+import android.support.annotation.Nullable;
+
+import com.novoda.notils.logger.simple.Log;
+
+final class WaitForDownloadServiceThenPerform {
+
+    interface Action<T> {
```
joetimmins, line 9: I don't think we need this - we can use Java 8's `Supplier` functional interface

library/src/main/java/com/novoda/downloadmanager/DownloadManager.java
```diff
@@ -123,48 +125,60 @@ public void removeDownloadBatchCallback(DownloadBatchCallback downloadBatchCallb
...
-        if (downloadService == null) {
-            executor.submit(
-                    WaitForLockRunnable.waitFor(waitForDownloadService)
-                            .thenPerform(() -> executeFirstLocalPathForDownloadWithMatching(downloadFileId, callback))
-            );
-        } else {
-            executeFirstLocalPathForDownloadWithMatching(downloadFileId, callback);
-        }
+    public DownloadFileStatus getDownloadStatusWithMatching(DownloadFileId downloadFileId) {
+        return WaitForDownloadServiceThenPerform.<DownloadFileStatus>waitFor(downloadService, waitForDownloadService)
```
danybony, line 158: How does it look adding a static import for WaitForDownloadServiceThenPerform.waitFor()?

library/src/main/java/com/novoda/downloadmanager/WaitForDownloadServiceThenPerform.java
joetimmins: Hang on yeah we do, proper Java 8 support requires API 24 :(

zegnus (approved):

danybony (commented):
Nice! Just a couple of questions

library/src/main/java/com/novoda/downloadmanager/WaitForDownloadServiceThenPerform.java
MechaRyRy: `Supplier` requires level 24 😬

library/src/main/java/com/novoda/downloadmanager/DownloadManager.java
MechaRyRy: I don't think you can add a static because of the generic type.

library/src/main/java/com/novoda/downloadmanager/WaitForDownloadServiceThenPerform.java
MechaRyRy: Ah no, it doesn't 😄

MechaRyRy: ba1cb36

joetimmins (approved):
👍

PR #299, opened 2018-01-12
Downloader then perform

MechaRyRy:
## Problem
Introduced a `WaitForDownloadServiceThenPerform` that collects the `service` waiting logic into a single class. Turns out we can use this on the `LiteDownloadManagerDownloader` too 😄 

## Solution
Use the `WaitForDownloadServiceThenPerform` on the `LiteDownloadManagerDownloader` 😄  Also rename the `WaitForDownloadServiceThenPerform` to `WaitForDownloadService` so the full command will appear:

```
        executor.submit(() -> WaitForDownloadService.<Void>waitFor(downloadService, waitForDownloadService)
                .thenPerform(executeDownload(downloadBatch)));
```

This assumes the action is extracted to another method which will now appear as:

```
    private WaitForDownloadService.ThenPerform.Action<Void> executeDownload(DownloadBatch downloadBatch) {
        return () -> {
            InternalDownloadBatchStatus downloadBatchStatus = downloadBatch.status();
            if (downloadBatchStatus.status() != DOWNLOADED) {
                updateStatusToQueuedIfNeeded(downloadBatchStatus);
                downloadService.download(downloadBatch, downloadBatchCallback());
            }
            return null;
        };
    }
```

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerDownloader.java
+++ b/library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerDownloader.java
@@ -68,11 +68,11 @@ class LiteDownloadManagerDownloader {
 
     public void download(DownloadBatch downloadBatch, Map<DownloadBatchId, DownloadBatch> downloadBatchMap) {
         downloadBatchMap.put(downloadBatch.getId(), downloadBatch);
-        executor.submit(() -> WaitForDownloadServiceThenPerform.<Void>waitFor(downloadService, waitForDownloadService)
+        executor.submit(() -> WaitForDownloadService.<Void>waitFor(downloadService, waitForDownloadService)
                 .thenPerform(executeDownload(downloadBatch)));
     }
 
-    private WaitForDownloadServiceThenPerform.Action<Void> executeDownload(DownloadBatch downloadBatch) {
+    private WaitForDownloadService.ThenPerform.Action<Void> executeDownload(DownloadBatch downloadBatch) {
         return () -> {
             InternalDownloadBatchStatus downloadBatchStatus = downloadBatch.status();
             if (downloadBatchStatus.status() != DOWNLOADED) {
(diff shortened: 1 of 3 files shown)


library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerDownloader.java
```diff
@@ -69,38 +68,19 @@ public void download(Batch batch, Map<DownloadBatchId, DownloadBatch> downloadBa
...
-        }
-    }
-
-    private void executeDownload(final DownloadBatch downloadBatch) {
-        InternalDownloadBatchStatus downloadBatchStatus = downloadBatch.status();
-        if (downloadBatchStatus.status() != DOWNLOADED) {
-            updateStatusToQueuedIfNeeded(downloadBatchStatus);
-            downloadService.download(downloadBatch, downloadBatchCallback());
-        }
+            return null;
```
MechaRyRy, line 82: `returns` null because of the callback, shown with the `Action<Void>`

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerDownloader.java
```diff
@@ -69,38 +68,19 @@ public void download(Batch batch, Map<DownloadBatchId, DownloadBatch> downloadBa
 
     public void download(DownloadBatch downloadBatch, Map<DownloadBatchId, DownloadBatch> downloadBatchMap) {
         downloadBatchMap.put(downloadBatch.getId(), downloadBatch);
-        if (downloadService == null) {
-            ensureDownloadServiceExistsAndDownload(downloadBatch);
-        } else {
-            executeDownload(downloadBatch);
-        }
+        executor.submit(() -> WaitForDownloadService.<Void>waitFor(downloadService, waitForDownloadService)
```
zegnus, line 71: fine for me, its exactly the same as in the DownloadManager with exactly the same objects because they are shared 👍 

This submission won't block this current executor because it submits into a dedicated download thread on the Service itself. So this submits and ends fast.

zegnus (approved):

danybony (approved):

Issue #300, opened 2018-01-12
Delete status updates not occurring


MechaRyRy:
## Steps to reproduce
- Download an asset using the download manager sample app.
- Once complete press the delete all button
**Expected:** Statuses show as deleted **Actual:** Statuses show as deleted 💯 

- Download an asset using the download manager sample app.
- Once complete swipe away the app from recents killing the service etc.
- Restart the app, you should see the last status of download complete.
- Press the delete button
**Expected:** Should show status as being deleted **Actual:** Status shows previous state of complete

## Solution 
When creating the `download-manager` from scratch we retrieve all of the batches from the db and then call `download` on all of the downloads that have not already been downloaded. The `download` attaches callbacks on the batch which you will receive when calling `LiteDownloadManagerCommands.addDownloadBatchCallback(DownloadBatchCallback downloadBatchCallback);`. Here's the thing though, in the `LiteDownloadManagerDownloader` we check for whether a downloaded is downloaded before setting up these callbacks. Downloaded? No callbacks for you.

Should be a relatively easy fix supplying the callback regardless of state.

PR #301, opened 2018-01-15
Delete status update

MechaRyRy:
### Problem
As per issue #300, when a batch is in the `DOWNLOADED` state and the app is cleared from memory (Removed from recents), the batch no longer has an attached `StatusCallback`. No `StatusCallback` means no updates when the `DOWNLOADED` asset is `DELETED`.

### Solution
Always attach the `StatusCallback` regardless of state. Ensure that we prevent certain events from occurring when in the `DOWNLOADED` state:

- PAUSE
- RESUME
- DOWNLOAD

All of the above triggering will change the state and then change it back immediately to `DOWNLOADED` resulting in an additional set of complete notifications being sent to the user.

### Screen Capture

Before | After
--- | ---
deleted_before | deleted_after

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/LiteDownloadService.java
+++ b/library/src/main/java/com/novoda/downloadmanager/LiteDownloadService.java
@@ -13,6 +13,8 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
+import static com.novoda.downloadmanager.DownloadBatchStatus.Status.DOWNLOADED;
+
 public class LiteDownloadService extends Service implements DownloadService {
 
     private static final long TEN_MINUTES_IN_MILLIS = TimeUnit.MINUTES.toMillis(10);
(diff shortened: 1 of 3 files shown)


joetimmins (approved):
EZPZ 🍋

library/src/main/java/com/novoda/downloadmanager/LiteDownloadService.java
```diff
@@ -46,10 +48,14 @@ public IBinder onBind(Intent intent) {
 
     @Override
     public void download(final DownloadBatch downloadBatch, final DownloadBatchCallback callback) {
-        callback.onUpdate(downloadBatch.status());
```
zegnus, line 49: are you sure about this? if we get here and the status is downloaded, the callback won't be notified...

MechaRyRy: Interesting, does this only work because of the `getAllBatchStatuses`... I'll take another look 💯

danybony (commented):
I have the same doubt as @zegnus , but a part from that 💯

MechaRyRy:
Still unsure how to fix this issue @zegnus, working towards it.

PR #302, opened 2018-01-15
Remove completed notification on delete

### LINKED ISSUE #300: Delete status updates not occurring

MechaRyRy:
### Problem
When trying to fix #300 I found that when downloads complete in the sample app and a user taps delete the notifications are still persisted, even though they don't exist anymore 😬 

### Solution
Make sure that we remove a notification that corresponds to a download that is deleted. 

I have created a `NotificationDispatcher` to encapsulate the logic related to sending notifications in one class rather than continue to include it in the already bloated `Downloader` class. 

Send a `NotificationInformation` when performing a dismiss on notifications so we can have the notification manager remove a specific notification. This would be the notifications not bound to the service anymore.

### Screen Capture

Before | After
--- | ---
delete_notifications_before | delete_notification_after

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
@@ -230,6 +230,8 @@ public final class DownloadManagerBuilder {
             notificationManager.createNotificationChannel(notificationChannel);
         }
 
+        NotificationDispatcher notificationDispatcher = new NotificationDispatcher(LOCK, notificationCreator);
+
         LiteDownloadManagerDownloader downloader = new LiteDownloadManagerDownloader(
                 LOCK,
                 EXECUTOR,
@@ -237,7 +239,7 @@ public final class DownloadManagerBuilder {
                 fileOperations,
                 downloadsBatchPersistence,
                 downloadsFilePersistence,
-                notificationCreator,
+                notificationDispatcher,
                 callbacks,
                 callbackThrottleCreator
         );
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
@@ -285,20 +285,36 @@ public final class DownloadManagerBuilder {
 
         @Override
         public Notification customNotificationFrom(NotificationCompat.Builder builder, DownloadBatchStatus payload) {
+            if (payload.status() == DownloadBatchStatus.Status.DELETION) {
+                return createDeletedNotification(builder, payload);
+            } else {
+                DownloadBatchTitle downloadBatchTitle = payload.getDownloadBatchTitle();
+                int percentageDownloaded = payload.percentageDownloaded();
+                int bytesFileSize = (int) payload.bytesTotalSize();
+                int bytesDownloaded = (int) payload.bytesDownloaded();
+                String title = downloadBatchTitle.asString();
+                String content = percentageDownloaded + "% downloaded";
+
+                return builder
+                        .setProgress(bytesFileSize, bytesDownloaded, NOT_INDETERMINATE)
+                        .setSmallIcon(notificationIcon)
+                        .setContentTitle(title)
+                        .setContentText(content)
+                        .build();
+
+            }
+        }
+
+        private Notification createDeletedNotification(NotificationCompat.Builder builder, DownloadBatchStatus payload) {
             DownloadBatchTitle downloadBatchTitle = payload.getDownloadBatchTitle();
-            int percentageDownloaded = payload.percentageDownloaded();
-            int bytesFileSize = (int) payload.bytesTotalSize();
-            int bytesDownloaded = (int) payload.bytesDownloaded();
             String title = downloadBatchTitle.asString();
-            String content = percentageDownloaded + "% downloaded";
+            String content = "Deleted";
 
             return builder
-                    .setProgress(bytesFileSize, bytesDownloaded, NOT_INDETERMINATE)
                     .setSmallIcon(notificationIcon)
                     .setContentTitle(title)
                     .setContentText(content)
                     .build();
-
         }
     }
 }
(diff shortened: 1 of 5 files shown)


library/src/main/java/com/novoda/downloadmanager/NotificationDispatcher.java
```diff
@@ -0,0 +1,45 @@
...
+
+    private DownloadService downloadService;
+
+    NotificationDispatcher(Object waitForDownloadService, NotificationCreator<DownloadBatchStatus> notificationCreator) {
+        this.waitForDownloadService = waitForDownloadService;
+        this.notificationCreator = notificationCreator;
+    }
+
+    void updateNotification(DownloadBatchStatus downloadBatchStatus) {
+        WaitForDownloadService.<Void>waitFor(downloadService, waitForDownloadService)
```
zegnus: Is this a blocking call?

library/src/main/java/com/novoda/downloadmanager/NotificationDispatcher.java
```diff
@@ -0,0 +1,45 @@
...
+                .thenPerform(executeUpdateNotification(downloadBatchStatus));
+    }
+
+    private WaitForDownloadService.ThenPerform.Action<Void> executeUpdateNotification(DownloadBatchStatus downloadBatchStatus) {
+        return () -> {
+            NotificationInformation notificationInformation = notificationCreator.createNotification(downloadBatchStatus);
+
+            if (downloadBatchStatus.status() == DELETION) {
```
jszmltr: How about : 
```
if () {
} else if (){
} else {
}
return null;
```

library/src/main/java/com/novoda/downloadmanager/NotificationDispatcher.java
MechaRyRy: It's only blocking the current thread if the `download-service` doesn't exist. Given that the `Downloader` would also be blocked I don't think this is an issue.

library/src/main/java/com/novoda/downloadmanager/NotificationDispatcher.java
MechaRyRy: I prefer the quickness and readability of the `if -> return` 😄  I guess its a preference thing!

jszmltr: What I don't like in this scenario that `else` is not really highlighted enough. But don't mind really.

library/src/main/java/com/novoda/downloadmanager/NotificationDispatcher.java
MechaRyRy: Ah right sorry. This is blocking yes. The `Downloader` waits for the service before performing the `download` on the `DownloadBatch` the batch then calls through to this dispatcher. So from what I understand this will all be on the `executor`. Correct me if I am wrong.

library/src/main/java/com/novoda/downloadmanager/NotificationDispatcher.java
MechaRyRy: 6ffb008

MechaRyRy:
@zegnus Updated the deleted notification rather than just removing. This means that there is an unused method on the `Service`, maybe this should be customisable by the client?
deleted

zegnus:
@Mecharyry Regarding offering a custom notification for just the deleted ones, the current API provides already a callback for providing a notification, is a client wants to customise it they can just do it providing now. We can always extend it and provide customisations for just the downloaded and the deleted statuses but is good enough as it is now in my opinion (we do not restrict functionalities and we are still flexible enough)

library/src/main/java/com/novoda/downloadmanager/NotificationDispatcher.java
zegnus: Yes you are right, shall we use the working thread annotation? ;)

library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
```diff
@@ -283,20 +285,36 @@ private CallbackThrottleCreator getCallbackThrottleCreator(CallbackThrottleCreat
...
-            int percentageDownloaded = payload.percentageDownloaded();
-            int bytesFileSize = (int) payload.bytesTotalSize();
-            int bytesDownloaded = (int) payload.bytesDownloaded();
             String title = downloadBatchTitle.asString();
-            String content = percentageDownloaded + "% downloaded";
+            String content = "Deleted";
 
             return builder
-                    .setProgress(bytesFileSize, bytesDownloaded, NOT_INDETERMINATE)
                     .setSmallIcon(notificationIcon)
```
zegnus: This code is also reused in the else, shall we try to reuse as much as possible?

MechaRyRy: Can do 👍

zegnus:
@Mecharyry sorry you were right, I missed the `stack` notification. Yeah we should add it as part of the customisation otherwise there is no way that the client can modify that. Let's add a task for it.

library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
MechaRyRy: 2b040c5

library/src/main/java/com/novoda/downloadmanager/NotificationDispatcher.java
MechaRyRy: 0c35ee2

zegnus (approved):

PR #303, opened 2018-01-15
Delete status update v2

MechaRyRy:
This is the second attempt of #301 😄 

### Problem
As per issue #300, when a batch is in the `DOWNLOADED` state and the app is cleared from memory (Removed from recents), the batch no longer has an attached `StatusCallback`. No `StatusCallback` means no updates when the `DOWNLOADED` asset is `DELETED`.

### Solution
Always attach the `StatusCallback` regardless of state. Ensure that we prevent certain events from occurring when in the `DOWNLOADED` state:

- PAUSE
- RESUME
- DOWNLOAD

All of the above triggering will change the state and then change it back immediately to `DOWNLOADED` resulting in an additional set of complete notifications being sent to the user.

Seems that `Object.hashcode` is a little broken 😬  it gave different values whenever a notification was created. For `DownloadBatchStatus` notifications we now use `DownloadBatchId.hashcode` which has resolved the double notification problem when the app is killed.

### Screen Capture

Before | After
--- | ---
deleted_before | deleted_after

### DIFF
--- a/demo-simple/src/main/java/com/novoda/downloadmanager/demo/CustomCallbackThrottle.java
+++ b/demo-simple/src/main/java/com/novoda/downloadmanager/demo/CustomCallbackThrottle.java
@@ -2,16 +2,16 @@ package com.novoda.downloadmanager.demo;
 
 import com.novoda.notils.logger.simple.Log;
 import com.novoda.downloadmanager.CallbackThrottle;
-import com.novoda.downloadmanager.DownloadBatchCallback;
+import com.novoda.downloadmanager.DownloadBatchStatusCallback;
 import com.novoda.downloadmanager.DownloadBatchStatus;
 
 // Must be public
 public class CustomCallbackThrottle implements CallbackThrottle {
 
-    private DownloadBatchCallback callback;
+    private DownloadBatchStatusCallback callback;
 
     @Override
-    public void setCallback(DownloadBatchCallback callback) {
+    public void setCallback(DownloadBatchStatusCallback callback) {
         Log.v("setCallback");
         this.callback = callback;
     }
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
@@ -25,7 +25,7 @@ class DownloadBatch {
     private final CallbackThrottle callbackThrottle;
 
     private long totalBatchSizeBytes;
-    private DownloadBatchCallback callback;
+    private DownloadBatchStatusCallback callback;
 
     DownloadBatch(InternalDownloadBatchStatus internalDownloadBatchStatus,
                   List<DownloadFile> downloadFiles,
@@ -39,7 +39,7 @@ class DownloadBatch {
         this.callbackThrottle = callbackThrottle;
     }
 
-    void setCallback(DownloadBatchCallback callback) {
+    void setCallback(DownloadBatchStatusCallback callback) {
         this.callback = callback;
         callbackThrottle.setCallback(callback);
     }
@@ -50,6 +50,10 @@ class DownloadBatch {
         }
 
         if (downloadBatchStatus.status() == DELETION) {
+            return;
+        }
+
+        if (downloadBatchStatus.status() == DOWNLOADED) {
             return;
         }
 
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
@@ -45,15 +45,16 @@ class DownloadBatch {
     }
 
     void download() {
-        if (downloadBatchStatus.status() == PAUSED) {
+        DownloadBatchStatus.Status status = downloadBatchStatus.status();
+        if (status == PAUSED) {
             return;
         }
 
-        if (downloadBatchStatus.status() == DELETION) {
+        if (status == DELETION) {
             return;
         }
 
-        if (downloadBatchStatus.status() == DOWNLOADED) {
+        if (status == DOWNLOADED) {
             return;
         }
 
@@ -92,7 +93,7 @@ class DownloadBatch {
             DownloadsNetworkRecoveryCreator.getInstance().scheduleRecovery();
         }
 
-        if (downloadBatchStatus.status() == DOWNLOADED) {
+        if (status == DOWNLOADED) {
             downloadBatchStatus.markAsDownloaded(downloadsBatchPersistence);
         }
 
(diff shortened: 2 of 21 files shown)


demo-simple/src/main/java/com/novoda/downloadmanager/demo/CustomCallbackThrottle.java
```diff
@@ -2,16 +2,16 @@
...
 import com.novoda.downloadmanager.CallbackThrottle;
-import com.novoda.downloadmanager.DownloadBatchCallback;
+import com.novoda.downloadmanager.DownloadBatchStatusCallback;
 import com.novoda.downloadmanager.DownloadBatchStatus;
 
 // Must be public
 public class CustomCallbackThrottle implements CallbackThrottle {
 
-    private DownloadBatchCallback callback;
+    private DownloadBatchStatusCallback callback;
```
MechaRyRy, line 11: Sorry, the fact that it handled `DownloadBatchStatus` but was called a `DownloadBatchCallback` was irritating me 😄

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -136,7 +140,7 @@ private long getTotalSize(List<DownloadFile> downloadFiles) {
     }
 
     void pause() {
-        if (downloadBatchStatus.status() == PAUSED) {
+        if (downloadBatchStatus.status() == PAUSED || downloadBatchStatus.status() == DOWNLOADED) {
```
MechaRyRy: This prevents pause being called when the asset is already `DOWNLOADED`. The same for `resume` below.

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerDownloader.java
```diff
@@ -74,23 +74,21 @@ public void download(DownloadBatch downloadBatch, Map<DownloadBatchId, DownloadB
...
-            }
+            updateStatusToQueuedIfNeeded(downloadBatchStatus);
+            downloadService.download(downloadBatch, downloadBatchCallback());
             return null;
         };
     }
 
     private void updateStatusToQueuedIfNeeded(InternalDownloadBatchStatus downloadBatchStatus) {
-        if (downloadBatchStatus.status() != PAUSED) {
+        if (downloadBatchStatus.status() != PAUSED && downloadBatchStatus.status() != DOWNLOADED) {
```
MechaRyRy, line 84: No need to queue something that is already `DOWNLOADED`.

library/src/main/java/com/novoda/downloadmanager/NotificationCreator.java
```diff
@@ -32,6 +32,10 @@ NotificationInformation createNotification(final T notificationPayload) {
         return new NotificationInformation() {
             @Override
             public int getId() {
+                if (notificationPayload instanceof DownloadBatchStatus) {
+                    return ((DownloadBatchStatus) notificationPayload).getDownloadBatchId().hashCode();
```
MechaRyRy, line 36: `Object.hashcode` is returning different values each time it is executed. Adding `hashcode` to the whole `DownloadBatchStatus` lags the UI hard because it needs to calculate this for each percentage point on a large object, hence doing it on the id only.

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -53,6 +53,10 @@ void download() {
             return;
         }
 
+        if (downloadBatchStatus.status() == DOWNLOADED) {
```
zegnus: shall we use an `or` now that we have three of them? :)

demo-simple/src/main/java/com/novoda/downloadmanager/demo/CustomCallbackThrottle.java
jszmltr: * was called `DownloadBatchCallback ` (so you have renamed it to `DownloadBatchStatusCallback `)

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
zegnus: Yeah good one, this solves lots of issues when the library is used without checking current download statuses, as pausing a paused download

jszmltr: Extract `downloadBrachStatus.status()` to a variable 😅  

Also applicable for `resume()` method.

jszmltr (approved):
No blockers for me 👍 💯

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerDownloader.java
```diff
@@ -74,23 +74,21 @@ public void download(DownloadBatch downloadBatch, Map<DownloadBatchId, DownloadB
     private WaitForDownloadService.ThenPerform.Action<Void> executeDownload(DownloadBatch downloadBatch) {
         return () -> {
             InternalDownloadBatchStatus downloadBatchStatus = downloadBatch.status();
-            if (downloadBatchStatus.status() != DOWNLOADED) {
-                updateStatusToQueuedIfNeeded(downloadBatchStatus);
-                downloadService.download(downloadBatch, downloadBatchCallback());
-            }
+            updateStatusToQueuedIfNeeded(downloadBatchStatus);
+            downloadService.download(downloadBatch, downloadBatchCallback());
```
zegnus, line 78: You're calling this now even if its downloaded, I assume that's ok because it won't be queued and because there is a check in the `download` that will abort and invoke the callback immediately right?

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
MechaRyRy: 3d12073

library/src/main/java/com/novoda/downloadmanager/NotificationCreator.java
zegnus: If you did this `.hashCode` at point of `DownloadBatchStatus` creation and you would extend that interface with an `int notificationId()`, would that solve all performance issues here and prevent the casting?

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerDownloader.java
MechaRyRy: That's correct 😄

library/src/main/java/com/novoda/downloadmanager/NotificationCreator.java
zegnus: let's think of a solution by which we do not have to calculate the hash every single time, being a known constant value, next PR

zegnus (approved):

PR #306, opened 2018-01-19
Ensure that error notifications are customised and swipeable

MechaRyRy:
### Problem
When customising notifications in a client application we noticed that error notifications are not swipeable. They are not customised in the sample either, so they appear as regular notifications, without any error text 😬 

### Solution
Ensure that the `ERROR` status resolves into a swipeable notification.

### Further work
This is a quick fix for now, we plan to allow clients to specify themselves which notifications are swipeable and which are not, in a future PR. 

### Screen Capture

Before | After
--- | ---
before_error | after_error

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
@@ -292,6 +292,8 @@ public final class DownloadManagerBuilder {
 
             if (payload.status() == DownloadBatchStatus.Status.DELETION) {
                 return createDeletedNotification(builder);
+            } else if (payload.status() == DownloadBatchStatus.Status.ERROR) {
+                return createErrorNotification(builder, payload);
             } else {
                 return createProgressNotification(builder, payload);
             }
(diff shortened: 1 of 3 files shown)


library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
```diff
@@ -304,6 +306,13 @@ private Notification createDeletedNotification(NotificationCompat.Builder builde
                     .build();
         }
 
+        private Notification createErrorNotification(NotificationCompat.Builder builder, DownloadBatchStatus downloadBatchStatus) {
+            String content = "Error: " + downloadBatchStatus.getDownloadErrorType();
```
danybony, line 310: Shouldn't these come from resources?

danybony (approved):
💯  for the fix, just one question not strictly related to this PR

library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
MechaRyRy: Probably but we haven't defined them 😬

MechaRyRy: I'll add a ticket in issues

MechaRyRy: https://github.com/novoda/download-manager/issues/307

danybony: 💯

xrigau (approved):
👍

PR #308, opened 2018-01-23
Absolute path for internal storage

danybony:
### Problem
We had a discrepancy about how we were creating `FilePath`s in the different file persistences:
* in the **external** one we were storing the whole absolute path (`/storage/emulated/0/something/something/file_name.ext`)
* in the **internal** one we were storing only the file name relative to the internal storage dir (`file_name.ext`)

This was done because when accessing files using `context.openFileOutput(path)` the param string cannot contain path separator chars. The problem was that the user didn't have knowledge about if the file was stored in internal or external memory when retrieving info, and this also broke the possibility of accessing the file using `file://` URI.

### Solution
Store the entire absolute path when creating a FilePath in the internal file persistence
(`/data/user/0/app.package.name/files/file_name.ext`)

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/InternalFilePersistence.java
+++ b/library/src/main/java/com/novoda/downloadmanager/InternalFilePersistence.java
@@ -38,7 +38,7 @@ class InternalFilePersistence implements FilePersistence {
 
         File internalFileDir = context.getFilesDir();
 
-        String absolutePath = internalFileDir.getAbsolutePath() + File.separatorChar + fileName.name();
+        String absolutePath = new File(internalFileDir, fileName.name()).getAbsolutePath();
         FilePath filePath = FilePathCreator.create(absolutePath);
         return create(filePath);
     }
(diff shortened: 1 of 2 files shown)


zegnus (approved):

library/src/main/java/com/novoda/downloadmanager/InternalFilePersistence.java
```diff
@@ -35,23 +36,30 @@ public FilePersistenceResult create(FileName fileName, FileSize fileSize) {
             return FilePersistenceResult.newInstance(Status.ERROR_INSUFFICIENT_SPACE);
         }
 
-        FilePath filePath = FilePathCreator.create(fileName.name());
+        File internalFileDir = context.getFilesDir();
+
+        String absolutePath = internalFileDir.getAbsolutePath() + File.separatorChar + fileName.name();
```
tasomaniac: What about `new File(internalFileDir, fileName.name()).getAbsolutePath()`?

I think we should not create paths by concatenating strings

tasomaniac (approved):
Just a small comment. Looks good.

library/src/main/java/com/novoda/downloadmanager/InternalFilePersistence.java
danybony: Good point, done in 3074d45

PR #309, opened 2018-01-23
Allow custom DB name for migration

danybony:
### Problem
Some client apps might have custom DB names, which will not be found when migrating from v1 to vs (we were hardcoding the filename `downloads.db`)

### Solution
Extract db name as parameter in migration

### DIFF
--- a/demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
+++ b/demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
@@ -119,7 +119,7 @@ public class MainActivity extends AppCompatActivity {
 
     private final MigrationCallback migrationCallback = migrationStatus -> databaseMigrationUpdates.setText(migrationStatus.status().toRawValue());
 
-    private final View.OnClickListener startMigrationOnClick = v -> downloadMigrator.startMigration(migrationCallback);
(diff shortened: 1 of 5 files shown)


MechaRyRy (approved):
EZPZ 🍋

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -119,7 +119,7 @@ protected void onCreate(Bundle savedInstanceState) {
 
     private final MigrationCallback migrationCallback = migrationStatus -> databaseMigrationUpdates.setText(migrationStatus.status().toRawValue());
 
-    private final View.OnClickListener startMigrationOnClick = v -> downloadMigrator.startMigration(migrationCallback);
+    private final View.OnClickListener startMigrationOnClick = v -> downloadMigrator.startMigration("downloads.db", migrationCallback);
```
zegnus, line 122: is this param mandatory?

danybony: Yes, we need to specify the db filename

danybony: I saw that in the v1 of the download manager we were using the metadata `com.novoda.downloadmanager.DatabaseFilename` to specify from a client perspective the db filename... we might want to use that one even now, in that case replacing the DB name param here

PR #311, opened 2018-01-26
v2 file naming

MechaRyRy:
### Problem
We had a problem in a client application where one download file was used to reference other download files. It references by using its path as the base and then applying an additional path to it to get the supporting files 😬  Currently in v2 we use the hash for a batch and the url for the filename. The filename is also used to derive the path for the downloaded file. 

### Solution
Use a `String` for the `FileName`, this way it matches what a user would expect to see when downloading a file from a given endpoint. 

Give clients the option to specify a relative path. The path is relative to whatever the base persistence does, for internal this will be added to the `data/data/package/files` path.

### Further work
- We need to double check `MigrationJob` to ensure that it works with this mechanism.
- We can probably alter the `Batch` builder to make this a little more user friendly. We could introduce a first class collection for example.


demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -123,7 +124,7 @@ protected void onCreate(Bundle savedInstanceState) {
 
     private final View.OnClickListener downloadBatchesOnClick = v -> {
         Batch batch = new Batch.Builder(BATCH_ID_1, "Made in chelsea")
-                .addFile(FILE_ID_1, FIVE_MB_FILE_URL)
+                .addFile(FILE_ID_1, FIVE_MB_FILE_URL, FilePathCreator.create("/foo/bar/10mb.zip"))
```
danybony: Looking at this I'm wondering... does the client still care about the file ID at this point?

library/src/main/java/com/novoda/downloadmanager/Batch.java
```diff
@@ -7,12 +7,12 @@
 
     private final DownloadBatchId downloadBatchId;
     private final String title;
-    private final Map<DownloadFileId, String> fileUrls;
+    private final Map<DownloadFileId, NetworkAddressAndFilePath> networkAddressAndFileNameById;
```
danybony: networkAddressAndFilePathById

library/src/main/java/com/novoda/downloadmanager/Batch.java
```diff
@@ -7,12 +7,12 @@
...
 
-    Batch(DownloadBatchId downloadBatchId, String title, Map<DownloadFileId, String> fileUrls) {
+    Batch(DownloadBatchId downloadBatchId, String title, Map<DownloadFileId, NetworkAddressAndFilePath> networkAddressAndFileNameById) {
```
danybony: networkAddressAndFilePathById

library/src/main/java/com/novoda/downloadmanager/Batch.java
```diff
@@ -23,8 +23,8 @@ String getTitle() {
         return title;
     }
 
-    Map<DownloadFileId, String> getFileUrls() {
-        return new HashMap<>(fileUrls);
+    Map<DownloadFileId, NetworkAddressAndFilePath> networkAddressAndFileNameById() {
```
danybony: networkAddressAndFilePathById()

library/src/main/java/com/novoda/downloadmanager/Batch.java
```diff
@@ -60,15 +61,15 @@ public String toString() {
...
+                + ", networkAddressAndFileNameById=" + networkAddressAndFileNameById
                 + '}';
     }
 
     public static class Builder {
 
         private final DownloadBatchId downloadBatchId;
         private final String title;
-        private final Map<DownloadFileId, String> fileUrls = new HashMap<>();
+        private final Map<DownloadFileId, NetworkAddressAndFilePath> networkAddressAndFileNameById = new HashMap<>();
```
danybony: networkAddressAndFilePathById

danybony: Well, you got it... please check also all the local variables below 👅

library/src/main/java/com/novoda/downloadmanager/Batch.java
```diff
@@ -78,17 +79,26 @@ public Builder(DownloadBatchId downloadBatchId, String title) {
...
         public Builder addFile(DownloadFileId downloadFileId, String fileUrl) {
-            fileUrls.put(downloadFileId, fileUrl);
+            NetworkAddressAndFilePath networkAddressAndFilePath = new NetworkAddressAndFilePath(fileUrl, FilePathCreator.unknownFilePath());
+            networkAddressAndFileNameById.put(downloadFileId, networkAddressAndFilePath);
+            return this;
+        }
+
+        public Builder addFile(DownloadFileId downloadFileId, String fileUrl, FilePath filePath) {
+            NetworkAddressAndFilePath networkAddressAndFilePath = new NetworkAddressAndFilePath(fileUrl, filePath);
+            networkAddressAndFileNameById.put(downloadFileId, networkAddressAndFilePath);
```
danybony: These 3 methods are growing, and are basically just adding params... what about just using `return this.addFile()`s adding each time the new param?

library/src/main/java/com/novoda/downloadmanager/ExternalFilePersistence.java
```diff
@@ -47,28 +47,41 @@ public FilePersistenceResult create(FileName fileName, FileSize fileSize) {
...
         } catch (FileNotFoundException e) {
             Log.e(e, "File could not be opened");
-            return FilePersistenceResult.newInstance(Status.ERROR_OPENING_FILE, filePath);
+            return FilePersistenceResult.newInstance(Status.ERROR_OPENING_FILE);
+        }
+
+        return FilePersistenceResult.newInstance(Status.SUCCESS, absoluteFilePath);
+    }
+
+    private boolean createParentDirectoriesIfDoesNotExist(File outputFile) {
```
danybony: `createParentDirectoryIfDoesNotExist()`?

danybony: or do not exist if directories

danybony: in general, consistency between singular and plural

library/src/main/java/com/novoda/downloadmanager/FileNameExtractor.java
```diff
@@ -0,0 +1,21 @@
+package com.novoda.downloadmanager;
+
+final class FileNameExtractor {
+
+    private static final String PATH_SEPARATOR = "/";
```
danybony, line 5: Why not using just `File.pathSeparatorChar`?

library/src/test/java/com/novoda/downloadmanager/FileNameExtractorTest.java
```diff
@@ -0,0 +1,27 @@
...
+        assertThat(fileName).isEqualTo(LiteFileName.from("10MB.zip"));
+    }
+
+    @Test
+    public void returnsFileName_whenAssetUrlContainsSubdirectories() {
+        String assetUrl = "foo/bar/10MB.zip";
+
+        FileName fileName = FileNameExtractor.extractFrom(assetUrl);
+
+        assertThat(fileName).isEqualTo(LiteFileName.from("10MB.zip"));
```
danybony, line 24: The content of this test is exactly as the previous one... 
it doesn't show that it contains subdirs

danybony (changes requested):
Did you test that it works with both external and internal persistence? Just asking because of the change in accessing the file when internal

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
MechaRyRy: They might not care about the file id. It's an optional really but I would rather keep it as is until I can find time to create a more fluent builder. For example I'd rather we had something like:

```
Batch.Builder(BATCH_ID_1, "Made in chelsea")
         .addFile(URL).withId(FILE_ID).done()
         .addFile(ADDITIONAL_URL).withStoragePath(PATH).done()
.build()
```

library/src/main/java/com/novoda/downloadmanager/ExternalFilePersistence.java
MechaRyRy: yea this is a bit weird 😆

library/src/main/java/com/novoda/downloadmanager/FileNameExtractor.java
MechaRyRy: I didn't know this existed tbh but now that I do I can say that it won't work 😄  From the docs:

```
     * The system-dependent path-separator character.  This field is
     * initialized to contain the first character of the value of the system
     * property <code>path.separator</code>.  This character is used to
     * separate filenames in a sequence of files given as a <em>path list</em>.
     * On UNIX systems, this character is <code>':'</code>; on Microsoft Windows systems it
     * is <code>';'</code>.
```

It's using the filesystem separator, we want the network address separator.

library/src/test/java/com/novoda/downloadmanager/FileNameExtractorTest.java
MechaRyRy: `returnsFileName_whenAssetUrlContainsSubdirectories`

That's the point of the test. Given an asset url that contains subdirectories `/` it should ONLY return the filename.

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
danybony: Ok, let's keep it for now, it's optional you are right...
I like that builder example you showed 💯  (I think it could work even without the `done()` part).
Not for this PR though

library/src/test/java/com/novoda/downloadmanager/FileNameExtractorTest.java
danybony: Ahhh my bad! I'm so used to the `given_when_then` notation in tests names that I quickly read "should contain subdirs", while it's the opposite!
👍  for the test working
👎  for the test naming

library/src/main/java/com/novoda/downloadmanager/FileNameExtractor.java
danybony: AH right, fair enough 👍

library/src/test/java/com/novoda/downloadmanager/FileNameExtractorTest.java
MechaRyRy: Everyone had the opportunity to comment on the test naming a while back and it was decided to go with this one. Personally I prefer this because it isn't nearly as verbose and states exactly what the output is.

danybony: Oh really? Ok then, not a big deal to me ¯\_(ツ)_/¯

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -124,7 +124,7 @@ protected void onCreate(Bundle savedInstanceState) {
 
     private final View.OnClickListener downloadBatchesOnClick = v -> {
         Batch batch = new Batch.Builder(BATCH_ID_1, "Made in chelsea")
-                .addFile(FILE_ID_1, FIVE_MB_FILE_URL, FilePathCreator.create("/foo/bar/10mb.zip"))
+                .addFile(FILE_ID_1, FIVE_MB_FILE_URL, FilePathCreator.create("/foo/bar/5mb.zip"))
```
danybony: 👍

MechaRyRy:
I've added a new method to the `FilePathCreator` that removes the `basePath` from a given `url` if it exists and then recreates the `absolutePath` from the base and relative path.

joetimmins (approved):

danybony (approved):

library/src/main/java/com/novoda/downloadmanager/DownloadBatchFactory.java
```diff
@@ -17,34 +17,41 @@ static DownloadBatch newInstance(Batch batch,
...
+            NetworkAddressAndFilePath networkAddressAndFilePath = networkAddressAndFilePathByDownloadId.getValue();
+
             InternalFileSize fileSize = InternalFileSizeCreator.unknownFileSize();
-            FilePath filePath = FilePathCreator.unknownFilePath();
-            DownloadFileId downloadFileId = urlByDownloadId.getKey();
+
+            String fileNetworkAddress = networkAddressAndFilePath.networkAddress();
+
+            FilePersistenceCreator filePersistenceCreator = fileOperations.filePersistenceCreator();
+            FilePersistence filePersistence = filePersistenceCreator.create();
```
zegnus, line 33: can we extract this outside the loop?

library/src/main/java/com/novoda/downloadmanager/ExternalFilePersistence.java
```diff
@@ -31,7 +31,12 @@ public void initialiseWith(Context context) {
     }
 
     @Override
-    public FilePersistenceResult create(FileName fileName, FileSize fileSize) {
+    public FilePath basePath() {
+        return FilePathCreator.create(getExternalFileDirWithBiggerAvailableSpace().getAbsolutePath(), "/");
```
zegnus, line 35: can you extract as constant? also is this a file path separator or a network path separator?

library/src/main/java/com/novoda/downloadmanager/FileNameExtractor.java
zegnus: should we rename this constant to  network address separator?

library/src/main/java/com/novoda/downloadmanager/RoomDownloadsPersistence.java
```diff
@@ -95,7 +95,7 @@ public void persistFile(DownloadsFilePersisted filePersisted) {
                     DownloadBatchIdCreator.createFrom(roomFile.batchId),
                     DownloadFileIdCreator.createFrom(roomFile.id),
                     LiteFileName.from(roomFile.name),
-                    FilePathCreator.create(roomFile.path),
+                    FilePathCreator.create("", roomFile.path),
```
zegnus, line 98: can you extract this "" so that is has a name?

library/src/main/java/com/novoda/downloadmanager/ExternalFilePersistence.java
MechaRyRy: It's a file path separator that android seems to deal with `/`

zegnus: we should use the android constant then https://developer.android.com/reference/java/io/File.html#separatorChar

PR #312, opened 2018-01-26
V2 file naming migrations

MechaRyRy:
## Problem
When performing #311 we broke the migration part of v2 😬  it attempted to put all of the downloads into the new directory with the same names, it was stripping the name from the network address instead of relying on the current download location 😱 

## Solution
Now that we are relying on a `String` and not hashes for the `FileName` and `FilePath` we can forget about migrating the files from v1 to v2, we can just update the db to point to those files instead! WOOP WOOP. We are assuming that this is ok because if you can access the files now, then you should be able to access them in v2! 

In the case of partial downloads, these are moved to the new internal scheme. We grab the `basePath` and append the `FileName` to generate the absolute path.

## Potential future work
We could allow clients to specify where they want their old files to go by supplying a relative path or absolute path for them.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
@@ -62,22 +62,6 @@ class DownloadBatch {
             return;
         }
 
-        DownloadFile.Callback fileDownloadCallback = downloadFileStatus -> {
-            fileBytesDownloadedMap.put(downloadFileStatus.downloadFileId(), downloadFileStatus.bytesDownloaded());
-            long currentBytesDownloaded = getBytesDownloadedFrom(fileBytesDownloadedMap);
-            downloadBatchStatus.update(currentBytesDownloaded, totalBatchSizeBytes);
-
-            if (currentBytesDownloaded == totalBatchSizeBytes && totalBatchSizeBytes != ZERO_BYTES) {
-                downloadBatchStatus.markAsDownloaded(downloadsBatchPersistence);
-            }
-
-            if (downloadFileStatus.isMarkedAsError()) {
-                downloadBatchStatus.markAsError(downloadFileStatus.error(), downloadsBatchPersistence);
-            }
-
-            callbackThrottle.update(downloadBatchStatus);
-        };
-
         for (DownloadFile downloadFile : downloadFiles) {
             downloadFile.download(fileDownloadCallback);
             if (batchCannotContinue()) {
@@ -91,6 +75,25 @@ class DownloadBatch {
 
         callbackThrottle.stopUpdates();
     }
+
+    private final DownloadFile.Callback fileDownloadCallback = new DownloadFile.Callback() {
+        @Override
+        public void onUpdate(InternalDownloadFileStatus downloadFileStatus) {
+            fileBytesDownloadedMap.put(downloadFileStatus.downloadFileId(), downloadFileStatus.bytesDownloaded());
+            long currentBytesDownloaded = getBytesDownloadedFrom(fileBytesDownloadedMap);
+            downloadBatchStatus.update(currentBytesDownloaded, totalBatchSizeBytes);
+
+            if (currentBytesDownloaded == totalBatchSizeBytes && totalBatchSizeBytes != ZERO_BYTES) {
+                downloadBatchStatus.markAsDownloaded(downloadsBatchPersistence);
+            }
+
+            if (downloadFileStatus.isMarkedAsError()) {
+                downloadBatchStatus.markAsError(downloadFileStatus.error(), downloadsBatchPersistence);
+            }
+
+            callbackThrottle.update(downloadBatchStatus);
+        }
+    };
 
     private boolean networkError() {
         DownloadBatchStatus.Status status = downloadBatchStatus.status();
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
@@ -67,6 +67,10 @@ class DownloadBatch {
             long currentBytesDownloaded = getBytesDownloadedFrom(fileBytesDownloadedMap);
             downloadBatchStatus.update(currentBytesDownloaded, totalBatchSizeBytes);
 
+            if (currentBytesDownloaded == totalBatchSizeBytes && totalBatchSizeBytes != ZERO_BYTES) {
+                downloadBatchStatus.markAsDownloaded(downloadsBatchPersistence);
+            }
+
             if (downloadFileStatus.isMarkedAsError()) {
                 downloadBatchStatus.markAsError(downloadFileStatus.error(), downloadsBatchPersistence);
             }
@@ -83,10 +87,6 @@ class DownloadBatch {
 
         if (networkError()) {
             DownloadsNetworkRecoveryCreator.getInstance().scheduleRecovery();
-        }
-
-        if (status == DOWNLOADED) {
-            downloadBatchStatus.markAsDownloaded(downloadsBatchPersistence);
         }
 
         callbackThrottle.stopUpdates();
--- a/library/src/main/java/com/novoda/downloadmanager/LiteDownloadBatchStatus.java
+++ b/library/src/main/java/com/novoda/downloadmanager/LiteDownloadBatchStatus.java
@@ -41,10 +41,6 @@ class LiteDownloadBatchStatus implements InternalDownloadBatchStatus {
         this.bytesDownloaded = currentBytesDownloaded;
         this.totalBatchSizeBytes = totalBatchSizeBytes;
         this.percentageDownloaded = getPercentageFrom(bytesDownloaded, totalBatchSizeBytes);
-
-        if (this.bytesDownloaded == this.totalBatchSizeBytes && this.totalBatchSizeBytes != ZERO_BYTES) {
-            this.status = Status.DOWNLOADED;
-        }
     }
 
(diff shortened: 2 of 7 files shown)


library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -85,10 +89,6 @@ void download() {
             DownloadsNetworkRecoveryCreator.getInstance().scheduleRecovery();
         }
 
-        if (status == DOWNLOADED) {
```
MechaRyRy, line 88: This was always false because of the guard at the beginning of this class 😬  so all batches were marked as `DOWNLOADING` even when they were `DOWNLOADED` 😬

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -67,6 +67,10 @@ void download() {
             long currentBytesDownloaded = getBytesDownloadedFrom(fileBytesDownloadedMap);
             downloadBatchStatus.update(currentBytesDownloaded, totalBatchSizeBytes);
 
+            if (currentBytesDownloaded == totalBatchSizeBytes && totalBatchSizeBytes != ZERO_BYTES) {
```
MechaRyRy: A version of this was originally in the `DownloadBatchStatus` but it never marked the `status` as `DOWNLOADED` at the persistence layer, just in memory 😬

library/src/main/java/com/novoda/downloadmanager/LiteDownloadBatchStatus.java
```diff
@@ -41,10 +41,6 @@ public void update(long currentBytesDownloaded, long totalBatchSizeBytes) {
         this.bytesDownloaded = currentBytesDownloaded;
         this.totalBatchSizeBytes = totalBatchSizeBytes;
         this.percentageDownloaded = getPercentageFrom(bytesDownloaded, totalBatchSizeBytes);
-
-        if (this.bytesDownloaded == this.totalBatchSizeBytes && this.totalBatchSizeBytes != ZERO_BYTES) {
-            this.status = Status.DOWNLOADED;
```
MechaRyRy, line 46: Only updating in memory, never hits the persistence layer 😬

library/src/main/java/com/novoda/downloadmanager/Migration.java
```diff
@@ -74,14 +74,15 @@ public String toString() {
     }
 
     static class FileMetadata {
+
         private final String originalFileLocation;
         private final FileSize fileSize;
-        private final String uri;
+        private final String originalNetworkAddress;
```
MechaRyRy, line 80: Renamed these because I had to keep jumping backwards and forwards to find out what each of them was doing 😄

library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
```diff
@@ -15,8 +15,8 @@
     private static final int MODIFIED_TIMESTAMP_COLUMN = 2;
 
     private static final String DOWNLOADS_QUERY = "SELECT uri, _data, total_bytes FROM Downloads WHERE batch_id = ?";
-    private static final int URI_COLUMN = 0;
-    private static final int FILE_NAME_COLUMN = 1;
+    private static final int NETWORK_ADDRESS_COLUMN = 0;
+    private static final int FILE_LOCATION_COLUMN = 1;
```
MechaRyRy, line 20: This naming was just plan wrong 😂

library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
```diff
@@ -35,26 +35,31 @@
...
+            List<Migration.FileMetadata> fileMetadataList = extractFileMetadataFrom(batchId, newBatchBuilder);
 
             Batch batch = newBatchBuilder.build();
             migrations.add(new Migration(batch, fileMetadataList, downloadedDateTimeInMillis));
         }
         batchesCursor.close();
         return migrations;
     }
+
+    private List<Migration.FileMetadata> extractFileMetadataFrom(String batchId, Batch.Builder newBatchBuilder) {
```
MechaRyRy: Extracted this in an attempt to make everything a little more readable.

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
```diff
@@ -103,10 +103,14 @@ private void migrateV1DataToV2Database(DownloadsPersistence downloadsPersistence
         downloadsPersistence.persistBatch(persistedBatch);
 
         for (Migration.FileMetadata fileMetadata : migration.getFileMetadata()) {
-            String url = fileMetadata.uri();
+            String url = fileMetadata.originalNetworkAddress();
 
+            FilePath filePath = new LiteFilePath(fileMetadata.originalFileLocation());
+            if (filePath.path() == null || filePath.path().isEmpty()) {
```
MechaRyRy, line 108: If there is no existing path then we create it from the new persistence base path and the file name of whatever we are trying to download.

MechaRyRy: This still suffers from the issue of duplicate files which will be sorted when we add a directory per batch.

jszmltr: It might be worth adding this as an `Issue` so we won't forget

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
```diff
@@ -144,8 +147,7 @@ private boolean hasValidFileLocation(Migration.FileMetadata metadata) {
     private void migrateCompleteDownloads(InternalMigrationStatus migrationStatus,
                                           SqlDatabaseWrapper database,
                                           MigrationExtractor migrationExtractor,
-                                          DownloadsPersistence downloadsPersistence,
-                                          InternalFilePersistence internalFilePersistence) {
+                                          DownloadsPersistence downloadsPersistence, String basePath) {
```
jszmltr: New line here :)

library/src/main/java/com/novoda/downloadmanager/LiteDownloadBatchStatus.java
zegnus: Are you 100% sure of this? the update is called from `DownloadsBatchPersistence.loadAsync()`

library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
```diff
@@ -35,26 +35,31 @@
...
+        Cursor downloadsCursor = database.rawQuery(DOWNLOADS_QUERY, batchId);
```
zegnus: can this return null?

library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
```diff
@@ -35,26 +35,31 @@
...
+        List<Migration.FileMetadata> fileMetadataList = new ArrayList<>();
+
+        while (downloadsCursor.moveToNext()) {
```
zegnus: we should protect all cursor operations with a try{}{cursor.close} ;) also review the rest code in this class regarding cursors

MechaRyRy:
Created an issue to track the duplicate file problem @jackSzm https://github.com/novoda/download-manager/issues/313

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
```diff
@@ -183,39 +184,4 @@ private void migrateCompleteDownloads(InternalMigrationStatus migrationStatus,
         onUpdate(migrationStatus);
     }
 
-    private void migrateV1FilesToV2Location(InternalFilePersistence internalFilePersistence, Migration migration) {
```
zegnus, line 186: nice, this might have improved the performance/time done for migrations, do you have any results on how long does it takes now?

library/src/main/java/com/novoda/downloadmanager/LiteDownloadBatchStatus.java
MechaRyRy: 100% sure, there is a check for `DOWNLOADED` above that instantly returns so if it was ever modified `in-memory` then it would never hit the persistence. I tested this quite a lot manually and with the debugger when I realised there was a problem.

library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
MechaRyRy: I mean the docs do not say it can return null 😬  I'll add a check to be safe.

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
MechaRyRy: It's pretty much instant now 😂

MechaRyRy:
@zegnus now accessing Cursors a little more safely! 73bb768

jszmltr (approved):

library/src/main/java/com/novoda/downloadmanager/LiteDownloadBatchStatus.java
zegnus: Yeah I know but the load async is different, because there the file status is set to downloaded immediately, doesn't need to be queued to be downloaded, processed and then is immediately changed to downloaded. But if you have tested that the loadAsync works fine for downloaded files, all good

zegnus (approved):

PR #314, opened 2018-01-29
Fluent BatchBuilder

MechaRyRy:
## Problem
We started introducing more params to the `Batch` builder but a lot of these parameters are optional. Most of the methods on the `Batch.Builder` actually create a `File` object and add to the `Batch`.

## Solution
Introduce a fluent batch builder, where calling `addFile` actually returns a `File.Builder` that can be used to create a `File` that will be added to the `Batch`. I've introduced `Optional` to the project now rather than handle `nulls` and `empty` in this `File` builder. 

## Screen Capture

Before | screen shot 2018-01-29 at 11 04 39
--- | --- | 
**After** | screen shot 2018-01-29 at 11 04 24


demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -123,14 +122,14 @@ protected void onCreate(Bundle savedInstanceState) {
 
     private final View.OnClickListener downloadBatchesOnClick = v -> {
         Batch batch = new Batch.Builder(BATCH_ID_1, "Made in chelsea")
-                .addFile(FILE_ID_1, FIVE_MB_FILE_URL, "foo/bar/5mb.zip")
-                .addFile(TEN_MB_FILE_URL)
+                .addFile(FIVE_MB_FILE_URL).withDownloadFileId(FILE_ID_1).withRelativePath("foo/bar/5mb.zip").apply()
+                .addFile(TEN_MB_FILE_URL).apply()
```
MechaRyRy: The `apply` returns the parent builder (`Batch.Builder`) so the user can continue to add additional files and build the final batch.

library/src/main/java/com/novoda/downloadmanager/Batch.java
```diff
@@ -61,40 +61,38 @@ public String toString() {
...
             this.downloadBatchId = downloadBatchId;
             this.title = title;
+            files = new ArrayList<>();
         }
 
-        public Builder addFile(String fileUrl) {
-            String rawId = downloadBatchId.rawId() + fileUrl;
-            DownloadFileId downloadFileId = DownloadFileIdCreator.createFrom(rawId);
-            return addFile(downloadFileId, fileUrl);
+        Builder withFile(File file) {
```
MechaRyRy: This is probably my biggest beef with this pattern, wondering if it would be better to introduce two different interfaces to better define these `Builder` methods. 

`withFile` is called from the `File.Builder` when the `apply` is called. The `apply` builds the final `File` object and then applies it to the parent builder with this `withFile` method.

library/src/main/java/com/novoda/downloadmanager/DownloadBatchFactory.java
```diff
@@ -17,26 +16,24 @@ static DownloadBatch newInstance(Batch batch,
...
-        DownloadBatchId downloadBatchId = batch.getDownloadBatchId();
+        DownloadBatchId downloadBatchId = batch.downloadBatchId();
         long downloadedDateTimeInMillis = System.currentTimeMillis();
 
-        for (Map.Entry<DownloadFileId, NetworkAddressAndFilePath> networkAddressAndFilePathByDownloadId : networkAddressAndFilePathById.entrySet()) {
-            NetworkAddressAndFilePath networkAddressAndFilePath = networkAddressAndFilePathByDownloadId.getValue();
+        List<DownloadFile> downloadFiles = new ArrayList<>(batch.files().size());
 
-            InternalFileSize fileSize = InternalFileSizeCreator.unknownFileSize();
+        for (File file : batch.files()) {
```
MechaRyRy: Back to a slightly simpler approach here rather than the map 😄

library/src/test/java/com/novoda/downloadmanager/BatchBuilderTest.java
```diff
@@ -0,0 +1,41 @@
+package com.novoda.downloadmanager;
+
+import java.util.Collections;
+
+import org.junit.Test;
+
+import static com.google.common.truth.Truth.assertThat;
+
+public class BatchBuilderTest {
```
MechaRyRy, line 9: Just wanted to make sure that the `Builder` operated as I expected as I was writing it 😄

library/src/main/java/com/novoda/downloadmanager/Batch.java
```diff
@@ -1,30 +1,34 @@
...
+        this.files = files;
     }
 
-    DownloadBatchId getDownloadBatchId() {
+    public DownloadBatchId downloadBatchId() {
         return downloadBatchId;
     }
 
-    String getTitle() {
+    public String title() {
```
tasomaniac, line 26: Just a question: Is this the getter naming we have for the `v2` of the library? Is there any discussion link around this? I personally prefer regular getters because it is more usable in kotlin and in java, they're ok.

library/src/main/java/com/novoda/downloadmanager/File.java
```diff
@@ -0,0 +1,137 @@
+package com.novoda.downloadmanager;
+
+public class File {
+
+    private final String networkAddress;
+    private final Optional<DownloadFileId> downloadFileId;
+    private final Optional<FileName> fileName;
+    private final Optional<String> relativePath;
+
+    File(String networkAddress, Optional<DownloadFileId> downloadFileId, Optional<FileName> fileName, Optional<String> relativePath) {
```
tasomaniac: Especially since this is related to files, downloads etc. I found this class name confusing. This is supposed to be used by users, isn't it? Or is it internal? 

Considering it is public API, it could be annoying to use this if the user also have `java.io.File` imported in their class.

library/src/main/java/com/novoda/downloadmanager/Batch.java
MechaRyRy: There hasn't been a conversation around this, I think most on the project prefer without the `get`

library/src/main/java/com/novoda/downloadmanager/File.java
MechaRyRy: I agree. What naming would you suggest?

library/src/main/java/com/novoda/downloadmanager/Optional.java
```diff
@@ -0,0 +1,88 @@
+package com.novoda.downloadmanager;
+
+public final class Optional<T> {
```
tasomaniac: How can we avoid this `Optional` class to be leaked to the client libraries?

With this addition, client's will be able to just use this class since it is public and exposed. 

First of all, If they have java8, then it will be a problem since it is redundant. 
Also they may already have `Optional` class in their project. So this will be a duplication in auto completions in IDE and can be annoying. 

**Possible Solutions:** 
- Create another gradle module called `internal` and move this there and depend on `internal` by using `implementation`. Gradle hides it for you. 
- Or if that's complicated, simply move this to `internal` package. (OkHttp does this). This way it will be little bit clear for library users that they are not supposed to use this.

tasomaniac: Also another problem is that, in the future where we increase our minSdk to suppor java 8 natively, it will be hard to remove this class because clients may use this and since it is not marked as internal, just removing this will be really annoying for them.

library/src/main/java/com/novoda/downloadmanager/Batch.java
tasomaniac: You mean our internal projects or `download-manager`?

library/src/main/java/com/novoda/downloadmanager/File.java
tasomaniac: Hmm, maybe `BatchFile`? It already has references to batch.
I was thinking `FileDescriptor` but that's also a `java.io` class. 😄

library/src/main/java/com/novoda/downloadmanager/Batch.java
MechaRyRy: `download-manager`

library/src/main/java/com/novoda/downloadmanager/Optional.java
MechaRyRy: 9351e07

library/src/main/java/com/novoda/downloadmanager/File.java
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        @Override
+        public Builder withRelativePath(String relativePath) {
+            this.relativePath = Optional.of(relativePath);
+            return this;
+        }
+
+        @Override
+        public Batch.Builder apply() {
```
joetimmins: What's the thinking behind this method?

library/src/main/java/com/novoda/downloadmanager/Batch.java
tasomaniac: Consistency ftw 👍

library/src/main/java/com/novoda/downloadmanager/File.java
MechaRyRy: So you can return the parent builder so you have a fluent builder, one where you can continue to add files.

library/src/main/java/com/novoda/downloadmanager/File.java
MechaRyRy: fa0527e

tasomaniac (approved):

zegnus (approved):

PR #316, opened 2018-01-30
V2 error status

MechaRyRy:
## Problem
Given a network error occurs when a reschedule job occurs the `download` file status fails to update. 

A batch is marked as `QUEUED` when retrying but the `File` is still marked as error. It will download the file because there aren't checks to block this in the `DownloadFile` but no updates are forwarded to the `UI` because it is blocked. 

## Solution
When the `File.Status == ERROR` mark the `File` as `DOWNLOADING` this will force the `file` to continue downloading and pass updates back to the UI and notification system.

`isMarkedAsError` was checking for an `Error` object which may not exist when coming from persistence layer, so the checks fail for `isMarkedAsError`. Instead we now have `Optional<Error>` and where this is exposed to a client we return either `Error.get` or `null`.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
@@ -56,7 +56,7 @@ class DownloadBatch {
         totalBatchSizeBytes = getTotalSize(downloadFiles);
 
         if (totalBatchSizeBytes <= ZERO_BYTES) {
-            DownloadError downloadError = new DownloadError(DownloadError.Error.NETWORK_ERROR_CANNOT_DOWNLOAD_FILE);
+            Optional<DownloadError> downloadError = Optional.of(new DownloadError(DownloadError.Error.NETWORK_ERROR_CANNOT_DOWNLOAD_FILE));
             downloadBatchStatus.markAsError(downloadError, downloadsBatchPersistence);
             notifyCallback(downloadBatchStatus);
             return;
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
@@ -46,7 +46,7 @@ class DownloadBatch {
 
     void download() {
         DownloadBatchStatus.Status status = downloadBatchStatus.status();
-        if (status == PAUSED || status == DELETION || status == DOWNLOADED) {
+        if (status == PAUSED || status == DELETION) {
             return;
         }
 
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadFile.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadFile.java
@@ -47,7 +47,7 @@ class DownloadFile {
     }
 
     void download(final Callback callback) {
-        moveStatusToDownloadingIfQueuedOrError();
+        downloadFileStatus.markAsDownloading();
 
         callback.onUpdate(downloadFileStatus);
 
@@ -144,12 +144,6 @@ class DownloadFile {
     private void updateAndFeedbackWithStatus(Error error, Callback callback) {
         downloadFileStatus.markAsError(error);
         callback.onUpdate(downloadFileStatus);
-    }
-
-    private void moveStatusToDownloadingIfQueuedOrError() {
-        if (downloadFileStatus.isMarkedAsQueued() || downloadFileStatus.isMarkedAsError()) {
-            downloadFileStatus.markAsDownloading();
-        }
     }
 
     void pause() {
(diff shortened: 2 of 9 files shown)


library/src/main/java/com/novoda/downloadmanager/DownloadFile.java
```diff
@@ -47,9 +47,7 @@
     }
 
     void download(final Callback callback) {
-        callback.onUpdate(downloadFileStatus);
-
-        moveStatusToDownloadingIfQueued();
+        moveStatusToDownloadingIfQueuedOrError();
```
MechaRyRy: When the `Batch.status` is Error it will be modified to `QUEUED` again, this is not the same for `File` which was just left as `ERROR`. Now we make sure that the `file` is updated. We could change this so that it is based off of the `DownloadBatch` instead, rather than just remapping.

library/src/main/java/com/novoda/downloadmanager/DownloadFile.java
```diff
@@ -68,6 +66,8 @@ void download(final Callback callback) {
         filePath = result.filePath();
         fileSize.setCurrentSize(filePersistence.getCurrentSize());
 
+        callback.onUpdate(downloadFileStatus);
```
MechaRyRy: Performing an initial update that might be error, see removal from line 50 above. Instead we now wait until we have updated all of the status and then emit before performing the potentially long running `Sync` below.

library/src/main/java/com/novoda/downloadmanager/DownloadFile.java
```diff
@@ -146,8 +146,8 @@ private void updateAndFeedbackWithStatus(Error error, Callback callback) {
         callback.onUpdate(downloadFileStatus);
     }
 
-    private void moveStatusToDownloadingIfQueued() {
-        if (downloadFileStatus.isMarkedAsQueued()) {
+    private void moveStatusToDownloadingIfQueuedOrError() {
+        if (downloadFileStatus.isMarkedAsQueued() || downloadFileStatus.isMarkedAsError()) {
```
MechaRyRy: This could be driven by the `Batch.status` instead.

library/src/main/java/com/novoda/downloadmanager/LiteDownloadBatchStatus.java
```diff
@@ -116,13 +113,8 @@ private void updateStatus(Status status, DownloadsBatchStatusPersistence persist
...
 
-    @Nullable
     @Override
     public DownloadError.Error getDownloadErrorType() {
-        if (downloadError != null) {
-            return downloadError.error();
-        } else {
-            return null;
-        }
+        return downloadError.or(new DownloadError(DownloadError.Error.UNKNOWN)).error();
```
MechaRyRy: I think this needs to still be `@Nullable` but the errors are not stored by the persistence layer so when we come back to the app and an error is displayed it will appear as `UNKNOWN` or `null` 😢

danybony (approved):

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -46,7 +46,7 @@ void setCallback(DownloadBatchStatusCallback callback) {
 
     void download() {
         DownloadBatchStatus.Status status = downloadBatchStatus.status();
-        if (status == PAUSED || status == DELETION || status == DOWNLOADED) {
```
MechaRyRy, line 49: This was added to get around a double call of notifications, when a file is marked as downloading and then again when the download is immediately marked as downloaded. Instead we should handle this as a new flag in the db, we still have outstanding issues with notifications anyway that this didn't fix that the db solution would.

library/src/main/java/com/novoda/downloadmanager/DownloadFile.java
```diff
@@ -47,9 +47,9 @@
     }
 
     void download(final Callback callback) {
-        callback.onUpdate(downloadFileStatus);
+        downloadFileStatus.markAsDownloading();
```
MechaRyRy, line 50: @zegnus I have removed the check that was marking files as `DOWNLOADING` when it was either `QUEUED` or `ERROR`. As far as I am aware we should always be able to mark them as `DOWNLOADING` initially when calling `download`, the status will be immediately updated after this point with the file size checks etc.

MechaRyRy: It's important to note this will currently retrigger a download if the file size is incorrect, which is fixed by #317

MechaRyRy:
#317 needs to be merged first otherwise this branch will continuously re-download because the `getCurrentSize` is returning as 0 so the file size checks all fail 😬

danybony:
💯

PR #317, opened 2018-01-30
File getCurrentSize

MechaRyRy:
## Problem
`fileOutputStream.getChannel().size();` is outputting 0 for the file size for internal persistence. I'm unsure as to why this is the case, it seems that `FileChannel` is readonly and shares the position information from the stream. It doesn't seem to guarantee that the `size` is referring to the size of the file as any changes on the stream can influence this `FileChannel` object. 

## Solution
From what I have read it seems generally better to use `File.length` instead or use `Files.size` but that's only available from api 26 and onwards.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/ExternalFilePersistence.java
+++ b/library/src/main/java/com/novoda/downloadmanager/ExternalFilePersistence.java
@@ -142,13 +142,8 @@ class ExternalFilePersistence implements FilePersistence {
 
     @Override
     public long getCurrentSize(FilePath filePath) {
-        try {
-            File file = new File(filePath.path());
-            return file.length();
-        } catch (NullPointerException e) {
-            Log.e(e, "Error requesting file size for " + filePath.path());
-            return 0;
-        }
+        File file = new File(filePath.path());
+        return file.length();
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


library/src/main/java/com/novoda/downloadmanager/ExternalFilePersistence.java
```diff
@@ -140,30 +140,10 @@ public void delete(FilePath absoluteFilePath) {
...
     @Override
     public long getCurrentSize(FilePath filePath) {
-        try {
-            FileOutputStream file = new FileOutputStream(filePath.path(), APPEND);
-            return file.getChannel().size();
-        } catch (IOException e) {
-            Log.e(e, "Error requesting file size for " + filePath);
-            return 0;
-        }
+        File file = new File(filePath.path());
```
MechaRyRy, line 145: Changed this to be consistent with the `InternalFilePersistence`, this works and seems to be the preferred method.

danybony (approved):
Good catch this! Did you check that the method is working for both internal and external? (I'm pretty sure it does, but just double-checking)

MechaRyRy:
Yes @danybony I did check both internal and external persistence.

PR #318, opened 2018-01-31
Check for allowed networks

danybony:
### Problem
Clients need to be able to specify the kind of network to be used for downloads (wifi-only or all networks)

### Solution
Wrap the usage of [Merlin](https://github.com/novoda/merlin) to check for the network status relative to the allowed network selected by the client. This check is then used in `DownloadBatch` when downloading: in case of not allowed network connection the download batch is marked as network error and a recovery job is created, using the network type specifyed by the client.
This is testable using the demo app.

### Screenshots

No Wifi, all network allowed | Wifi enabled later | No wifi, all networks allowed later 
--- | --- | ---
no_wifi_no_cgecj mp4 | no_wifi_then_wifi_enabled mp4 | no_wifi_toggle_changed mp4

### DIFF
--- a/demo-simple/src/main/java/com/novoda/downloadmanager/demo/DemoApplication.java
+++ b/demo-simple/src/main/java/com/novoda/downloadmanager/demo/DemoApplication.java
@@ -5,6 +5,7 @@ import android.os.Handler;
 import android.os.Looper;
 
 import com.facebook.stetho.Stetho;
+import com.novoda.downloadmanager.ConnectionType;
 import com.novoda.downloadmanager.DownloadManagerBuilder;
 import com.novoda.downloadmanager.LiteDownloadManagerCommands;
 
@@ -24,6 +25,7 @@ public class DemoApplication extends Application {
 
         liteDownloadManagerCommands = DownloadManagerBuilder
                 .newInstance(this, handler, R.mipmap.ic_launcher_round)
+                .withAllowedConnectionType(ConnectionType.UNMETERED)
                 .build();
     }
 
--- a/demo-simple/src/main/java/com/novoda/downloadmanager/demo/DemoApplication.java
+++ b/demo-simple/src/main/java/com/novoda/downloadmanager/demo/DemoApplication.java
@@ -5,7 +5,6 @@ import android.os.Handler;
 import android.os.Looper;
 
 import com.facebook.stetho.Stetho;
-import com.novoda.downloadmanager.ConnectionType;
 import com.novoda.downloadmanager.DownloadManagerBuilder;
 import com.novoda.downloadmanager.LiteDownloadManagerCommands;
 
@@ -25,7 +24,6 @@ public class DemoApplication extends Application {
 
         liteDownloadManagerCommands = DownloadManagerBuilder
                 .newInstance(this, handler, R.mipmap.ic_launcher_round)
-                .withAllowedConnectionType(ConnectionType.UNMETERED)
                 .build();
     }
 
--- a/demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
+++ b/demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
@@ -44,7 +44,6 @@ public class MainActivity extends AppCompatActivity {
     private static final String FIVE_MB_FILE_URL = "http://ipv4.download.thinkbroadband.com/5MB.zip";
     private static final String TEN_MB_FILE_URL = "http://ipv4.download.thinkbroadband.com/10MB.zip";
     private static final String TWENTY_MB_FILE_URL = "http://ipv4.download.thinkbroadband.com/20MB.zip";
-    private static final String FIFTHY_MB_FILE_URL = "http://ipv4.download.thinkbroadband.com/50MB.zip";
 
     private TextView databaseCloningUpdates;
     private TextView databaseMigrationUpdates;
@@ -79,9 +78,6 @@ public class MainActivity extends AppCompatActivity {
         textViewBatch1 = findViewById(R.id.batch_1);
         textViewBatch2 = findViewById(R.id.batch_2);
 
-        CheckBox checkWifiOnly = findViewById(R.id.check_wifi_only);
-        checkWifiOnly.setOnCheckedChangeListener(wifiOnlyOnCheckedChange);
-
         versionOneDatabaseCloner = DatabaseClonerFactory.databaseCloner(this, cloneCallback);
 
         downloadFileSizeSpinner = findViewById(R.id.database_download_file_size);
@@ -95,6 +91,9 @@ public class MainActivity extends AppCompatActivity {
         databaseMigrationUpdates = findViewById(R.id.database_migration_updates);
         View buttonMigrate = findViewById(R.id.button_migrate);
         buttonMigrate.setOnClickListener(startMigrationOnClick);
+
+        CheckBox checkWifiOnly = findViewById(R.id.check_wifi_only);
+        checkWifiOnly.setOnCheckedChangeListener(wifiOnlyOnCheckedChange);
 
         View buttonDownload = findViewById(R.id.button_start_downloading);
         buttonDownload.setOnClickListener(downloadBatchesOnClick);
@@ -118,14 +117,6 @@ public class MainActivity extends AppCompatActivity {
 
     private final VersionOneDatabaseCloner.CloneCallback cloneCallback = updateMessage -> databaseCloningUpdates.setText(updateMessage);
 
-    private final CompoundButton.OnCheckedChangeListener wifiOnlyOnCheckedChange = (buttonView, isChecked) -> {
-        LiteDownloadManagerCommands downloadManagerCommands = ((DemoApplication) getApplication()).getLiteDownloadManagerCommands();
-        if (isChecked) {
-            downloadManagerCommands.updateAllowedConnectionType(ConnectionType.UNMETERED);
-        } else {
-            downloadManagerCommands.updateAllowedConnectionType(ConnectionType.ALL);
-        }
-    };
 
     private final View.OnClickListener createDatabaseOnClick = v -> {
         String selectedFileSize = (String) downloadFileSizeSpinner.getSelectedItem();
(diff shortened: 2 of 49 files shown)


demo-simple/src/main/java/com/novoda/downloadmanager/demo/DemoApplication.java
```diff
@@ -24,6 +25,7 @@ private void createLiteDownloadManager() {
 
         liteDownloadManagerCommands = DownloadManagerBuilder
                 .newInstance(this, handler, R.mipmap.ic_launcher_round)
+                .withAllowedConnectionType(ConnectionType.UNMETERED)
```
MechaRyRy: Can we remove this, I had a problem when I first worked on the project where we had a bunch of these defined in `Application` and they were changing the `download-manager` without me knowing 😄  would prefer to keep just the vanilla in the sample app.

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -40,7 +43,8 @@
     private static final DownloadFileId FILE_ID_1 = DownloadFileIdCreator.createFrom("file_id_1");
     private static final String FIVE_MB_FILE_URL = "http://ipv4.download.thinkbroadband.com/5MB.zip";
     private static final String TEN_MB_FILE_URL = "http://ipv4.download.thinkbroadband.com/10MB.zip";
-    private static final String TWENTY_FILE_URL = "http://ipv4.download.thinkbroadband.com/20MB.zip";
+    private static final String TWENTY_MB_FILE_URL = "http://ipv4.download.thinkbroadband.com/20MB.zip";
+    private static final String FIFTHY_MB_FILE_URL = "http://ipv4.download.thinkbroadband.com/50MB.zip";
```
MechaRyRy: typo `FIFTY` also not used?

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -75,6 +79,9 @@ protected void onCreate(Bundle savedInstanceState) {
         textViewBatch1 = findViewById(R.id.batch_1);
         textViewBatch2 = findViewById(R.id.batch_2);
 
+        CheckBox checkWifiOnly = findViewById(R.id.check_wifi_only);
+        checkWifiOnly.setOnCheckedChangeListener(wifiOnlyOnCheckedChange);
```
MechaRyRy: Would move this to be in order with how it is displayed in the UI.

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -111,6 +118,15 @@ protected void onCreate(Bundle savedInstanceState) {
 
     private final VersionOneDatabaseCloner.CloneCallback cloneCallback = updateMessage -> databaseCloningUpdates.setText(updateMessage);
 
+    private final CompoundButton.OnCheckedChangeListener wifiOnlyOnCheckedChange = (buttonView, isChecked) -> {
```
MechaRyRy: Order based on usage when the above is modified too.

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -50,6 +53,13 @@ void download() {
             return;
         }
 
+        if(!downloadConnectionAllowedChecker.isAllowedToDownload()) {
```
MechaRyRy: How does this work if you `resume` a download? It's using the same instance right?

danybony: Mmm can you describe the steps / expected result?

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
danybony: https://github.com/novoda/download-manager/pull/318/commits/ba90d9f436d912b9b56b09b81c2b52cecaf09052

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
danybony: https://github.com/novoda/download-manager/pull/318/commits/ba90d9f436d912b9b56b09b81c2b52cecaf09052

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
danybony: https://github.com/novoda/download-manager/pull/318/commits/ba90d9f436d912b9b56b09b81c2b52cecaf09052

demo-simple/src/main/java/com/novoda/downloadmanager/demo/DemoApplication.java
danybony: https://github.com/novoda/download-manager/pull/318/commits/b1559fd34d40862bf8f63d42f9ab6230c22adcf0

library/src/main/java/com/novoda/downloadmanager/DownloadManager.java
```diff
@@ -181,4 +184,13 @@ public void getDownloadStatusWithMatching(DownloadFileId downloadFileId, Downloa
                 }));
     }
 
+    @Override
+    public void updateAllowedConnectionType(ConnectionType allowedConnectionType) {
+        downloadConnectionAllowedChecker.updateAllowedConnectionType(allowedConnectionType);
+        DownloadsNetworkRecoveryCreator.getInstance().updateAllowedConnection(allowedConnectionType);
+        submitAllStoredDownloads(() -> {
+            // no-op
```
MechaRyRy: Could add a `verbose` log?

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -120,6 +127,15 @@ protected void onCreate(Bundle savedInstanceState) {
 
     private final View.OnClickListener startMigrationOnClick = v -> downloadMigrator.startMigration("downloads.db", migrationCallback);
 
+    private final CompoundButton.OnCheckedChangeListener wifiOnlyOnCheckedChange = (buttonView, isChecked) -> {
+        LiteDownloadManagerCommands downloadManagerCommands = ((DemoApplication) getApplication()).getLiteDownloadManagerCommands();
+        if (isChecked) {
+            downloadManagerCommands.updateAllowedConnectionType(ConnectionType.UNMETERED);
```
MechaRyRy, line 133: Should we be handling the case where the client passes us `Null`? it will crash internally if they do right.

MechaRyRy: I think most other methods on the DLM will have a guard against nulls either when retrieving from a map or checking against other objects.

danybony: Good point... it's however right to crash if that happens, to me. What about if we throw an `IllegalArgumentException` in that case?

MechaRyRy: In this case it is probably valid to throw a `NullPointer` now that you mention it, the others work on ids so if you get an invalid id you just swallow the problem. Here if you pass an incorrect connection type it is strange, especially as it is normally an enum 😄

danybony: https://github.com/novoda/download-manager/pull/318/commits/2d03303065f9c816e8e46e455c4bdb48ed14f218

library/src/main/java/com/novoda/downloadmanager/DownloadManager.java
danybony: https://github.com/novoda/download-manager/pull/318/commits/39a04b46263d0b43c110c31890f52b882ab814e7

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
MechaRyRy: - Start downloading
- Pause a download
- Trigger the wifi check thing (submitAllStoredDownloads)

Does it resume? 

Then try:

- Start downloading
- Pause a download
- Trigger the wifi check thing (submitAllStoredDownloads)
- Resume the download where the download now violates the network type check

Does it resume anyway?

danybony: Ok, I just checked the two scenarios:
in the first one it doesn't resume only toggling the allowed connection, it stays paused.
In the second one when resuming after changing the condition to a violating one, it immediately fail for network error (restored later if the condition changes)

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -31,12 +32,14 @@
                   List<DownloadFile> downloadFiles,
                   Map<DownloadFileId, Long> fileBytesDownloadedMap,
                   DownloadsBatchPersistence downloadsBatchPersistence,
-                  CallbackThrottle callbackThrottle) {
+                  CallbackThrottle callbackThrottle,
+                  DownloadConnectionAllowedChecker downloadConnectionAllowedChecker) {
```
tasomaniac: What about just `DownloadConnectionChecker` or even `ConnectionChecker`. `allowed` word is already available in `isAllowedToDownload` method which I think makes it clear.

library/src/main/java/com/novoda/downloadmanager/DownloadManager.java
```diff
@@ -181,4 +186,14 @@ public void getDownloadStatusWithMatching(DownloadFileId downloadFileId, Downloa
                 }));
     }
 
+    @Override
+    public void updateAllowedConnectionType(ConnectionType allowedConnectionType) {
+        if (allowedConnectionType == null) {
+            throw new IllegalArgumentException("Allowed connection type cannot be null");
+        }
+        downloadConnectionAllowedChecker.updateAllowedConnectionType(allowedConnectionType);
+        DownloadsNetworkRecoveryCreator.getInstance().updateAllowedConnection(allowedConnectionType);
```
tasomaniac: Maybe rename this method to be consistent with `updateAllowedConnectionType`

library/src/test/java/com/novoda/downloadmanager/DownloadManagerTest.java
```diff
@@ -52,6 +52,7 @@
     private final FileOperations fileOperations = mock(FileOperations.class);
     private final DownloadsBatchPersistence downloadsBatchPersistence = mock(DownloadsBatchPersistence.class);
     private final LiteDownloadManagerDownloader downloadManagerDownloader = mock(LiteDownloadManagerDownloader.class);
+    private DownloadConnectionAllowedChecker downloadConnectionAllowedChecker = mock(DownloadConnectionAllowedChecker.class);
```
tasomaniac: `final` is missing?

tasomaniac (commented):
Just small comments. 

I think we may also add more unit tests around connection then? Covering newly added implementation.

tasomaniac (approved):

PR #320, opened 2018-02-01
Queued download when wrong connection type

danybony:
Queue instead of mark as error a download when the connection is not the allowed one.
The behaviour is exactly the same as before


MechaRyRy:
I'm not 100% sure what affect this has on the internals of `dlm` but I guess it has been checked 😄

PR #321, opened 2018-02-01
Duplicate notifications on restart

MechaRyRy:
## Problem
1. When restarting the app after having shown a stacked notification, ERROR, DOWNLOADED, then the app will reshow these notifications even when originally they were swiped away by the user.

2. When pausing a download in the `download-manager` sample the notification switches to the other non-paused download. So the user has no idea what happened to the other paused download, we aren't showing a notification for it 😬 

## Solution
1. Add a persisted flag to the `DownloadBatchStatus` for `notificationSeen` that will be triggered when a notification for `DOWNLOADED` is shown. When the app restarts the first event from the `DownloadBatchStatus` will show the `notificationSeen` as true and swallow the notification. For the `ERROR` status the `download-manager` will force a re-queue so a new batch of notifications are required.

2. Add a `stackNotificationNotDismissible` which is used for the `PAUSED` state to show a persistent notification. This can be changed to an ordinary stacked notification if desired. 

#### Downloaded 
Before | After
--- | ---
before_downloaded | after_downloaded

#### Paused 
Before | After
--- | ---
before_paused | after_paused

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/NotificationDispatcher.java
+++ b/library/src/main/java/com/novoda/downloadmanager/NotificationDispatcher.java
@@ -7,6 +7,7 @@ import com.novoda.notils.logger.simple.Log;
 import static com.novoda.downloadmanager.DownloadBatchStatus.Status.DELETION;
 import static com.novoda.downloadmanager.DownloadBatchStatus.Status.DOWNLOADED;
 import static com.novoda.downloadmanager.DownloadBatchStatus.Status.ERROR;
+import static com.novoda.downloadmanager.DownloadBatchStatus.Status.PAUSED;
 
 class NotificationDispatcher {
 
@@ -45,6 +46,8 @@ class NotificationDispatcher {
             if (status == DOWNLOADED) {
                 notificationSeenPersistence.updateNotificationSeenAsync(downloadBatchStatus.getDownloadBatchId(), NOTIFICATION_SEEN);
                 downloadService.stackNotification(notificationInformation);
+            } else if (status == PAUSED) {
+                downloadService.stackNotificationNotDismissable(notificationInformation);
             } else if (status == DELETION || status == ERROR) {
                 downloadService.stackNotification(notificationInformation);
             } else {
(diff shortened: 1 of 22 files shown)


library/src/main/java/com/novoda/downloadmanager/NotificationDispatcher.java
```diff
@@ -29,7 +37,18 @@ void updateNotification(DownloadBatchStatus downloadBatchStatus) {
             NotificationInformation notificationInformation = notificationCreator.createNotification(downloadBatchStatus);
             DownloadBatchStatus.Status status = downloadBatchStatus.status();
 
-            if (status == DOWNLOADED || status == DELETION || status == ERROR) {
+            downloadService.dismissStackedNotification(notificationInformation);
+            if (downloadBatchStatus.notificationSeen()) {
+                Log.v("DownloadBatchStatus: ", downloadBatchStatus.getDownloadBatchId(), " notification has already been seen.");
```
danybony: There is no need to add " " between params of one of the Log methods, they are added automatically when concatenating params.
Just FYI 😉

library/src/test/java/com/novoda/downloadmanager/FakeDownloadsPersistence.java
```diff
@@ -73,4 +73,9 @@ public void delete(DownloadBatchId downloadBatchId) {
     public void update(DownloadBatchId downloadBatchId, DownloadBatchStatus.Status status) {
 
     }
+
+    @Override
+    public void update(DownloadBatchId downloadBatchId, boolean notificationSeen) {
+
```
danybony: All these empty method could use a `// no-op` comment, just to specify that we didn't forget about these

library/src/main/java/com/novoda/downloadmanager/NotificationDispatcher.java
```diff
@@ -29,7 +37,18 @@ void updateNotification(DownloadBatchStatus downloadBatchStatus) {
...
+                return null;
+            }
+
+            if (status == DOWNLOADED) {
+                notificationSeenPersistence.updateNotificationSeenAsync(downloadBatchStatus.getDownloadBatchId(), NOTIFICATION_SEEN);
+                downloadService.stackNotification(notificationInformation);
+            } else if (status == PAUSED) {
+                downloadService.stackNotificationNotDismissible(notificationInformation);
+            } else if (status == DELETION || status == ERROR) {
```
danybony, line 51: Can we somehow test this new behaviour?

danybony (commented):
Looking good, I'd just like some tests around that new part

library/src/main/java/com/novoda/downloadmanager/NotificationDispatcher.java
MechaRyRy: I forgot 😄

danybony (approved):

PR #322, opened 2018-02-01
307/extract strings

### LINKED ISSUE #307: Use resources for defining strings
### Problem
We are hardcoding strings for notifications when we should ideally be using string resources that can be localised. 

### Solution
Use string resources.

tasomaniac:
Extracted the hardcoded Strings as they are. Let's quickly talk about here about maybe slightly modifying them. I think we can be more friendly and curious about what you think.

Fixes #307


danybony (approved):
👍  regarding the strings these are only the default ones. A client would be able to specify the desired text implementing one or more (for migrations vs downloads) `NotificationCreator`s

PR #325, opened 2018-02-02
Stop downloading when the allowed connection type changes

danybony:
Stop downloading when the allowed connection type changes.
It resumes automatically once the allowed connection is changed again or the required connection is activated

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadManager.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadManager.java
@@ -196,7 +196,7 @@ class DownloadManager implements LiteDownloadManagerCommands {
 
         if (!connectionChecker.isAllowedToDownload()) {
             for (DownloadBatch downloadBatch : downloadBatchMap.values()) {
-                downloadBatch.waitForNetwork(); // TODO check what happens for the queued batches
+                downloadBatch.waitForNetwork();
             }
         } else {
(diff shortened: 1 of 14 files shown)


tasomaniac:
Are we supposed to always `stop`? 

I can imagine the user can be on WIFI already and go to settings and choose `Download over WIFI` only and then the download will stop. But the user is already on `WIFI`. 

Am I missing something?

danybony:
You are right, actually if the user is already on the right connection there is no point in stopping / resume... will update

library/src/main/java/com/novoda/downloadmanager/DownloadManager.java
```diff
@@ -193,7 +193,14 @@ public void updateAllowedConnectionType(ConnectionType allowedConnectionType) {
         }
         connectionChecker.updateAllowedConnectionType(allowedConnectionType);
         DownloadsNetworkRecoveryCreator.getInstance().updateAllowedConnectionType(allowedConnectionType);
-        submitAllStoredDownloads(() -> Log.v("Allowed connectionType updated to " + allowedConnectionType + ". All jobs submitted"));
+
+        if (!connectionChecker.isAllowedToDownload()) {
```
tasomaniac: `nit` To have a positive check, we can maybe reverse the if and else. Seems like it would be more clear.

tasomaniac (approved):
The PR seems much more straightforward. But since I don't know the other parts very much, I didn't understand it completely. I only have 1 small comment about clarity. Rest looks good to me.

zegnus (approved):
I'd be removing all the logs

PR #326, opened 2018-02-07
updated dependencys

aclassen:


hal9002:

tasomaniac:
Thanks for taking time to do this. There is an ongoing big refactoring going on and we will release `v2` soon. All the tasks around this big works are labeled with `v2` already. 

Not just the `NotificationChannels`, background services are also problem with Android Oreo. All will be fixed with version 2 and it will be fully compatible with Oreo.

PR #328, opened 2018-02-13
Notification dispatcher test failure

MechaRyRy:
## Problem
Seems that a test on NotificationDispatcher was not updated in #327 but the CI didn't fail 🤔 

## Solution
Fix the broken test! I'll take a look at the CI once this is merged.


Mecharyry:
test this

danybony (approved):
Weird that the CI didn't pick this, but 💯 for the fix

tasomaniac (approved):

PR #329, opened 2018-02-13
319/custom notification channels

### LINKED ISSUE #319: Allow users to provide custom `NotificationChannel`
Right now, the `v2` of the library creates the `NotificationChannel` automatically for Oreo users with bare minimum functionality. 

This takes away the flexibility because `NotificationChannel` class has many other properties and it is impossible for user to configure them. 

Suggested fix: 
Create an overloaded constructor in `NotificationCreator` with `@RequiresApi(O)` to take the whole `NotificationChannel`. Then users will be able to give whatever channel they want.

tasomaniac:
Fixes #319 

### Problem 

The current v2 API didn't allow users to have custom `NotificationChannel` for Android Oreo. It only allowed users to have custom name and importance for the channels created by the library. 

2nd problem I noticed during development was that the channel logic and notification itself was tightly coupled. 
- User has to provide channel information together with `NotificationCustomizer`
- Also `NotificationCreator<T>` class has channel information only used for channel creation which is nothing related to `T` etc. 

### Solution

Introduced `NotificationChannelProvider` package-scoped class to model channel creation. It has 2 different implementation. `Default` one is basic and allows user to just define `name` and `importance` like before. But also has `Oreo` version which has a full reference to `NotificationChannel` instance provided by the user. 

**Note:** This change is breaking the API but I believe in a very good and much more flexible way. 

### Old usage:

```
        NotificationCreator<MigrationStatus> notificationCreator = new NotificationCreator<>(
                this,
                "chocolate",
                "Migration notifications",
                NotificationManagerCompat.IMPORTANCE_DEFAULT,
                notificationCustomizer
        );

        downloadMigrator = DownloadMigratorBuilder.newInstance(this)
                .withNotification(notificationCreator)
                .build();
```

Problems with above ⬆️ 

- breaks the fluent API because new instance of `NotificationCreator` needs to be created.
- No way to provide custom `NotificationChannel`
- You have to provide custom channel params to provide `NotificationCustomizer`.

### Example usages:

**Usage with full custom channel (Only for Oreo)**

```
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
    DownloadManagerBuilder
            .withNotificationChannel(customChannel)
            .withNotification(notificationCustomizer)
            .build();
}
```

**Usage with custom channel with main channel properties**

```
DownloadManagerBuilder
                .withNotificationChannel(
                        "chocolate",
                        "I'm a channel description",
                        NotificationManagerCompat.IMPORTANCE_DEFAULT
                )
                .withNotification(notificationCustomizer)
                .build()
```

**Light usage with no custom channel**

```
DownloadManagerBuilder
                .withNotification(notificationCustomizer)
                .build()
```


library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
```diff
@@ -185,6 +208,9 @@ public DownloadManagerBuilder withCallbackThrottleByProgressIncrease() {
     }
 
     public DownloadManager build() {
+        if (!notificationChannelProvider.channelId().equals(notificationCreator.channelId())) {
+            throw new DeveloperError("Please ensure to pass same Notification channelId when withNotification methods are used.");
```
MechaRyRy: I think `DeveloperError` is used for an internal error inside the library, this can be influenced by the client so should be an `IllegalArgumentException` or something similar. 

"Please ensure to pass the same Notification `channelId` when calling `withNotification`." might slightly better wording 😺

tasomaniac: Makes sense. I was not sure while doing this :)

library/src/main/java/com/novoda/downloadmanager/DefaultNotificationChannelProvider.java
```diff
@@ -0,0 +1,36 @@
...
+        NotificationChannel notificationChannel = new NotificationChannel(channelId, name, importance);
+        notificationManager(context).createNotificationChannel(notificationChannel);
+    }
+
+    @Override
+    public String channelId() {
+        return channelId;
+    }
+
+    private static NotificationManager notificationManager(Context context) {
```
danybony: this can be inlined

library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
tasomaniac: Addressed here https://github.com/novoda/download-manager/pull/329/commits/f26a71ddef1071efb513b202f1e551bc4752338f
Also added the same check to Migrator.

MechaRyRy (approved):

library/src/main/java/com/novoda/downloadmanager/NotificationCreator.java
```diff
@@ -47,8 +35,7 @@ public Notification getNotification() {
         };
     }
 
-    @RequiresApi(Build.VERSION_CODES.O)
-    NotificationChannel createNotificationChannel() {
-        return new NotificationChannel(channelId, userFacingChannelDescription, importance);
+    String channelId() {
+        return channelId;
```
danybony: I'm missing something: where is this used now?
Can we not just use the value set to the channelProvider (thus requiring the client to specify it only once?)

library/src/main/java/com/novoda/downloadmanager/OreoNotificationChannelProvider.java
```diff
@@ -0,0 +1,31 @@
...
+    public void registerNotificationChannel(Context context) {
+        notificationManager(context).createNotificationChannel(notificationChannel);
+    }
+
+    @Override
+    public String channelId() {
+        return notificationChannel.getId();
+    }
+
+    private static NotificationManager notificationManager(Context context) {
```
danybony: inline?

danybony (commented):
A couple of questions/suggestions

library/src/main/java/com/novoda/downloadmanager/DefaultNotificationChannelProvider.java
tasomaniac: I thought it looks better like this. Also because of the `getSystemService` call I need to suppress the nullibility check because `getSystemService` returns Nullable. Should I still inline in that case?

library/src/main/java/com/novoda/downloadmanager/NotificationCreator.java
tasomaniac: That would mean that I would need an instance of the provider here. Let me see what I can do for that.

library/src/main/java/com/novoda/downloadmanager/DefaultNotificationChannelProvider.java
danybony: Usually I don't like this kind of one-line / one-usage function, I'd be still up to the inlining, but up to you, not a blocker

tasomaniac: done. Especially since it is a library, I think it makes sense to inline.

library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
tasomaniac: After applying @danybony 's comments, this check was not necessary at all.

danybony (approved):

PR #331, opened 2018-02-14
Client specifies notification stackable or not

MechaRyRy:
## Problem 
Currently the client is not able to specify whether a notification should display as stackable or non-stackable, as well as stating whether a notification should be dismissible. 

## Solution
Introduce a `NotificationStackState` enum that defines:
`SINGLE_PERSISTENT_NOTIFICATION`
`STACK_NOTIFICATION_NOT_DISMISSIBLE`
`STACK_NOTIFICATION_DISMISSIBLE`

which when dispatching notifications will determine what method is called on the corresponding `Service`. 

## Future work
- Introduce a new interface that is used to collate the different types of notifications that the `download-manager` supports. Attach this new interface to both the `DownloadService` and the `MigrationService`.
- Ensure that the `MigrationService` deals with all `NotificationStackState`s.
- Repurpose `NotificationDispatcher` that checks the `NotificationStackState` and forwards to the appropriate `Service` method to display the notification.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
+++ b/library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
@@ -14,8 +14,6 @@ import android.util.Log;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
-import static com.novoda.downloadmanager.MigrationStatus.Status;
-
 public class LiteDownloadMigrationService extends Service implements DownloadMigrationService {
 
     private static final String TAG = "MigrationService";
@@ -78,7 +76,7 @@ public class LiteDownloadMigrationService extends Service implements DownloadMig
         public void onUpdate(MigrationStatus migrationStatus) {
             NotificationInformation notification = notificationCreator.createNotification(migrationStatus);
 
-            if (migrationStatus.status() == Status.COMPLETE || migrationStatus.status() == Status.DB_NOT_PRESENT) {
+            if (notification.isStackable()) {
                 stackNotification(notification);
             } else {
                 updateNotification(notification);
--- a/library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
+++ b/library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
@@ -14,6 +14,8 @@ import android.util.Log;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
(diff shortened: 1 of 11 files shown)


library/src/main/java/com/novoda/downloadmanager/NotificationDispatcher.java
```diff
@@ -45,11 +44,12 @@ void updateNotification(DownloadBatchStatus downloadBatchStatus) {
 
             if (status == DOWNLOADED) {
                 notificationSeenPersistence.updateNotificationSeenAsync(downloadBatchStatus.getDownloadBatchId(), NOTIFICATION_SEEN);
+            }
+
+            if (notificationInformation.notificationStackState() == STACK_NOTIFICATION_DISMISSIBLE) {
```
tasomaniac: Since this is an enum now, wouldn't it be better to use `switch` here?

MechaRyRy: edaaa50

MechaRyRy:
Updating the NotificationDispatcherTests

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
```diff
@@ -78,10 +76,21 @@ private void createNotificationChannel() {
...
-            } else {
-                updateNotification(notification);
+            switch (notification.notificationStackState()) {
+                case SINGLE_PERSISTENT_NOTIFICATION:
+                    updateNotification(notification);
+                    break;
+                case STACK_NOTIFICATION_DISMISSIBLE:
+                    stackNotification(notification);
+                    break;
+                case STACK_NOTIFICATION_NOT_DISMISSIBLE:
```
danybony, line 86: Isn't this a condition the client could create?

danybony (commented):
LGTM, just one question/comment.

Also: do you think it's worth adding a forth `NotificationStackState` for a "ignored/not shown" notification? doing so we would allow clients to not show notifications for some cases (like for db not present in migration)

library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
MechaRyRy: Yes, I've left this for now because in a follow up PR I would like to do the following:

> Introduce a new interface that is used to collate the different types of notifications that the download-manager supports. Attach this new interface to both the DownloadService and the MigrationService.

MechaRyRy:
Adding the extra state is a great idea too! That's the good thing about the enum, can add in the next PR when I start sorting the `Service`s more!

danybony (approved):
Cool, let's keep these points for the next PR 👍

tasomaniac (approved):

PR #332, opened 2018-02-15
Generic NotificationDispatcher

MechaRyRy:
Trying to keep these PRs small because I can 😄 should make reviews quicker and easier, hopefully 💃 

## Problem 
The Migration and standard download functionality have more or less the same logic for receiving a `payload` and dispatching `notifications` but they are represented by two different classes 😬 

## Solution
Create a generic `NotificationDispatcher<T>` that can be used by both the Migration service and the standard Download service. 

Created a clone of the `NotificationDispatcher` currently in the epic branch and made it generic. Added this generic `NotificationDispatcher` to the `DownloadBatchStatusNotificationDispatcher`. `DownloadBatchStatusNotificationDispatcher` now deals with the code specific for `DownloadBatchStatus`, so:
- Do nothing when a notification has already been seen (DOWNLOADED)
- Mark as seen when the status is `DOWNLOADED`
- Dispatch to the generic `NotificationDispatcher<T>` for all other occurrences. 

## Tests Added
Split the tests that were originally in the `NotificationDispatcher` into two. One, the persistence and notification seen check tests have been moved to the `DownloadBatchStatusNotificationDispatcherTest` class. Two, the remaining checks on `NotificationStackState` now exist in the `NotificationDispatcherTest` class.

## Follow up
- Create a generic interface for the Services, Download and Migration, that represents just the displaying of `Notifications`.
- Use the generic `NotificationDispatcher<T>` in the `LiteMigrationService` where we do the same checks for `NotificationStackState` and display notifications.
- Add additional state that represents `do not display anything` 😄

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatchStatusNotificationDispatcher.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatchStatusNotificationDispatcher.java
@@ -1,7 +1,5 @@
 package com.novoda.downloadmanager;
 
-import android.support.annotation.WorkerThread;
-
 import com.novoda.notils.logger.simple.Log;
 
 import static com.novoda.downloadmanager.DownloadBatchStatus.Status.DOWNLOADED;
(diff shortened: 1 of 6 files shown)


danybony (approved):
💯

PR #333, opened 2018-02-15
Move notification display from service

MechaRyRy:
⚠️ Added the `discussion` label as I am still not 100% sure if this is the best approach ⚠️ 

## Problem
The display of notifications is tied to an Android `Service` and could vary between `Service`s if they both handle notifications 😬 

## Solution
Move logic used to display notifications on a `Service` to one place so that there will be no variation in how these operate between the two `Service`s. The `ServiceNotificationDispatcher` knows about an Android `Service` and the `NotificationManager` in order to perform the display for all `Service`s in the library.

This `NotificationDispatcher` is dependent on the `Wait.waitFor` which only waits on the `DownloadService`, this has been made more generic to wait for any object that we need it to wait for in the library, so it can be used for the `MigrationService` too. 

I'm wondering that if we use this `Wait.waitFor` on the migration service if we can flip the dependencies slightly so that it operates similar to the `DownloadService`:

Current: We create a service and in the service we start a job
Future: We could create a migration job that waits for the Service to exist before kicking off a job.


library/src/main/java/com/novoda/downloadmanager/Wait.java
```diff
@@ -0,0 +1,52 @@
...
+
+import com.novoda.notils.logger.simple.Log;
+
+final class Wait {
+
+    private Wait() {
+        // Uses static factory method.
+    }
+
+    static <T> ThenPerform<T> waitFor(@Nullable Object instanceToWaitFor, Object lock) {
```
MechaRyRy: Wait is now more generic and can be used for any object in the library for flexibility. We could interface this out to enforce type safety but this would mean casting in some places where we rely on a different scope.

library/src/test/java/com/novoda/downloadmanager/ServiceNotificationDispatcherTest.java
```diff
@@ -0,0 +1,107 @@
...
+import static com.novoda.downloadmanager.NotificationCustomizer.NotificationStackState.STACK_NOTIFICATION_NOT_DISMISSIBLE;
+import static com.novoda.downloadmanager.NotificationInformationFixtures.notificationInformation;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.BDDMockito.given;
+import static org.mockito.Mockito.inOrder;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
+
+public class ServiceNotificationDispatcherTest {
```
MechaRyRy, line 25: Just a rename

library/src/main/java/com/novoda/downloadmanager/ServiceNotificationDispatcher.java
```diff
@@ -0,0 +1,84 @@
...
+        Notification notification = notificationInformation.getNotification();
+        notification.flags |= Notification.FLAG_ONGOING_EVENT;
+        notificationManager.notify(NOTIFICATION_TAG, notificationInformation.getId(), notification);
+    }
+
+    private void dismissStackedNotification(NotificationInformation notificationInformation) {
+        notificationManager.cancel(NOTIFICATION_TAG, notificationInformation.getId());
+    }
+
+    void setService(Service service) {
```
MechaRyRy: We could create a new interface rather than relying on the Android `Service` scope.

library/src/main/java/com/novoda/downloadmanager/ServiceNotificationDispatcher.java
```diff
@@ -0,0 +1,84 @@
+package com.novoda.downloadmanager;
+
+import android.app.Notification;
+import android.app.Service;
+import android.support.annotation.WorkerThread;
+
+class ServiceNotificationDispatcher<T> {
```
MechaRyRy: The plan is to use this for the `Migration` Service as well.

library/src/main/java/com/novoda/downloadmanager/NotificationManager.java
```diff
@@ -0,0 +1,22 @@
+package com.novoda.downloadmanager;
+
+import android.app.Notification;
+import android.support.annotation.Nullable;
+import android.support.v4.app.NotificationManagerCompat;
+
+class NotificationManager {
```
MechaRyRy: Wrapping so that we can mock and test the `ServiceNotificationDispatcher`

library/src/main/java/com/novoda/downloadmanager/DownloadBatchStatusNotificationDispatcher.java
```diff
@@ -30,6 +32,15 @@ void updateNotification(DownloadBatchStatus downloadBatchStatus) {
     }
 
     void setDownloadService(DownloadService downloadService) {
-        notificationDispatcher.setDownloadService(downloadService);
+        if (downloadService instanceof Service) {
```
MechaRyRy: Adding a new interface that `DownloadService` and `MigrationService` extends from may alleviate instance check issue.

library/src/main/java/com/novoda/downloadmanager/NotificationManager.java
tasomaniac: can't we mock `notificationManagerCompat` in tests? This wrapping seems to be unnecessary.

library/src/main/java/com/novoda/downloadmanager/Wait.java
```diff
@@ -0,0 +1,52 @@
...
+
+        private final Object instanceToWaitFor;
+        private final Object lock;
+
+        ThenPerform(Object instanceToWaitFor, Object lock) {
+            this.instanceToWaitFor = instanceToWaitFor;
+            this.lock = lock;
+        }
+
+        T thenPerform(Action<T> action) {
```
tasomaniac, line 31: I saw `return null` above because of this method which was used in a case where we don't care about the return type.

Our `Action` here is more like `Callable` from java. And we may make use of `Runnable` in case where we don't care about the output. This means we would have these 2 methods. And we don't have to `return null`  anymore.

```
T thenPerform(Callable<T> callable)

void thenPerform(Runnable runnable)
``

File: library/src/main/java/com/novoda/downloadmanager/ServiceNotificationDispatcher.java (hunk 3 above)
Comment (danybony) [reply library/src/main/java/com/novoda/downloadmanager/ServiceNotificationDispatcher.java]: Yes definitely... as you mentioned in other comments, but here is pretty evident

File: library/src/main/java/com/novoda/downloadmanager/Wait.java (hunk 1 above)
Comment (danybony) [reply library/src/main/java/com/novoda/downloadmanager/Wait.java]: yeah, I don't really like the use of all these `Object`s...
Just making a test with interfaces, is it gonna be too limiting?

File: library/src/main/java/com/novoda/downloadmanager/Wait.java (hunk 7 above)
Comment (MechaRyRy) [reply library/src/main/java/com/novoda/downloadmanager/Wait.java:31]: Happy to do this in a follow-up PR I am having issues with lint when doing this and I'm not sure why atm. I think it's failing for some reason with the lambdas

**Review by tasomaniac — APPROVED** [review]

File: library/src/main/java/com/novoda/downloadmanager/Wait.java (hunk 7 above)
Comment (tasomaniac) [reply library/src/main/java/com/novoda/downloadmanager/Wait.java:31]: Sounds good 👍 It is a little improvement imho. It's also fine without it.

File: library/src/main/java/com/novoda/downloadmanager/NotificationManager.java (hunk 5 above)
Comment (MechaRyRy) [reply library/src/main/java/com/novoda/downloadmanager/NotificationManager.java]: No, it's final unfortunately. I did take a look 😢

Comment (tasomaniac) [reply library/src/main/java/com/novoda/downloadmanager/NotificationManager.java]: `Mockito` is able to mock final classes since version 2.x
https://github.com/mockito/mockito/wiki/What%27s-new-in-Mockito-2#unmockable

File: library/src/main/java/com/novoda/downloadmanager/ServiceNotificationDispatcher.java (hunk 3 above)
Comment (MechaRyRy) [reply library/src/main/java/com/novoda/downloadmanager/ServiceNotificationDispatcher.java]: 79e4494

File: library/src/main/java/com/novoda/downloadmanager/Wait.java (hunk 1 above)
Comment (MechaRyRy) [reply library/src/main/java/com/novoda/downloadmanager/Wait.java]: 79e4494

File: library/src/main/java/com/novoda/downloadmanager/LiteDownloadService.java [hunk 8]
```diff
@@ -17,19 +16,26 @@
...
     public void onCreate() {
         super.onCreate();
         executor = Executors.newSingleThreadExecutor();
         binder = new DownloadServiceBinder();
-        notificationManagerCompat = NotificationManagerCompat.from(this);
+    }
+
+    @Override
+    public void start(int id, Notification notification) {
+        startForeground(id, notification);
```
MechaRyRy, line 33: This is the only weird thing now I guess @danybony but probably a fine tradeoff

danybony: yeah, I think it's acceptable 👍

danybony (approved):

library/src/main/java/com/novoda/downloadmanager/Wait.java
Mecharyry: Ah good point! Forgot they added that incubating feature. d352695

library/src/main/java/com/novoda/downloadmanager/NotificationManager.java
MechaRyRy: Ah good point! Forgot they added that incubating feature. d352695

tasomaniac:
Awesome :100:

PR #334, opened 2018-02-15
Duplicate files across batches

danybony:
Create a folder for every batch, containing all the relative files.
The folder uses the batch id as name, sanitised so that it doesn't contain not-allowed chars.
Updated the Main Activity log method to display all the files also in subfolders, so that now we have this log:

```
LogFileDirectory /data/user/0/com.novoda.downloadmanager.demo.simple/files/batch_id_1/foo/bar/5mb.zip 
LogFileDirectory /data/user/0/com.novoda.downloadmanager.demo.simple/files/batch_id_1/10MB.zip 
LogFileDirectory /data/user/0/com.novoda.downloadmanager.demo.simple/files/batch_id_2/10MB.zip 
LogFileDirectory /data/user/0/com.novoda.downloadmanager.demo.simple/files/batch_id_2/20MB.zip 
```


MechaRyRy (approved):
Does this also work for the migrations?

danybony:
Checking for migrations...

danybony:
Checked that the batch id prefix is not needed when migrating, but fixed a bug in which we were deleting files for completed downloads when migrating

PR #335, opened 2018-02-15
Make download error public so that it can be checked by the client

zegnus:
When a download batch status is triggered by our update callbacks, it could be possible that the batch has an error.

We need this error to be public so that a client can check it out.


MechaRyRy:
Shouldn't you make the `error()` getter public too?

zegnus:
@Mecharyry is not needed, we only need the type itself, the get is for internal use to return the error type from `getDownloadErrorType()`. As this method returns directly the `Error` enum, there is no need to make the `error()` also public, we can keep it private

MechaRyRy (approved):

danybony (approved):

PR #336, opened 2018-02-16
Align Services

MechaRyRy:
## Problem
Both of the `Service`s in the `download-manager` behave in different ways in terms of performing an action and reporting the status of that action. Currently the `MigrationService` creates and binds all of it's dependencies and kicks off the migration process. All status events come back to this service at which point it determines which notifications to display and which statuses to forward to the activity. 😬 

## Solution
Change the flow of the `MigrationService` so that the `MigrationBuilder` creates and binds all of the dependencies to the `LiteDownloadMigrator`. The `LiteDownloadMigrator` will trigger a `MigrationJob` when the `Service` is bound using the `Wait` mechanism. The `LiteDownloadMigrator` receives all callbacks and forwards to the `NotificationDispatcher` and the  `MainActivity` if a callback is registered. 

Important to note that the `MigrationService` essentially operates in the same way as the `DownloadService` now.

## Follow-up work
Introduce additional state that represents hiding a notification.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadMigratorBuilder.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadMigratorBuilder.java
@@ -30,7 +30,6 @@ public final class DownloadMigratorBuilder {
 
     private NotificationChannelProvider notificationChannelProvider;
     private NotificationCreator<MigrationStatus> notificationCreator;
-    private DownloadMigrationService migrationService;
     private LiteDownloadMigrator downloadMigrator;
     private MigrationCallback migrationCallback;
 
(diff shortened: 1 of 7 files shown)


library/src/main/java/com/novoda/downloadmanager/DownloadMigratorBuilder.java
```diff
@@ -34,19 +48,28 @@ public static DownloadMigratorBuilder newInstance(Context context) {
                 customizer,
                 notificationChannelProvider
         );
-
         Handler handler = new Handler(Looper.getMainLooper());
-        return new DownloadMigratorBuilder(applicationContext, handler, notificationChannelProvider, defaultNotificationCreator);
+        MigrationCallback migrationCallback = (status) -> {
+            Log.v(status.toString());
```
zegnus, line 53: we might not need this for prod release

zegnus (approved):

library/src/main/java/com/novoda/downloadmanager/DownloadMigratorBuilder.java
MechaRyRy: Yea I was thinking we should do something similar for this as we did for `Merlin` https://github.com/novoda/merlin/blob/master/core/src/main/java/com/novoda/merlin/Logger.java

Where clients can specify a log handle and if they do then they receive debug messaging. Would be super useful if anyone ever opens an issue.

danybony (approved):

PR #337, opened 2018-02-16
Hidden state

MechaRyRy:
## Problem
All of our `NotificationDisplayState`s assume that the client wants to display a notification 😱 

## Solution
Add an additional state that will not show a notification for the currently emitted status, `DownloadBatchStatus` or `MigrationStatus`.

Also updated the `DownloadBatchStatus` side to ensure that `PAUSED` batch notifications are dismissible. 

`NotificationStackState` has been renamed to `NotificationDisplayState` because it's not just the Stack state. 

#### Paused

Before | After
--- | ---
before_paused | after_paused

#### DB migration

Before | After
--- | ---
before_db_migrate | after_db_migrate

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
@@ -330,10 +330,8 @@ public final class DownloadManagerBuilder {
         @Override
         public NotificationStackState notificationStackState(DownloadBatchStatus payload) {
             DownloadBatchStatus.Status status = payload.status();
-            if (status == DOWNLOADED || status == DELETED || status == ERROR) {
+            if (status == DOWNLOADED || status == DELETED || status == ERROR || status == PAUSED) {
                 return NotificationStackState.STACK_NOTIFICATION_DISMISSIBLE;
-            } else if (status == PAUSED) {
-                return NotificationStackState.STACK_NOTIFICATION_NOT_DISMISSIBLE;
             } else {
                 return NotificationStackState.SINGLE_PERSISTENT_NOTIFICATION;
             }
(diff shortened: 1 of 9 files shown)


library/src/main/java/com/novoda/downloadmanager/ServiceNotificationDispatcher.java
```diff
@@ -4,6 +4,8 @@
 import android.support.annotation.WorkerThread;
 import android.support.v4.app.NotificationManagerCompat;
 
+import com.novoda.notils.logger.simple.Log;
```
zegnus, line 7: Probably not needed ;)

zegnus (approved):

library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
```diff
@@ -328,14 +328,12 @@ private CallbackThrottleCreator getCallbackThrottleCreator(CallbackThrottleCreat
...
 
         @Override
-        public NotificationStackState notificationStackState(DownloadBatchStatus payload) {
+        public NotificationDisplayState notificationDisplayState(DownloadBatchStatus payload) {
             DownloadBatchStatus.Status status = payload.status();
-            if (status == DOWNLOADED || status == DELETED || status == ERROR) {
-                return NotificationStackState.STACK_NOTIFICATION_DISMISSIBLE;
-            } else if (status == PAUSED) {
-                return NotificationStackState.STACK_NOTIFICATION_NOT_DISMISSIBLE;
+            if (status == DOWNLOADED || status == DELETED || status == ERROR || status == PAUSED) {
```
danybony, line 333: 👍

danybony (approved):

zegnus (approved):

PR #338, opened 2018-02-19
Use file system for file size

MechaRyRy:
## Problem
- When performing a migration from the demo the db transitions to error when in airplane mode. This occurs because the library determines the file size as being incorrect 😬 after some investigation this occurs because we rely on the v1 db as the source of truth for a file size when we should use the file system instead 🎉 

- Notifications are showing for all migrated downloads when restarting the demo application. This will affect production too. This happens because we are marking downloads from the migrator with the `notificationSeen = false` 😬 

## Solution
- Use the file system as a means to determining a file size.

- All migrated files should be marked with `notificationSeen = true` to prevent notifications showing on app restart.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
+++ b/library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
@@ -15,15 +15,16 @@ class MigrationExtractor {
     private static final int TITLE_COLUMN = 1;
     private static final int MODIFIED_TIMESTAMP_COLUMN = 2;
 
-    private static final String DOWNLOADS_QUERY = "SELECT uri, _data, total_bytes FROM Downloads WHERE batch_id = ?";
+    private static final String DOWNLOADS_QUERY = "SELECT uri, _data FROM Downloads WHERE batch_id = ?";
     private static final int NETWORK_ADDRESS_COLUMN = 0;
     private static final int FILE_LOCATION_COLUMN = 1;
-    private static final int FILE_SIZE_COLUMN = 2;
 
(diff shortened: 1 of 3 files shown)


library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
```diff
@@ -16,7 +16,7 @@
 
     private static final String TABLE_BATCHES = "batches";
     private static final String WHERE_CLAUSE_ID = "_id = ?";
-    private static final boolean NOTIFICATION_NOT_SEEN = false;
+    private static final boolean NOTIFICATION_SEEN = true;
```
MechaRyRy, line 19: Fix two, EZPZ.

library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
```diff
@@ -69,7 +70,8 @@
                 String originalFileLocation = downloadsCursor.getString(FILE_LOCATION_COLUMN);
                 newBatchBuilder.addFile(originalNetworkAddress).apply();
 
-                long rawFileSize = downloadsCursor.getLong(FILE_SIZE_COLUMN);
+                FilePath filePath = new LiteFilePath(originalFileLocation);
+                long rawFileSize = internalFilePersistence.getCurrentSize(filePath);
```
MechaRyRy, line 74: Rely on the file system and not the db.

rock3r (approved):
Noice! EZPZ indeed

danybony (approved):
nice 👍

PR #339, opened 2018-02-19
Check whether a wakeLock is held before releasing

MechaRyRy:
## Problem
We aren't checking whether we hold a wake lock before attempting to release it, this can lead to a crash 😬 

## Solution
Check whether we hold a given wake lock before attempting to release it.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
+++ b/library/src/main/java/com/novoda/downloadmanager/LiteDownloadMigrationService.java
@@ -61,7 +61,7 @@ public class LiteDownloadMigrationService extends Service implements DownloadMig
         executor.execute(() -> {
             acquireCpuWakeLock();
             migrationJob.run();
-            releaseCpuWakeLock();
+            releaseHeldCpuWakeLock();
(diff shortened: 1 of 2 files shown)


MechaRyRy:
Wondering if we should also attempt to catch the exception?

danybony (approved):
I wouldn't try to catch the exception.
With this check we should be safe from un-held wakelocks, if something else it's happening it's not something we can control/recover

rock3r (approved):
I'm against swallowing the exception too; I am ok with the fix but I am wondering how on earth would we not have the wakelock at that point? Would it be because of App Standby or Doze not granting us wakelocks?

MechaRyRy:
We've seen this on Samsung devices :trollface: it seems to be from bugs in other libraries 😬 

https://stackoverflow.com/questions/12140844/java-lang-runtimeexception-wakelock-under-locked-c2dm-lib

https://github.com/evernote/android-job/issues/29

We've added it here to try and guard against this happening.

rock3r:
Oh nice 🙄 Well let's ship!

PR #340, opened 2018-02-20
Expose downloads dir

danybony:
A client might want to keep download files separate from other files stored in internal storage, so by default we store files in a `downloads` directory, both for internal and external storage.
A client can always override this creating a custom FilePersistence.
Expose that directory through the `LiteDownloadManagerCommands` interface, for easier access for the clients

### DIFF
--- a/demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
+++ b/demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
@@ -152,8 +152,12 @@ public class MainActivity extends AppCompatActivity {
     };
 
     private final View.OnClickListener logFileDirectoryOnClick = v -> {
-        File[] files = getFilesDir().listFiles();
-        logAllFiles(files);
+        LiteDownloadManagerCommands downloadManagerCommands = ((DemoApplication) getApplication()).getLiteDownloadManagerCommands();
+        File downloadsDir = downloadManagerCommands.getDownloadsDir();
+        Log.d("LogFileDirectory", "Downloads dir:", downloadsDir.getAbsolutePath());
+        if (downloadsDir.exists()) {
+            logAllFiles(downloadsDir.listFiles());
+        }
     };
 
     private void logAllFiles(File[] files) {
--- a/library/src/main/java/com/novoda/downloadmanager/ExternalFilePersistence.java
+++ b/library/src/main/java/com/novoda/downloadmanager/ExternalFilePersistence.java
@@ -30,7 +30,7 @@ class ExternalFilePersistence implements FilePersistence {
 
     @Override
     public FilePath basePath() {
-        return FilePathCreator.create(getExternalFileDirWithBiggerAvailableSpace().getAbsolutePath(), "/");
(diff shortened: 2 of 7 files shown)


demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -152,8 +152,12 @@ protected void onCreate(Bundle savedInstanceState) {
     };
 
     private final View.OnClickListener logFileDirectoryOnClick = v -> {
-        File[] files = getFilesDir().listFiles();
-        logAllFiles(files);
+        LiteDownloadManagerCommands downloadManagerCommands = ((DemoApplication) getApplication()).getLiteDownloadManagerCommands();
+        File downloadsDir = downloadManagerCommands.getDownloadsDir();
+        Log.d("LogFileDirectory", "Downloads dir:", downloadsDir.getAbsolutePath());
+        if (downloadsDir.exists()) {
+            logAllFiles(downloadsDir.listFiles());
```
zegnus, line 159: :100:

MechaRyRy (approved):

library/src/main/java/com/novoda/downloadmanager/ExternalFilePersistence.java
```diff
@@ -30,7 +30,7 @@ public void initialiseWith(Context context) {
 
     @Override
     public FilePath basePath() {
-        return FilePathCreator.create(getExternalFileDirWithBiggerAvailableSpace().getAbsolutePath(), "/");
+        return FilePathCreator.create(getExternalFileDirWithBiggerAvailableSpace().getAbsolutePath(), "/downloads/");
```
zegnus: extract as constant?

library/src/main/java/com/novoda/downloadmanager/InternalFilePersistence.java
```diff
@@ -25,7 +25,7 @@ public void initialiseWith(Context context) {
 
     @Override
     public FilePath basePath() {
-        return FilePathCreator.create(context.getFilesDir().getAbsolutePath(), "/");
+        return FilePathCreator.create(context.getFilesDir().getAbsolutePath(), "/downloads/");
     }
```
zegnus: extract as constant?

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerCommands.java
```diff
@@ -33,4 +34,6 @@
     void getDownloadStatusWithMatching(DownloadFileId downloadFileId, DownloadFileStatusCallback callback);
 
     void updateAllowedConnectionType(ConnectionType allowedConnectionType);
+
+    File getDownloadsDir();
```
zegnus, line 38: :100:

zegnus (approved):
LGTM, the directory could also be specified in the constructor as a param but as we have a `getDir` might not be that important.

I have noticed though that the MigratorJob:44 creates an internal storage instance directly, do you mind to use `FilePersistenceCreator.newInternalFilePersistenceCreator(context).create();` instead?

PR #341, opened 2018-02-20
On network error, not all scenarios recovers

zegnus:
When a batch has a network error, we currently submit the batch for recovery but we do not mark the batch as queued and keeps with an error.

The fix is to ensure that the batch is marked as queued and submitted for recovery when there is a network error. This solves all situations and the batch is able to continue even when there is an HTTP error produced while switching to airplane mode.

There was also a problem when retrieving the full size of the batch. When there was a network error we were not submitting the batch for recovery so that the batch was lost forever.

after mp4


MechaRyRy (approved):

danybony (approved):
👍

PR #343, opened 2018-02-21
Check if a file is fully downloaded before creating a new file

zegnus:
**Problem**
Given one download has fully finished, and a second download runs out of space
when the download manager restarts
then the downloaded batch appears also as a failure

**Solution**
I have moved the check for the fully downloaded batch before the creation of the file, in the way I have realised that the `basePath` from the getter in the `FilePersistenceResult` has the same value passed in the constructor, so I've eliminated it altogether

### DIFF
--- a/demo-simple/src/main/java/com/novoda/downloadmanager/demo/CustomFilePersistence.java
+++ b/demo-simple/src/main/java/com/novoda/downloadmanager/demo/CustomFilePersistence.java
@@ -30,7 +30,7 @@ public class CustomFilePersistence implements FilePersistence {
     @Override
     public FilePersistenceResult create(FilePath absoluteFilePath, FileSize fileSize) {
         Log.v("create " + absoluteFilePath.toString() + ", " + fileSize.toString());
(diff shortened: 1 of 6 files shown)


demo-simple/src/main/java/com/novoda/downloadmanager/demo/CustomFilePersistence.java
```diff
@@ -30,7 +30,7 @@ public FilePath basePath() {
     @Override
     public FilePersistenceResult create(FilePath absoluteFilePath, FileSize fileSize) {
         Log.v("create " + absoluteFilePath.toString() + ", " + fileSize.toString());
-        return FilePersistenceResult.newInstance(FilePersistenceResult.Status.SUCCESS);
+        return FilePersistenceResult.SUCCESS;
```
MechaRyRy, line 33: 👍

MechaRyRy (approved):

danybony (approved):

PR #345, opened 2018-02-22
Stop service in case of hidden notification

danybony:
This is done in order to dismiss previously ongoing notifications, for example when coming from a DOWNLOADING state (with not dismissible notification) to a PAUSED one (which a client might not want to display).

Before stopping the service (and then dismissing persistent notification) we check the notification id, and we stop only if the current one is the same as the last persisted notification.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/ServiceNotificationDispatcher.java
+++ b/library/src/main/java/com/novoda/downloadmanager/ServiceNotificationDispatcher.java
@@ -37,7 +37,7 @@ class ServiceNotificationDispatcher<T> {
 
             switch (notificationInformation.notificationDisplayState()) {
                 case SINGLE_PERSISTENT_NOTIFICATION:
-                    updateNotification(notificationInformation);
+                    updatePersistentNotification(notificationInformation);
                     break;
                 case STACK_NOTIFICATION_NOT_DISMISSIBLE:
                     stackNotificationNotDismissible(notificationInformation);
@@ -65,7 +65,7 @@ class ServiceNotificationDispatcher<T> {
         notificationManager.cancel(NOTIFICATION_TAG, notificationInformation.getId());
     }
 
-    private void updateNotification(NotificationInformation notificationInformation) {
+    private void updatePersistentNotification(NotificationInformation notificationInformation) {
         persistentNotificationId = notificationInformation.getId();
         service.start(notificationInformation.getId(), notificationInformation.getNotification());
     }
--- a/library/src/main/java/com/novoda/downloadmanager/ServiceNotificationDispatcher.java
+++ b/library/src/main/java/com/novoda/downloadmanager/ServiceNotificationDispatcher.java
@@ -4,8 +4,6 @@ import android.app.Notification;
 import android.support.annotation.WorkerThread;
 import android.support.v4.app.NotificationManagerCompat;
 
-import com.novoda.notils.logger.simple.Log;
-
 class ServiceNotificationDispatcher<T> {
 
     private static final String NOTIFICATION_TAG = "download-manager";
@@ -47,7 +45,7 @@ class ServiceNotificationDispatcher<T> {
                     stackNotification(notificationInformation);
                     break;
                 case HIDDEN_NOTIFICATION:
-                    Log.d("Notification not required, hiding.");
+                    service.stop(true);
                     break;
                 default:
                     String message = String.format(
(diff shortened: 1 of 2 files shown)


library/src/main/java/com/novoda/downloadmanager/ServiceNotificationDispatcher.java
```diff
@@ -62,25 +61,32 @@ void updateNotification(T payload) {
         };
     }
 
+    private void dismissStackedNotification(NotificationInformation notificationInformation) {
+        notificationManager.cancel(NOTIFICATION_TAG, notificationInformation.getId());
+    }
+
     private void updateNotification(NotificationInformation notificationInformation) {
```
rock3r: Maybe we can make it clearer in this method's name that it only deals with persistent notifications? Otherwise it may seem a bug that it always saves the ID without checking if it is a persistent notification. I had the doubt and had to read the rest of the class to understand how it's used.

library/src/test/java/com/novoda/downloadmanager/ServiceNotificationDispatcherTest.java
```diff
@@ -44,70 +34,122 @@
...
     @Test
     public void stacksDismissibleNotification() {
-        given(notificationCreator.createNotification(DOWNLOAD_BATCH_STATUS)).willReturn(STACKABLE_DISMISSIBLE_NOTIFICATION_INFORMATION);
+        NotificationInformation notificationInfo = createNotificationInfo(STACK_NOTIFICATION_DISMISSIBLE, 100);
+        given(notificationCreator.createNotification(DOWNLOAD_BATCH_STATUS)).willReturn(notificationInfo);
+
+        notificationDispatcher.updateNotification(DOWNLOAD_BATCH_STATUS);
+
+        InOrder inOrder = inOrder(downloadService, notificationManager);
+        inOrder.verify(downloadService, never()).stop(true);
```
rock3r: Do you need the `inOrder` here? You're checking for a `never` and then for a single call. Wouldn't it be the same if they were not in an `inOrder`?

library/src/test/java/com/novoda/downloadmanager/ServiceNotificationDispatcherTest.java
```diff
@@ -44,70 +34,122 @@
...
     @Test
     public void stacksNonDismissibleNotification() {
-        given(notificationCreator.createNotification(DOWNLOAD_BATCH_STATUS)).willReturn(STACKABLE_NON_DISMISSIBLE_NOTIFICATION_INFORMATION);
+        NotificationInformation notificationInfo = createNotificationInfo(STACK_NOTIFICATION_NOT_DISMISSIBLE, 100);
+        given(notificationCreator.createNotification(DOWNLOAD_BATCH_STATUS)).willReturn(notificationInfo);
+
+        notificationDispatcher.updateNotification(DOWNLOAD_BATCH_STATUS);
+
+        InOrder inOrder = inOrder(downloadService, notificationManager);
+        inOrder.verify(downloadService, never()).stop(true);
```
rock3r: Ditto here

rock3r (changes requested):
Nice! I left a couple of notes, nothing too serious but I'd like them to be addressed (even just with a "nah" 🤣)

library/src/test/java/com/novoda/downloadmanager/ServiceNotificationDispatcherTest.java
danybony: Yeah, in the tests with the `never` the `inOrder` can be omitted... will remove

library/src/test/java/com/novoda/downloadmanager/ServiceNotificationDispatcherTest.java
danybony: 797d951

library/src/test/java/com/novoda/downloadmanager/ServiceNotificationDispatcherTest.java
danybony: 797d951

library/src/main/java/com/novoda/downloadmanager/ServiceNotificationDispatcher.java
danybony: 0f654ac

rock3r (approved):
Super quick turnaround, thanks! 💯

MechaRyRy (approved):
👌 super good

PR #346, opened 2018-02-23
Submit all stored downloads after migration completed

danybony:
This is needed because otherwise the migrated downloads are not gonna be available unless the user restarts the app


zegnus (approved):

rock3r (approved):
EZ, should've been more explicit it was just about the demo tho :)

danybony:
@rock3r the visibility modification is needed for clients too, that one is in the library.
Speaking of which I opened this Issue to consider moving this logic in the library entirely https://github.com/novoda/download-manager/issues/347

rock3r:
What I meant is that the changes are mostly on the client side 💯 on the issue tho!

PR #348, opened 2018-02-23
Use synchronized block for modifying the callback list

MechaRyRy:
Oh boy, deja vu 😄 

### Problem
Turns out the `CopyOnWriteArrayList` didn't work as we envisioned 😬 

### Solution
Go back to using the `Synchronised` lock 😄


rock3r (approved):
Ok, but... what was the issue with the COWAL?

MechaRyRy:
> COWAL 

Sorry @rock3r, we were still experiencing the `ConcurrentModificationException`, @zegnus  linked this because he surmised it was probably the problem:

> CopyOnWriteArrayList.subLists throw ConcurrentModificationExceptions if the containing list changes out from underneath it

rock3r:
I see thanks @Mecharyry :)

danybony (approved):

PR #349, opened 2018-02-23
Migration file ids

MechaRyRy:
## Problem
A client cannot access a specific file from `v1` without knowing the identifier we allocate during the `MigrationJob` 😬 currently this is based on the `batchTitle` and `System.nanoTime` 😱 

## Solution
For now we'll grab the `notificationextras` from the `Downloads` table in v1 and use this as the `FileId` for v2. In future we should allow clients to specify this mapping. https://github.com/novoda/download-manager/issues/350


Mecharyry:
Awaiting client app test ❌

danybony (approved):
💯

PR #351, opened 2018-02-26
Fail fast when a file size request fails

zegnus:
**Problem**

When a file request files inside the general batch total size request, the batch ends up with the incorrect total batch size

**Solution**
Is it possible that if a batch has several files, one file size request but the others are ok. In this scenario, the total batch size will be wrong.

The solution has been to check if the file size request returns zero as a total file size, and return it back to the requester that will deal correctly with that scenario.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
@@ -64,7 +64,9 @@ class DownloadBatch {
             notifyCallback(downloadBatchStatus);
         }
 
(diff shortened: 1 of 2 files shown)


MechaRyRy:
@zegnus won't `DownloadsBatchPersistence` need the same sort of fix? It also calculates the `totalSize` and updates a `batch`.

zegnus:
@Mecharyry I'm going to check

danybony (approved):
👍 for the fix.

MechaRyRy (approved):

PR #353, opened 2018-02-27
Flashing notification on 4.4.2

MechaRyRy:
## Problem
QA on a client application has picked up that notifications flash on Android 4.4.2 for the persisted downloading state 😢 

## Solution
After some fixes that failed and more research into the issue I stumbled upon [this](https://stackoverflow.com/questions/19025056/how-can-i-avoid-blinking-notification-update-while-changing-button), which suggests that using the same `NotificationCompat.Builder` for similar notifications rectifies the issue, and indeed it does.

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -35,7 +35,7 @@ android {
 dependencies {
     annotationProcessor 'android.arch.persistence.room:compiler:1.0.0'
     implementation 'com.novoda:notils:2.2.16'
-    implementation 'com.novoda:merlin:1.1.4'
+    implementation 'com.novoda:merlin:1.1.6'
     implementation 'com.facebook.stetho:stetho:1.5.0'
     implementation 'com.android.support:support-v4:27.0.2'
     implementation 'com.squareup.okhttp:okhttp:2.7.5'
--- a/library/src/main/java/com/novoda/downloadmanager/NotificationCreator.java
+++ b/library/src/main/java/com/novoda/downloadmanager/NotificationCreator.java
@@ -21,7 +21,6 @@ class NotificationCreator<T> {
     }
 
     private int notificationId;
-    private NotificationCustomizer.NotificationDisplayState notificationDisplayState;
     private NotificationCompat.Builder builder;
 
     NotificationInformation createNotification(final T notificationPayload) {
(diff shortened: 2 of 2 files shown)


library/build.gradle
```diff
@@ -35,7 +35,7 @@ android {
 dependencies {
     annotationProcessor 'android.arch.persistence.room:compiler:1.0.0'
     implementation 'com.novoda:notils:2.2.16'
-    implementation 'com.novoda:merlin:1.1.4'
+    implementation 'com.novoda:merlin:1.1.6'
```
MechaRyRy, line 38: lint failure because this has been updated.

library/src/main/java/com/novoda/downloadmanager/NotificationCreator.java
```diff
@@ -33,10 +37,22 @@ public int getId() {
 
             @Override
             public Notification getNotification() {
-                NotificationCompat.Builder builder = new NotificationCompat.Builder(applicationContext, notificationChannelProvider.channelId());
+                int newNotificationId = getId();
+                NotificationCustomizer.NotificationDisplayState newNotificationDisplayState = notificationDisplayState();
+
+                if (builder == null || hasChanged(newNotificationId, newNotificationDisplayState)) {
```
MechaRyRy: Now we create a new `Builder` whenever there isn't one present OR whenever the `id` and `displayState` has changed. If the display state and notification Id has not changed then there probably is no need to create a new builder. 

🤔 in fact, there might be no point in calling down to the `notificationCustomizer` if the `builder` has not changed between calls 🤔 What do you think? ❓

danybony: well, you still need to call the `notificationCustomizer` in any case, because id and displayState might have not changed, but the payload could, isn't it?

library/src/main/java/com/novoda/downloadmanager/NotificationCreator.java
```diff
@@ -33,10 +37,22 @@ public int getId() {
...
+                    builder = new NotificationCompat.Builder(applicationContext, notificationChannelProvider.channelId());
+                    notificationId = newNotificationId;
+                    notificationDisplayState = newNotificationDisplayState;
+                }
+
                 return notificationCustomizer.customNotificationFrom(builder, notificationPayload);
             }
 
+            private boolean hasChanged(int newNotificationId, NotificationCustomizer.NotificationDisplayState newNotificationDisplayState) {
+                return notificationId != newNotificationId && !notificationDisplayState.equals(newNotificationDisplayState);
```
danybony: shouldn't be an `||` here?

library/src/main/java/com/novoda/downloadmanager/NotificationCreator.java
MechaRyRy: Yea you are completely correct 😄 just occurred to me too after reading it again 😂

library/src/main/java/com/novoda/downloadmanager/NotificationCreator.java
MechaRyRy: 🤔 I'm pretty sure this only needs to be `state` too now that I think about it. If the id changes and the state doesn't then it would still be using the same customiser, the clients don't have access to the ids after-all.

MechaRyRy:
Awaiting verification that c27c594 hasn't regressed the fix.

MechaRyRy:
Verified, everything is fine.

danybony (approved):

ataulm (approved):

PR #354, opened 2018-03-02
Register a notification channel for download migration

MechaRyRy:
## Problem
There's a crash when a custom notification channel is specified on the `DownloadMigratorBuilder`, there is no check in the `build` for android `Oreo` and we never register a `NotificationChannel` 😬 

I believe it works on the default cases in the demo because the `DownloadManager` and the `Migrator` share the same channel name.

## Solution
Add the check for `Oreo` and register a `NotificationChannel`.

❓ Should the `Migrator` and `DownloadManager` share the same notification channel by default? They do currently, which is why this doesn't show up unless you create a custom one. ❓


rock3r (approved):
The fix LGTM, cannot help much on the philosophical question :)

danybony (approved):
👍 
I still think they should share the same channel name by default
If a client wants to provide two different ones fine, but for the user perspective, they are all downloads

PR #356, opened 2018-03-05
Add equals and hashcode to our models

MechaRyRy:
### Problem
It's hard to add logs and test scenarios when our models do not have `equals`, `hashcode` or `toString` 😂 

### Solution
Add equals, hashcode and toString.


zegnus:
👍 be aware that an equals for two batch statuses includes the downloaded bytes ;)

zegnus (approved):

PR #357, opened 2018-03-05
Progress percentage shown on completed downloads

MechaRyRy:
## Problem 
We currently use `NotifcationId` to determine when we should create a new `NotificationCompat.Builder`. Creating a builder when it is necessary rather than every-time removes a flashing persisted notification in `4.4.2`

The `NotificationId` is not enough to determine when we should create this builder, take the following: 
```
Id = 1
State = PERSISTENT
Status = DOWNLOADING
Display as downloading WITH progress

Id = 1
State = PERSISTENT
Status = COMPLETED
Display as completed WITHOUT progress
```

If we use `State` and `Id` check for the above then both will show the progress because the `Builder` is not recreated between the two scenarios 😬 

## Solution
Create an interface `NotificationCreator<T>` and provide an implementation for `DownloadBatchStatus` and `MigrationStatus`. Only recreate the `NotificationCompat.Builder` when the `Id` or the `Status` has changed. This will rectify the problem above. 

Fix the Migration notifications, it had a very bad default implementation that would show 95% rather than completed 😬 I've removed the cloned `NotifcationCustomizer` from the `MainActivity` too, would rather keep the code here concise to avoid confusion. 

### Screen Capture

Before | After
--- | ---
before | after


library/src/main/java/com/novoda/downloadmanager/NotificationCreator.java
```diff
@@ -1,55 +1,8 @@
 package com.novoda.downloadmanager;
 
-import android.app.Notification;
-import android.content.Context;
-import android.support.v4.app.NotificationCompat;
+interface NotificationCreator<T> {
 
-class NotificationCreator<T> {
+    void setNotificationChannelProvider(NotificationChannelProvider notificationChannelProvider);
```
danybony, line 5: Do we really need this to be in the interface?
I don't know, it looks weird to have a setter here...

MechaRyRy: The `interface` is internal only. If I understand the `DownloadManagerBuilder` correctly, this can't be removed. The client has the option to change the `NotificationChannel` using the `builder` pattern, after the client has already decided on the type of `customisation` they want.

danybony (approved):
The changes LGTM, just what do you think about leaving or providing another notification customiser to the demo? To show users how to customise it in the client

MechaRyRy:
I don't think it is necessary. We'll probably have a wiki when we release and clients can always check the default implementation and go off of that.

zegnus (approved):

PR #359, opened 2018-03-08
Migration Fixes

MechaRyRy:
## Problem
This PR fixes a few problems with the `MigrationJob`:

- A client application needs to be able to specify the `DownloadBatchId` from version one 😬 otherwise it is unable to locate some additional data that is required to work with a download.
- The `MigrationJob` completes too quickly and relies on mutation of a `MigrationStatus`. If the `Migration` occurs too fast, given that we are reporting back to the `main-thread` using a `Handler` we could be potentially emitting the same status multiple times. It's mutated before the handler posts to the `main-thread`, modifying an already existing reference 😬 
- The `MigrationJob` uses the `hashcode` of the entire `MigrationStatus` object when posting notifications 😬 so it will recreate notifications each and every time / may not remove persistent notifications etc because they all rely on the notification ids being the same! 
- Partial migrations were never taken into account when updating the migration percentage 😬 

## Solution

- For now we are using the `notificationExtras` from the first `file` retrieved from the `Downloads` table in v1 for the `DownloadBatchId`. We will need to provide a different mechanism for specifying the mapping before we release this library publicly. 
- It's ok to mutate the `MigrationStatus` anywhere we have control of it **BUT** before we give control of it over to clients we should create an `ImmutableCopy`.
- Use the `hashcode` of the migration job identifier as the notification id. So all notifications from a given `MigrationJob` share the same notification. 
- Ensure that the `total` and `current` completed migrations and the associated migration percentage is driven by both the partial and complete migrations. 

## Screen Capture
migration mp4

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadMigrator.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadMigrator.java
@@ -4,6 +4,6 @@ import java.io.File;
 
 public interface DownloadMigrator {
 
-    void startMigration(File databaseFile);
+    void startMigration(String jobIdentifier, File databaseFile);
 
 }
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadMigratorBuilder.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadMigratorBuilder.java
@@ -151,7 +151,7 @@ public final class DownloadMigratorBuilder {
 
         @Override
         public Notification customNotificationFrom(NotificationCompat.Builder builder, MigrationStatus migrationStatus) {
-            String title = migrationStatus.status().toRawValue();
+            String title = resources.getString(R.string.migration_notification_content_title);
             builder.setSmallIcon(notificationIcon)
                     .setContentTitle(title);
 
--- a/library/src/main/java/com/novoda/downloadmanager/InternalMigrationStatus.java
+++ b/library/src/main/java/com/novoda/downloadmanager/InternalMigrationStatus.java
@@ -2,8 +2,6 @@ package com.novoda.downloadmanager;
 
 interface InternalMigrationStatus extends MigrationStatus {
 
-    void update(int numberOfMigrationsCompleted, int totalNumberOfMigrations);
-
     void migrationComplete();
 
     void markAsExtracting();
--- a/library/src/main/java/com/novoda/downloadmanager/InternalMigrationStatus.java
+++ b/library/src/main/java/com/novoda/downloadmanager/InternalMigrationStatus.java
@@ -14,6 +14,6 @@ interface InternalMigrationStatus extends MigrationStatus {
 
     void markAsComplete();
 
-    VersionOneToVersionTwoMigrationStatus copy();
+    InternalMigrationStatus copy();
 
 }
--- a/library/src/main/java/com/novoda/downloadmanager/InternalMigrationStatus.java
+++ b/library/src/main/java/com/novoda/downloadmanager/InternalMigrationStatus.java
@@ -2,7 +2,9 @@ package com.novoda.downloadmanager;
 
 interface InternalMigrationStatus extends MigrationStatus {
 
-    void update(int currentBatch, int numberOfBatches);
+    void update(int numberOfMigrationsCompleted, int totalNumberOfMigrations);
+
+    void migrationComplete();
 
     void markAsExtracting();
 
@@ -11,5 +13,7 @@ interface InternalMigrationStatus extends MigrationStatus {
     void markAsDeleting();
 
     void markAsComplete();
+
+    VersionOneToVersionTwoMigrationStatus copy();
 
 }
--- a/library/src/main/java/com/novoda/downloadmanager/InternalMigrationStatus.java
+++ b/library/src/main/java/com/novoda/downloadmanager/InternalMigrationStatus.java
@@ -2,7 +2,7 @@ package com.novoda.downloadmanager;
 
 interface InternalMigrationStatus extends MigrationStatus {
 
-    void migrationComplete();
+    void onSingleBatchMigrated();
 
     void markAsExtracting();
 
--- a/library/src/main/java/com/novoda/downloadmanager/Migration.java
+++ b/library/src/main/java/com/novoda/downloadmanager/Migration.java
@@ -5,14 +5,21 @@ import java.util.List;
 
 class Migration {
 
+    enum Type {
+        COMPLETE,
+        PARTIAL
+    }
+
     private final Batch batch;
     private final List<FileMetadata> fileMetadata;
     private final long downloadedDateTimeInMillis;
+    private final Type type;
 
-    Migration(Batch batch, List<FileMetadata> fileMetadata, long downloadedDateTimeInMillis) {
+    Migration(Batch batch, List<FileMetadata> fileMetadata, long downloadedDateTimeInMillis, Type type) {
         this.batch = batch;
         this.fileMetadata = Collections.unmodifiableList(fileMetadata);
         this.downloadedDateTimeInMillis = downloadedDateTimeInMillis;
+        this.type = type;
     }
 
     Batch batch() {
@@ -27,13 +34,8 @@ class Migration {
         return downloadedDateTimeInMillis;
     }
 
-    boolean hasDownloadedBatch() {
-        for (FileMetadata fileMetadatum : fileMetadata) {
-            if (fileMetadatum.fileSize().currentSize() != fileMetadatum.fileSize().totalSize()) {
-                return false;
-            }
-        }
-        return true;
+    Type type() {
+        return type;
     }
 
     @Override
@@ -50,17 +52,21 @@ class Migration {
         if (downloadedDateTimeInMillis != migration.downloadedDateTimeInMillis) {
             return false;
         }
-        if (!batch.equals(migration.batch)) {
+        if (batch != null ? !batch.equals(migration.batch) : migration.batch != null) {
             return false;
         }
-        return fileMetadata.equals(migration.fileMetadata);
+        if (fileMetadata != null ? !fileMetadata.equals(migration.fileMetadata) : migration.fileMetadata != null) {
+            return false;
+        }
+        return type == migration.type;
     }
 
     @Override
     public int hashCode() {
-        int result = batch.hashCode();
-        result = 31 * result + fileMetadata.hashCode();
+        int result = batch != null ? batch.hashCode() : 0;
+        result = 31 * result + (fileMetadata != null ? fileMetadata.hashCode() : 0);
         result = 31 * result + (int) (downloadedDateTimeInMillis ^ (downloadedDateTimeInMillis >>> 32));
+        result = 31 * result + (type != null ? type.hashCode() : 0);
         return result;
     }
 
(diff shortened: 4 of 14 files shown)


library/src/main/java/com/novoda/downloadmanager/DownloadMigrator.java
```diff
@@ -4,6 +4,6 @@
 
 public interface DownloadMigrator {
 
-    void startMigration(File databaseFile);
+    void startMigration(String jobIdentifier, File databaseFile);
```
MechaRyRy, line 7: Supplying a job identifier means that notifications for a given job share the same id, allowing the `NotificationDispatcher` to switch between states correctly.

library/src/main/java/com/novoda/downloadmanager/DownloadMigratorBuilder.java
```diff
@@ -151,7 +151,7 @@ public NotificationDisplayState notificationDisplayState(MigrationStatus migrati
 
         @Override
         public Notification customNotificationFrom(NotificationCompat.Builder builder, MigrationStatus migrationStatus) {
-            String title = migrationStatus.status().toRawValue();
+            String title = resources.getString(R.string.migration_notification_content_title);
```
MechaRyRy, line 154: Notification title update to display: Migrating Download Manager V1

library/src/main/java/com/novoda/downloadmanager/InternalMigrationStatus.java
```diff
@@ -2,7 +2,7 @@
 
 interface InternalMigrationStatus extends MigrationStatus {
 
-    void update(int currentBatch, int numberOfBatches);
+    void migrationComplete();
```
MechaRyRy: `increments` the internal count of the `numberOfCompletedMigrations`

library/src/main/java/com/novoda/downloadmanager/Migration.java
```diff
@@ -5,14 +5,21 @@
 
 class Migration {
 
+    enum Type {
```
MechaRyRy, line 8: This is used to drive whether we should retrieve the file size for a given download. For partials we don't care, it should be zero so that the `DownloadManager` can request from the network again.

library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
```diff
@@ -56,36 +90,13 @@
...
-                        originalFileLocation,
-                        fileSize,
-                        originalNetworkAddress
-                );
-                fileMetadataList.add(fileMetadata);
-            }
-            return fileMetadataList;
-        } finally {
-            downloadsCursor.close();
+    private DownloadBatchId createDownloadBatchIdFrom(String originalFileId, String batchId) {
```
MechaRyRy, line 93: Client application needs the `DownloadBatchId` to be derived from the first files `notificationExtras` 😬 we will need to provide a mechanism for clients to specify this mapping before a public release.

MechaRyRy: The same applies for `FileIds` too.

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
```diff
@@ -41,35 +49,51 @@ public void run() {
...
 
-    private void migratePartialDownloads(SqlDatabaseWrapper database,
-                                         PartialDownloadMigrationExtractor partialDownloadMigrationExtractor,
+    private void migratePartialDownloads(InternalMigrationStatus migrationStatus,
+                                         SqlDatabaseWrapper database,
+                                         List<Migration> partialMigrations,
                                          DownloadsPersistence downloadsPersistence,
                                          String basePath) {
-        List<Migration> partialMigrations = partialDownloadMigrationExtractor.extractMigrations();
+        Log.d(TAG, "Partial migrations starting " + System.nanoTime());
```
MechaRyRy: Removed a lot of the logs from this class so that we notify before and after a method executes.

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
```diff
@@ -157,43 +187,25 @@ private boolean hasValidFileLocation(Migration.FileMetadata metadata) {
...
             downloadsPersistence.endTransaction();
             database.setTransactionSuccessful();
             database.endTransaction();
+            migrationStatus.migrationComplete();
+            onUpdate(migrationStatus);
         }
-
-        Log.d(TAG, "all data migrations are COMMITTED, about to delete the old database, time is " + System.nanoTime());
-
-        migrationStatus.markAsDeleting();
```
MechaRyRy: Not relevant to `complete migrations` moved to it's own method that executes after **all** migrations are completed.

library/src/main/java/com/novoda/downloadmanager/MigrationStatusNotificationCreator.java
```diff
@@ -29,7 +29,7 @@ public NotificationInformation createNotification(final MigrationStatus migratio
         return new NotificationInformation() {
             @Override
             public int getId() {
-                return migrationStatus.hashCode();
+                return migrationStatus.migrationId().hashCode();
```
MechaRyRy, line 32: Now the `NotificationDispatcher` can correctly switch between the different types of notifications 😄 if it was left how it was then it would keep the persisted notification **and** display the completed notification.

library/src/main/java/com/novoda/downloadmanager/PartialDownloadMigrationExtractor.java
```diff
@@ -14,12 +14,10 @@
     private static final int TITLE_COLUMN = 1;
     private static final int MODIFIED_TIMESTAMP_COLUMN = 2;
 
-    private static final String DOWNLOADS_QUERY = "SELECT uri, _data, current_bytes, total_bytes, notificationextras FROM Downloads WHERE batch_id = ?";
```
MechaRyRy, line 17: `current` and `total` bytes is not required for partial migrations. We want the `download-manager` to re-download these and get the size from the network call it will do!

library/src/main/java/com/novoda/downloadmanager/PartialDownloadMigrationExtractor.java
```diff
@@ -38,32 +36,40 @@
...
 
             Batch batch = newBatchBuilder.build();
-            migrations.add(new Migration(batch, fileMetadataList, downloadedDateTimeInMillis));
+            migrations.add(new Migration(batch, fileMetadataList, downloadedDateTimeInMillis, Migration.Type.PARTIAL));
         }
         batchesCursor.close();
         return migrations;
     }
+
+    private DownloadBatchId createDownloadBatchIdFrom(String originalFileId, String batchId) {
```
MechaRyRy, line 67: Same as for the `Complete Migrations`.

library/src/main/java/com/novoda/downloadmanager/InternalMigrationStatus.java
danybony: This is pretty confusing, compared with `markAsComplete()`...
Maybe some name referencing that this is a single migration, compared to the whole migration progress?

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
```diff
@@ -31,8 +33,14 @@ void addCallback(MigrationCallback callback) {
     }
 
     public void run() {
-        InternalMigrationStatus migrationStatus = new VersionOneToVersionTwoMigrationStatus(MigrationStatus.Status.DB_NOT_PRESENT);
         if (!databasePath.exists()) {
+            InternalMigrationStatus migrationStatus = new VersionOneToVersionTwoMigrationStatus(
+                    jobIdentifier,
+                    MigrationStatus.Status.DB_NOT_PRESENT,
+                    0,
```
zegnus: can we extract those numbers as constants with names?

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
```diff
@@ -41,35 +49,51 @@ public void run() {
         SqlDatabaseWrapper database = new SqlDatabaseWrapper(sqLiteDatabase);
 
         FilePersistence filePersistence = FilePersistenceCreator.newInternalFilePersistenceCreator(context).create();
+        String basePath = filePersistence.basePath().path();
```
zegnus, line 52: can you move this near line 77?

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
```diff
@@ -41,35 +49,51 @@ public void run() {
...
+        List<Migration> completeMigrations = migrationExtractor.extractMigrations();
         DownloadsPersistence downloadsPersistence = RoomDownloadsPersistence.newInstance(context);
         LocalFilesDirectory localFilesDirectory = new AndroidLocalFilesDirectory(context);
         UnlinkedDataRemover unlinkedDataRemover = new UnlinkedDataRemover(downloadsPersistence, localFilesDirectory);
 
+        int totalNumberOfMigrations = partialMigrations.size() + completeMigrations.size();
+        InternalMigrationStatus migrationStatus = new VersionOneToVersionTwoMigrationStatus(
+                jobIdentifier,
+                MigrationStatus.Status.DB_NOT_PRESENT,
+                0,
```
zegnus: can you extract this as a constant?

zegnus (commented):
Can we remove the Log.d that includes the nanoseconds?

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
MechaRyRy: I didn't do this because I wanted to keep the creation of dependencies together, away from where the execution happens.

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
MechaRyRy: 64ad585

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
MechaRyRy: 64ad585

MechaRyRy:
We can remove the `Log.d`, would rather have a ticket to add a `Logger` where we can attach a `Handle` if we need to.

library/src/main/java/com/novoda/downloadmanager/InternalMigrationStatus.java
MechaRyRy: I'm open to ideas 😂

MechaRyRy: cecb7bb

danybony: Waaay better! 💯

danybony (approved):

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
zegnus: I wasn't sure where that was used... this is why I asked

zegnus:
@Mecharyry yeah let's add the logger, and remove it for now

zegnus (approved):

PR #361, opened 2018-03-09
Add batchId when requesting for a particular file

zegnus:
It is possible that the same file id is used in different batches, in order to ensure that when we request a particular file, is the one that belongs to the correct batch we must pass the batch id as part of the request.

Paired with @danybony


library/build.gradle
```diff
@@ -51,7 +51,7 @@ publish {
     userOrg = 'novoda'
     groupId = 'com.novoda'
     artifactId = 'download-manager'
-    publishVersion = 'SNAPSHOT-2.0.25'
+    publishVersion = 'SNAPSHOT-2.0.26'
```
danybony: 👍

MechaRyRy: ❌

danybony (approved):

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerCommands.java
```diff
@@ -29,9 +29,9 @@
 
     @Nullable
     @WorkerThread
-    DownloadFileStatus getDownloadStatusWithMatching(DownloadFileId downloadFileId);
+    DownloadFileStatus getDownloadStatusWithMatching(DownloadBatchId downloadBatchId, DownloadFileId downloadFileId);
```
MechaRyRy: I just noticed the method says `DownloadStatus`, probably should be `FileDownloadStatus`.

MechaRyRy: Would rather change it now because we'll need to update in client app anyway!

MechaRyRy (approved):

PR #362, opened 2018-03-12
Reschedule job improvements

MechaRyRy:
## Problem
Network recovery jobs are taking a life time to execute and they ignore whether a network connection actually exists 😬 

According to the docs: 

```
 * Note that if the deadline is met and the requirements aren't enforced, then your job
 * will run and ignore this requirement.
```

```
 * <b>NOTE:</b> It's not recommended to have such big execution windows. The {@code AlarmManager} used
 * as fallback API doesn't allow setting a start date. Although being inexact, the execution time is
 * the arithmetic average of {@code startInMs} and {@code endInMs}. The result could be that your job never
 * runs on pre Android 5.0 devices, if one argument is too large.
```

## Solution
Ensure that we `enforceRequirements` so that the job does not run unless we have met the network requirements. 

Change the window end time to 5mins so that we don't take half a day for a job to start on pre-5 devices 😬 

Update the `Job.TAG` to something more meaningful than `test` 😄

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/LiteJobCreator.java
+++ b/library/src/main/java/com/novoda/downloadmanager/LiteJobCreator.java
@@ -5,7 +5,7 @@ import com.evernote.android.job.JobCreator;
 
 class LiteJobCreator implements JobCreator {
 
-    static final String TAG = "test";
+    static final String TAG = "download-manager-reschedule";
 
     private final DownloadManager downloadManager;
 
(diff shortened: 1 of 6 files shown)


library/src/main/java/com/novoda/downloadmanager/LiteJobCreator.java
```diff
@@ -5,7 +5,7 @@
 
 class LiteJobCreator implements JobCreator {
 
-    static final String TAG = "test";
+    static final String TAG = "download-manager-reschedule";
```
danybony, line 8: 💯

danybony (approved):

rock3r (approved):
EZ

PR #363, opened 2018-03-13
Fix deletion crash

zegnus:
**Problem**
1. Queue 10 batches
2. Pause one
3. Delete 7 (including the downloading and the paused one)
4. Crash

**Solution**
The problem was detected when the deletion is not perfectly timed with the next batch to be processed. The next batch will start requesting the fileSize and storing it locally, but at the same time, the batch database is going to be deleted. At the subsequent fileSize write to the database, the parent batch does not exist and the constraint in the database fails (as expected), as you cannot insert a file without an existing parent batch.

The solution is to not delete immediately the batch on delete request, but let the batch process as normal and delete at the end.

There is also the `pause` case. In this scenario the paused batch finishes and no longer exists in the executor. For this scenario, an immediate delete must be done on request.

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -51,7 +51,7 @@ publish {
     userOrg = 'novoda'
     groupId = 'com.novoda'
     artifactId = 'download-manager'
-    publishVersion = 'SNAPSHOT-2.0.27'
+    publishVersion = 'SNAPSHOT-2.0.28'
     description = 'Download manager based on AOSP DM but allowing downloading to internal private storage.'
     website = 'https://github.com/novoda/download-manager'
 }
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
@@ -62,7 +62,6 @@ class DownloadBatch {
 
         if (connectionNotAllowedForDownload(status)) {
             processNetworkError();
-            deleteBatchIfNeeded();
             return;
         }
 
@@ -77,7 +76,6 @@ class DownloadBatch {
 
         if (totalBatchSizeBytes <= ZERO_BYTES) {
             processNetworkError();
-            deleteBatchIfNeeded();
             return;
         }
 
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
@@ -50,12 +50,19 @@ class DownloadBatch {
 
     void download() {
         DownloadBatchStatus.Status status = downloadBatchStatus.status();
-        if (status == PAUSED || status == DELETED) {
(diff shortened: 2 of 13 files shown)


library/build.gradle
```diff
@@ -51,7 +51,7 @@ publish {
     userOrg = 'novoda'
     groupId = 'com.novoda'
     artifactId = 'download-manager'
-    publishVersion = 'SNAPSHOT-2.0.27'
+    publishVersion = 'SNAPSHOT-2.0.28'
```
MechaRyRy, line 54: 👍

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -207,7 +222,10 @@ void resume() {
     }
 
     void delete() {
-        downloadsBatchPersistence.deleteAsync(downloadBatchStatus.getDownloadBatchId());
+        if (downloadBatchStatus.status() == PAUSED) {
+            downloadsBatchPersistence.deleteAsync(downloadBatchStatus.getDownloadBatchId());
```
danybony, line 225: Just to check, do we still need this to be async?

zegnus: yes in this case, because this `delete()` is an immediate request (main thread)

library/src/main/java/com/novoda/downloadmanager/DownloadsBatchPersistence.java
```diff
@@ -44,15 +44,14 @@ void persistAsync(DownloadBatchTitle downloadBatchTitle,
...
-                for (DownloadFile downloadFile : downloadFiles) {
-                    downloadFile.persistSync();
-                }
-
                 downloadsPersistence.transactionSuccess();
             } finally {
                 downloadsPersistence.endTransaction();
             }
+
+            for (DownloadFile downloadFile : downloadFiles) {
```
MechaRyRy, line 52: Why was this moved?

zegnus: now the `downloadFile.persistSync` has his own transaction operations

danybony (approved):

library/src/main/java/com/novoda/downloadmanager/DownloadsBatchPersistence.java
MechaRyRy: I was wondering if we should move this transaction logic to the `DownloadsPersistence` rather than having it everywhere else.

MechaRyRy: Downside would be that any new `DownloadPersistence` would have to pay attention to implement the transactions too 🤔

MechaRyRy (approved):

PR #364, opened 2018-03-13
Apply same fix to external persistence

MechaRyRy:
## Problem 
We applied a fix to the internal persistence in #363 but we didn't apply it to the external persistence 😬 

## Solution 
Apply the same fix to the external persistence.


danybony (approved):
👍

zegnus:
💯

zegnus (approved):

PR #365, opened 2018-03-13
Avoid duplicated fileId

danybony:
### Problem
Before it was possible for a client to pass the same FileId multiple times when creating a Batch.
This was causing the DownloadBatch download logic to fail when checking the total batch size.
The same was happening when adding multiple files to a Batch without FileId but with the same network address.

### Solution
We decided to fail fast during the `Batch` creation, checking if the builder is configured with duplicated file Ids.
Added tests about that.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerDownloader.java
+++ b/library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerDownloader.java
@@ -2,6 +2,8 @@ package com.novoda.downloadmanager;
 
 import android.os.Handler;
 
+import com.novoda.notils.logger.simple.Log;
+
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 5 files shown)


MechaRyRy (approved):
Wondering how we should handle this for duplicate batch identifiers 😬

danybony:
Interesting point... maybe we should throw at `LiteDownloadManagerCommands.download(batch)` too when trying to add another batch with the same id of a present one?

MechaRyRy:
Yea I think we should, atm it will just swallow the second batch id, what do you think @zegnus ?

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerDownloader.java
```diff
@@ -55,6 +57,7 @@
     public void download(Batch batch, Map<DownloadBatchId, DownloadBatch> downloadBatchMap) {
         DownloadBatch runningDownloadBatch = downloadBatchMap.get(batch.downloadBatchId());
         if (runningDownloadBatch != null) {
+            Log.w("Ignoring duplicate batch with id " + batch.downloadBatchId());
```
MechaRyRy: This might be triggered by `submitAllStoredDownloads` right? Would it be better to say:

"Already running download for DownloadBatchId: %s , ensure you are not duplicating identifiers."

danybony: No, that method can only be called by the client using the interface

danybony: What you are referring to is the next one, with different signature (taking a `DownloadBatch`).
But we'll never reach that point: batches are filtered first

zegnus (approved):

PR #366, opened 2018-03-13
Queueing and deleting the same batches crashes the application

zegnus:
**Problem**
1. Start the demo app
2. Click on download add
3. Click very fast delete and then download all again
4. Crash

**Solution**
The problem was that the deletion of a queued batch occurs in the batch thread and only happens when the batch finishes. But when we enqueue a new batch we immediately persist it.

When the same batch is queued -> deleted -> queued again very fast, the delayed deletion occurs after the second enqueue happens, then the second enqueue wants to use the database but the batch was previously deleted.

The solution is to ensure that the batch is persisted when the batch starts to download.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
@@ -1,6 +1,7 @@
 package com.novoda.downloadmanager;
 
 import android.support.annotation.Nullable;
+import android.support.annotation.WorkerThread;
 
 import java.util.List;
 import java.util.Map;
@@ -261,6 +262,7 @@ class DownloadBatch {
         );
     }
 
+    @WorkerThread
     void persist() {
         downloadsBatchPersistence.persist(
                 downloadBatchStatus.getDownloadBatchTitle(),
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
@@ -2,6 +2,8 @@ package com.novoda.downloadmanager;
 
 import android.support.annotation.Nullable;
 
+import com.novoda.notils.logger.simple.Log;
+
 import java.util.List;
 import java.util.Map;
 
@@ -250,8 +252,19 @@ class DownloadBatch {
         return null;
     }
 
-    void persist() {
+    void persistAsync() {
         downloadsBatchPersistence.persistAsync(
+                downloadBatchStatus.getDownloadBatchTitle(),
+                downloadBatchStatus.getDownloadBatchId(),
+                downloadBatchStatus.status(),
+                downloadFiles,
+                downloadBatchStatus.downloadedDateTimeInMillis(),
+                downloadBatchStatus.notificationSeen()
+        );
+    }
+
+    void persist() {
+        downloadsBatchPersistence.persist(
                 downloadBatchStatus.getDownloadBatchTitle(),
                 downloadBatchStatus.getDownloadBatchId(),
                 downloadBatchStatus.status(),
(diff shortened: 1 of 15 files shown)


library/src/main/java/com/novoda/downloadmanager/LiteDownloadService.java
```diff
@@ -51,13 +51,14 @@ public IBinder onBind(Intent intent) {
...
     @Override
-    public void download(final DownloadBatch downloadBatch, final DownloadBatchStatusCallback callback) {
+    public void download(DownloadBatch downloadBatch, DownloadBatchStatusCallback callback) {
         callback.onUpdate(downloadBatch.status());
 
         downloadBatch.setCallback(callback);
 
         executor.execute(() -> {
             acquireCpuWakeLock();
+            downloadBatch.persist();
```
zegnus, line 61: this is the fix, we guarantee that when the download starts, the necessary information in the database exists

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -260,4 +260,15 @@ void persist() {
                 downloadBatchStatus.notificationSeen()
         );
     }
+
+    void persist() {
```
MechaRyRy: All this switching between `sync` and `async` is hurting my head. Would it be better to have an executor that lives at the db level so that we can queue these things? Rather than switching on the higher levels?

MechaRyRy: Wait, the executor does live on the persistence 😕

library/src/main/java/com/novoda/downloadmanager/LiteDownloadBatchStatus.java
```diff
@@ -109,25 +109,29 @@ public void markAsDeleted() {
...
-        updateStatus(status, persistence);
+        updateStatusAsync(status, persistence);
     }
 
-    private void updateStatus(Status status, DownloadsBatchStatusPersistence persistence) {
+    private void updateStatusAsync(Status status, DownloadsBatchStatusPersistence persistence) {
         persistence.updateStatusAsync(downloadBatchId, status);
     }
 
+    private void updateStatus(Status status, DownloadsBatchStatusPersistence persistence) {
```
danybony, line 136: So in the end this one is only used for the DOWNLOADING state

danybony (approved):
👍 for the fix

zegnus:
There is still an issue, closing to review

zegnus:
This is good enough for a normal usage of the library, but we continue to work in order to find a solution that survives for spamming the download manager with the following sequence:
1. start download batch 1 with 2 files
2 start download batch 2 with 2 files
3. delete all
4. start 1 (with the same batch ids and files)

repeat clicking the buttons as fast as possible

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -260,4 +261,16 @@ void persist() {
                 downloadBatchStatus.notificationSeen()
         );
     }
+
+    @WorkerThread
+    void persist() {
```
MechaRyRy, line 266: I thought this one wasn't necessary? Only called from one place and that should be async too?

library/src/main/java/com/novoda/downloadmanager/LiteDownloadBatchStatus.java
```diff
@@ -81,6 +82,7 @@ public long downloadedDateTimeInMillis() {
         return downloadedDateTimeInMillis;
     }
 
+    @WorkerThread
```
MechaRyRy, line 85: I thought `DOWNLOADING` came from the `download-manager` and therefore was already on an executor, not on the main thread.

PR #369, opened 2018-03-15
Fix pausing affects other batches

zegnus:
**Problem**
When we click on the `pause` button in the demo, we can see how the other batch is affected

**Solution**
The problem is that the class that handles pausing the actual file is shared across batches. Providing new instances to all the files solves properly this issue

| Before | After |
| ---- | ---- | 
| before mp4 | after mp4


MechaRyRy (approved):
💯

library/src/main/java/com/novoda/downloadmanager/FileDownloaderType.java
```diff
@@ -0,0 +1,28 @@
+package com.novoda.downloadmanager;
+
+import java.security.InvalidParameterException;
+
+public enum FileDownloaderType {
```
danybony: This can be package protected

library/src/main/java/com/novoda/downloadmanager/FileDownloaderCreator.java
```diff
@@ -0,0 +1,74 @@
...
+    FileDownloaderCreator(FileDownloaderType type, @Nullable Class<? extends FileDownloader> customClass) {
+        this.type = type;
+        this.customClass = customClass;
+    }
+
+    FileDownloader create() {
+        return create(type);
+    }
+
+    private FileDownloader create(FileDownloaderType type) {
```
danybony: can you inline this?

library/src/main/java/com/novoda/downloadmanager/FileDownloaderCreator.java
```diff
@@ -0,0 +1,74 @@
...
+            case NETWORK:
+                HttpClient httpClient = HttpClientFactory.getInstance();
+                NetworkRequestCreator requestCreator = new NetworkRequestCreator();
+                fileDownloader = new NetworkFileDownloader(httpClient, requestCreator);
+                break;
+            case CUSTOM:
+                fileDownloader = createCustomFileDownloader();
+                break;
+            default:
+                throw new IllegalStateException("Persistence of type " + type + " is not supported");
```
danybony: This is not Persistence, is FileDownloader

library/src/main/java/com/novoda/downloadmanager/FileDownloaderCreator.java
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        return fileDownloader;
+    }
+
+    private FileDownloader createCustomFileDownloader() {
+        if (customClass == null) {
+            throw new CustomFilePersistenceException("CustomFilePersistence class cannot be accessed, is it public?");
```
danybony: `CustomFileDownloader`

library/src/main/java/com/novoda/downloadmanager/FileDownloaderCreator.java
```diff
@@ -0,0 +1,74 @@
...
+        } catch (IllegalAccessException e) {
+            throw new CustomFilePersistenceException(customClass, "Class cannot be accessed, is it public?", e);
+        } catch (ClassNotFoundException e) {
+            throw new CustomFilePersistenceException(customClass, "Class does not exist", e);
+        } catch (InstantiationException e) {
+            throw new CustomFilePersistenceException(customClass, "Class cannot be instantiated", e);
+        }
+    }
+
+    private static class CustomFilePersistenceException extends RuntimeException {
```
danybony: `CustomFileDownloaderException`

library/src/main/java/com/novoda/downloadmanager/FileDownloaderCreator.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    private FileDownloader createCustomFileDownloader() {
+        if (customClass == null) {
+            throw new CustomFilePersistenceException("CustomFilePersistence class cannot be accessed, is it public?");
+        }
+
+        try {
+            ClassLoader systemClassLoader = getClass().getClassLoader();
+            Class<?> customFilePersistenceClass = systemClassLoader.loadClass(customClass.getCanonicalName());
```
danybony: customFileDownloaderClass

library/src/main/java/com/novoda/downloadmanager/FileDownloaderType.java
```diff
@@ -0,0 +1,28 @@
...
+    CUSTOM("custom");
+
+    private final String rawValue;
+
+    FileDownloaderType(String rawValue) {
+        this.rawValue = rawValue;
+    }
+
+    static FileDownloaderType from(String rawValue) {
+        for (FileDownloaderType filePersistenceType : values()) {
```
danybony: fileDownloaderType

danybony (changes requested):
Good fix, some comments about visibility and naming

library/src/main/java/com/novoda/downloadmanager/FileDownloaderType.java
```diff
@@ -0,0 +1,24 @@
+package com.novoda.downloadmanager;
+
+import java.security.InvalidParameterException;
+
+enum FileDownloaderType {
+    NETWORK("network"),
+    CUSTOM("custom");
+
+    private final String rawValue;
```
danybony: Is this even used at this point?
the `from()` method is introduced in this PR but never called... I'd go removing the rawValue entirely if not used

danybony (approved):

PR #370, opened 2018-03-16
Fix crash with queue-delete-pause actions while in fast spamming

zegnus:
**Problem**

In the demo application, if you spam the buttons `start downloading` + `delete all` + `pause` or any combination between them, the library will eventually crash.

**Solution**

I have enabled logs through the `DownloadManagerBuilder` with `.withLogs()` so now all the logs can be toggled from any client.

I have added a new state `DELETING` so that the batch being deleted its only removed from the main downloading map when the status finishes with `DELETED`.

I have refactored the main `DownloadBatch.download()` method with more explicit names and making its method static so that it has a lower memory footprint.

I have added checks for when a download should not continue.

I have added status checks after long performing operations such as `getTotalSize()`. This is due to the `pause` and `delete` methods that can be called at any point from the main thread and are immediate actions.

I have created a `copy` method for the `DownloadBatchStatus` so that it can be passed safely to the callbacks and no further async modification of the status will be affected.

In the `DownloadBatchStatusNotificationDispatcher` I have added a `set` in order to prevent the same batch to be stored in the db for the same values.

For operations against the database, I've made sure that all the operations that can be done synchronously are properly tagged and created. This avoids issues by which a submission of an operation is outdates respected from where it has been issued. Following this, I have protected asynchronous operations that can crash with a `try-catch` block.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
@@ -122,6 +122,11 @@ class DownloadBatch {
                                                                  DownloadsBatchPersistence downloadsBatchPersistence,
                                                                  DownloadBatchStatusCallback callback,
                                                                  long totalBatchSizeBytes) {
+        if (downloadBatchStatus.status() == PAUSED) {
+            notifyCallback(callback, downloadBatchStatus);
+            return true;
+        }
+
         if (downloadBatchStatus.status() == DELETING) {
             deleteBatchIfNeeded(downloadBatchStatus, downloadsBatchPersistence, callback);
             notifyCallback(callback, downloadBatchStatus);
@@ -150,8 +155,8 @@ class DownloadBatch {
                                             DownloadsBatchPersistence downloadsBatchPersistence,
                                             DownloadBatchStatusCallback callback) {
         if (downloadBatchStatus.status() == DELETING) {
+            Log.v("deleteBatchIfNeeded: " + downloadBatchStatus.getDownloadBatchId().rawId());
             downloadsBatchPersistence.deleteSync(downloadBatchStatus);
-            Log.v("deleteBatchIfNeeded markAsDeleted: " + downloadBatchStatus.getDownloadBatchId().rawId());
             downloadBatchStatus.markAsDeleted();
             notifyCallback(callback, downloadBatchStatus);
         }
@@ -183,7 +188,7 @@ class DownloadBatch {
             notifyCallback(callback, downloadBatchStatus);
             return true;
         } else {
-            return status == ERROR || status == DELETING || status == PAUSED || status == WAITING_FOR_NETWORK;
+            return status == ERROR || status == DELETING || status == DELETED || status == PAUSED || status == WAITING_FOR_NETWORK;
         }
     }
 
@@ -241,7 +246,7 @@ class DownloadBatch {
         long totalBatchSize = 0;
         for (DownloadFile downloadFile : downloadFiles) {
             DownloadBatchStatus.Status status = downloadBatchStatus.status();
-            if (status == DELETING || status == DELETED) {
+            if (status == DELETING || status == DELETED || status == PAUSED) {
                 return 0;
             }
 
@@ -258,6 +263,7 @@ class DownloadBatch {
     }
 
     void pause() {
+        Log.v("pause batch " + downloadBatchStatus.getDownloadBatchId().rawId() + ", status: " + downloadBatchStatus.status());
         DownloadBatchStatus.Status status = downloadBatchStatus.status();
         if (status == PAUSED || status == DOWNLOADED) {
             return;
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
@@ -1,6 +1,7 @@
 package com.novoda.downloadmanager;
 
 import android.support.annotation.Nullable;
+import android.support.annotation.WorkerThread;
 
 import java.util.List;
 import java.util.Map;
@@ -261,6 +262,7 @@ class DownloadBatch {
         );
     }
 
+    @WorkerThread
     void persist() {
         downloadsBatchPersistence.persist(
                 downloadBatchStatus.getDownloadBatchTitle(),
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
@@ -55,15 +55,18 @@ class DownloadBatch {
 
         if (status == DELETED) {
             deleteBatchIfNeeded();
+            notifyCallback(downloadBatchStatus);
             return;
         }
 
         if (status == PAUSED) {
+            notifyCallback(downloadBatchStatus);
             return;
         }
 
         if (connectionNotAllowedForDownload(status)) {
             processNetworkError();
+            notifyCallback(downloadBatchStatus);
             return;
         }
 
@@ -79,6 +82,7 @@ class DownloadBatch {
         if (totalBatchSizeBytes <= ZERO_BYTES) {
             deleteBatchIfNeeded();
             processNetworkError();
+            notifyCallback(downloadBatchStatus);
             return;
         }
 
@@ -98,8 +102,9 @@ class DownloadBatch {
             processNetworkError();
         }
 
-        callbackThrottle.stopUpdates();
+        notifyCallback(downloadBatchStatus);
         deleteBatchIfNeeded();
+        callbackThrottle.stopUpdates();
     }
 
     private void deleteBatchIfNeeded() {
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
@@ -2,6 +2,8 @@ package com.novoda.downloadmanager;
 
 import android.support.annotation.Nullable;
 
+import com.novoda.notils.logger.simple.Log;
+
 import java.util.List;
 import java.util.Map;
 
@@ -250,8 +252,19 @@ class DownloadBatch {
         return null;
     }
 
-    void persist() {
+    void persistAsync() {
         downloadsBatchPersistence.persistAsync(
+                downloadBatchStatus.getDownloadBatchTitle(),
+                downloadBatchStatus.getDownloadBatchId(),
+                downloadBatchStatus.status(),
+                downloadFiles,
+                downloadBatchStatus.downloadedDateTimeInMillis(),
+                downloadBatchStatus.notificationSeen()
+        );
+    }
+
+    void persist() {
+        downloadsBatchPersistence.persist(
                 downloadBatchStatus.getDownloadBatchTitle(),
                 downloadBatchStatus.getDownloadBatchId(),
                 downloadBatchStatus.status(),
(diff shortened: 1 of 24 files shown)


library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -49,73 +53,144 @@ void setCallback(DownloadBatchStatusCallback callback) {
...
+            return;
+        }
         downloadBatchStatus.markAsWaitingForNetwork(downloadsBatchPersistence);
-        notifyCallback(downloadBatchStatus);
+        notifyCallback(callback, downloadBatchStatus);
         DownloadsNetworkRecoveryCreator.getInstance().scheduleRecovery();
     }
 
-    private boolean connectionNotAllowedForDownload(DownloadBatchStatus.Status status) {
+    private static boolean connectionNotAllowedForDownload(ConnectionChecker connectionChecker, DownloadBatchStatus.Status status) {
```
danybony: ❌ What is the amount of memory gained having these methods `static`?
I don't like to have these static unless there is a clear gain

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -49,73 +53,144 @@ void setCallback(DownloadBatchStatusCallback callback) {
     }
 
     void download() {
```
danybony, line 55: Please review the methods order in this class, following the invocation order, because now there are some on locations where you wouldn't expect them

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -152,28 +227,30 @@ private boolean networkError() {
...
+            callback.onUpdate(downloadBatchStatus.copy());
         }
     }
 
-    private long getTotalSize(List<DownloadFile> downloadFiles) {
+    private static long getTotalSize(List<DownloadFile> downloadFiles, InternalDownloadBatchStatus downloadBatchStatus) {
         long totalBatchSize = 0;
         for (DownloadFile downloadFile : downloadFiles) {
+            DownloadBatchStatus.Status status = downloadBatchStatus.status();
+            if (status == DELETING || status == DELETED || status == PAUSED) {
```
danybony: Why is this returning 0 if paused?
Are we still displaying the right value in the client app for example if we pause the download, kill the app and reopen it?

library/src/main/java/com/novoda/downloadmanager/DownloadBatchFactory.java
```diff
@@ -65,12 +69,17 @@ static DownloadBatch newInstance(Batch batch,
...
+
         InternalDownloadBatchStatus liteDownloadBatchStatus = new LiteDownloadBatchStatus(
                 downloadBatchId,
                 downloadBatchTitle,
                 downloadedDateTimeInMillis,
-                DownloadBatchStatus.Status.QUEUED,
-                NOTIFICATION_NOT_SEEN
+                BYTES_DOWNLOADED,
+                TOTAL_BATCH_SIZE_BYTES,
+                DownloadBatchStatus.Status.UNKNOWN,
```
danybony, line 80: Why can't we keep it to queued when created?

library/src/main/java/com/novoda/downloadmanager/DownloadsBatchPersistence.java
```diff
@@ -1,13 +1,20 @@
...
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Executor;
 
 class DownloadsBatchPersistence implements DownloadsBatchStatusPersistence, DownloadsNotificationSeenPersistence {
 
+    private static final Optional<DownloadError> DOWNLOAD_ERROR = Optional.absent();
```
danybony: Better to rename it to `NO_DOWNLOAD_ERROR`

library/src/main/java/com/novoda/downloadmanager/DownloadsBatchPersistence.java
```diff
@@ -33,125 +40,167 @@ void persistAsync(DownloadBatchTitle downloadBatchTitle,
...
+                downloadBatchTitle,
+                downloadedDateTimeInMillis,
+                currentBytesDownloaded,
+                totalBatchSizeBytes,
+                status,
+                notificationSeen,
+                DOWNLOAD_ERROR
+        );
+
+        liteDownloadBatchStatus.update(currentBytesDownloaded, totalBatchSizeBytes);
```
danybony: This is not needed, you already set everything in the constructor above

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -49,73 +53,144 @@ void setCallback(DownloadBatchStatusCallback callback) {
...
 
         if (totalBatchSizeBytes == 0) {
-            totalBatchSizeBytes = getTotalSize(downloadFiles);
+            totalBatchSizeBytes = getTotalSize(downloadFiles, downloadBatchStatus);
         }
 
-        if (totalBatchSizeBytes <= ZERO_BYTES) {
-            deleteBatchIfNeeded();
-            processNetworkError();
+        if (shouldAbortAfterGettingTotalBatchSize(downloadBatchStatus, downloadsBatchPersistence, callback, totalBatchSizeBytes)) {
```
MechaRyRy, line 70: This is exactly what I envisioned when all of these issues started cropping up. Perform an action, check that we can continue 💯

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -152,28 +227,30 @@ private boolean networkError() {
...
             bytesDownloaded += entry.getValue();
         }
         return bytesDownloaded;
     }
 
-    private void notifyCallback(InternalDownloadBatchStatus downloadBatchStatus) {
+    private static void notifyCallback(DownloadBatchStatusCallback callback, InternalDownloadBatchStatus downloadBatchStatus) {
         if (callback != null) {
-            callback.onUpdate(downloadBatchStatus);
+            callback.onUpdate(downloadBatchStatus.copy());
```
MechaRyRy: 💯

library/src/main/java/com/novoda/downloadmanager/DownloadFile.java
```diff
@@ -60,7 +62,14 @@ void download(Callback callback) {
 
         fileSize.setCurrentSize(filePersistence.getCurrentSize(filePath));
 
-        persistSync();
+        if (downloadFileStatus.isMarkedAsDeleted()) {
+            return;
+        }
+
+        Log.v("persist file " + downloadFileId.rawId() + ", with status: " + downloadFileStatus.status());
+        if (persist()) {
```
MechaRyRy: I don't follow this bit. If it is persisted then return?

library/src/main/java/com/novoda/downloadmanager/LiteDownloadBatchStatus.java
```diff
@@ -81,22 +89,29 @@ public long downloadedDateTimeInMillis() {
         return downloadedDateTimeInMillis;
     }
 
+    @WorkerThread
```
danybony: Why this, if we are calling the async update inside?

library/src/main/java/com/novoda/downloadmanager/LiteDownloadBatchStatus.java
```diff
@@ -105,26 +120,43 @@ public void markAsDeleted() {
...
     @Override
     public void markAsError(Optional<DownloadError> downloadError, DownloadsBatchStatusPersistence persistence) {
         this.status = Status.ERROR;
         this.downloadError = downloadError;
-        updateStatus(status, persistence);
+        updateStatusAsync(status, persistence);
     }
 
+    @WorkerThread
```
MechaRyRy: Why does this one need to be on a `workerThread` if it immediately goes into `Async`? Same with the others here.

danybony (changes requested):

MechaRyRy (commented):
Man, this PR is 👌 

It resolves a lot of problems in the library and I feel this needs more time to be reviewed. I'm going to review again on Monday with fresh eyes and then I'd like it if you, @zegnus, can take me through it too?

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
MechaRyRy: I would also be interested in seeing this 😄 

I don't mind statics for the most part, it highlights that this method is stateless.

library/src/main/java/com/novoda/downloadmanager/DownloadFile.java
zegnus: this is wrong actually, good spot

library/src/main/java/com/novoda/downloadmanager/LiteDownloadBatchStatus.java
zegnus: Some of them do not apply anymore, on it

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
zegnus: Is zero because that file is paused and it shouldn't be processed, but I will double check the status when we kill the app

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
zegnus: Will change by first call order

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
zegnus: As discussed online, this is part also of the process for removing this logic from this big class

library/src/main/java/com/novoda/downloadmanager/DownloadBatchFactory.java
zegnus: These are going to now default values, along with the bytes downloaded and total batch size

library/src/main/java/com/novoda/downloadmanager/DownloadsBatchPersistence.java
zegnus: will do

library/src/main/java/com/novoda/downloadmanager/DownloadsBatchPersistence.java
zegnus: yup!

library/src/main/java/com/novoda/downloadmanager/LiteDownloadBatchStatus.java
zegnus: removed

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
zegnus: checked, it's 👍

library/src/main/java/com/novoda/downloadmanager/DownloadFile.java
```diff
@@ -60,7 +62,16 @@ void download(Callback callback) {
...
         fileSize.setCurrentSize(filePersistence.getCurrentSize(filePath));
 
-        persistSync();
+        if (downloadFileStatus.isMarkedAsDeleted()) {
+            return;
+        }
+
+        Log.v("persist file " + downloadFileId.rawId() + ", with status: " + downloadFileStatus.status());
+        boolean persist = persist();
+        if (!persist) {
```
MechaRyRy: Can we inline the variable? Looks like the only place it is used.

zegnus: yes, I didn't do it in the first place because of the negation, will do

MechaRyRy (approved):

danybony (approved):

PR #371, opened 2018-03-20
Fix delete status is ignored by download abort checks

zegnus:
**Problem**

A queue batch is marked for deletion when the queue starts should be deleted (as should be marked for deletion) but the check says that is downloading

```
delete request for batch -468903206, mark as deleting from QUEUED 
delete request for batch -468903206, status: DELETING, should be deleting 
mark file as deleted for batchId: -468903206 
mark file as deleted for batchId: -468903206 
mark file as deleted for batchId: -468903206 
mark file as deleted for batchId: -468903206 
mark file as deleted for batchId: -468903206 
mark file as deleted for batchId: -468903206 
delete request for batch end -468903206, status: DOWNLOADING, should be deleting 
```

**Solution**

The check that says that is downloading is due to the fact that there is a local variable extracting the status for the entire check method `shouldAbortStartingBatch`, so when the status has changed in the main thread, and the check is finally executed, the status is outdated.

The solution has been to actually check the real status of all the checks and avoid creating a local variable.

The rest of the code changes have been small code improvements, remove unnecessary code and add better logging

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
@@ -22,6 +22,7 @@ import static com.novoda.downloadmanager.DownloadBatchStatus.Status.WAITING_FOR_
 class DownloadBatch {
 
     private static final int ZERO_BYTES = 0;
+    private static final String STATUS = "status";
 
     private final Map<DownloadFileId, Long> fileBytesDownloadedMap;
     private final InternalDownloadBatchStatus downloadBatchStatus;
@@ -54,10 +55,10 @@ class DownloadBatch {
 
     void download() {
         String rawBatchId = downloadBatchStatus.getDownloadBatchId().rawId();
-        Log.v("start download " + rawBatchId + ", status: " + downloadBatchStatus.status());
+        Log.v("start sync download " + rawBatchId + ", " + STATUS + " " + downloadBatchStatus.status());
 
         if (shouldAbortStartingBatch(connectionChecker, callback, downloadBatchStatus, downloadsBatchPersistence)) {
-            Log.v("abort starting download " + rawBatchId);
+            Log.v("abort starting download " + rawBatchId + ", " + STATUS + " " + downloadBatchStatus.status());
             return;
         }
 
@@ -67,8 +68,12 @@ class DownloadBatch {
             totalBatchSizeBytes = getTotalSize(downloadFiles, downloadBatchStatus);
         }
 
+        Log.v("batch " + downloadBatchStatus.getDownloadBatchId().rawId()
+                + " " + STATUS + " " + downloadBatchStatus.status()
+                + " totalBatchSize " + totalBatchSizeBytes);
+
         if (shouldAbortAfterGettingTotalBatchSize(downloadBatchStatus, downloadsBatchPersistence, callback, totalBatchSizeBytes)) {
-            Log.v("abort after getting total batch size download " + rawBatchId);
+            Log.v("abort after getting total batch size download " + rawBatchId + ", " + STATUS + " " + downloadBatchStatus.status());
             return;
         }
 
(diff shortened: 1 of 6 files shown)


library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -22,6 +22,7 @@
 class DownloadBatch {
 
     private static final int ZERO_BYTES = 0;
+    private static final String STATUS = "status";
```
zegnus, line 25: this was a PMD warning

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -86,31 +91,34 @@ void download() {
...
         callbackThrottle.stopUpdates();
-        Log.v("end download " + rawBatchId);
+        Log.v("end sync download " + rawBatchId);
     }
 
     private static boolean shouldAbortStartingBatch(ConnectionChecker connectionChecker,
                                                     DownloadBatchStatusCallback callback,
                                                     InternalDownloadBatchStatus downloadBatchStatus,
                                                     DownloadsBatchPersistence downloadsBatchPersistence) {
-        DownloadBatchStatus.Status status = downloadBatchStatus.status();
```
zegnus, line 96: this is the actual fix

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -122,8 +130,10 @@ private static boolean shouldAbortStartingBatch(ConnectionChecker connectionChec
     private static void deleteBatchIfNeeded(InternalDownloadBatchStatus downloadBatchStatus,
                                             DownloadsBatchPersistence downloadsBatchPersistence,
                                             DownloadBatchStatusCallback callback) {
-        if (downloadBatchStatus.status() == DELETING && downloadsBatchPersistence.deleteSync(downloadBatchStatus)) {
+        if (downloadBatchStatus.status() == DELETING) {
+            Log.v("sync delete and mark as deleted batch " + downloadBatchStatus.getDownloadBatchId().rawId());
             downloadBatchStatus.markAsDeleted();
+            downloadsBatchPersistence.deleteSync(downloadBatchStatus);
```
zegnus, line 136: we need to mark the batch immediately as deleting, this is why this check is now in this position instead of being part of the `if` check. For the callback, it does not matter the result of the deletion as it has marked as deleted anyway and that is what it should be propagated.

library/src/main/java/com/novoda/downloadmanager/DownloadBatchFactory.java
```diff
@@ -69,8 +68,6 @@ static DownloadBatch newInstance(Batch batch,
             downloadFiles.add(downloadFile);
         }
 
-        downloadFiles = Collections.unmodifiableList(downloadFiles);
```
zegnus, line 72: don't needed, nobody else will add files once the batch is created

library/src/main/java/com/novoda/downloadmanager/DownloadManager.java
```diff
@@ -97,23 +99,24 @@ public void pause(DownloadBatchId downloadBatchId) {
...
+            Log.v("abort resume batch " + downloadBatchId + " will not be resume as it does not exists in the running batches map");
             return;
         }
 
         if (downloadBatch.status().status() == DownloadBatchStatus.Status.DOWNLOADING) {
+            Log.v("abort resume batch " + downloadBatchId + " will not be resume as it's already downloading");
             return;
         }
 
-        downloadBatchMap.remove(downloadBatchId);
```
zegnus, line 107: this is not needed and makes it complex to understand. I've added a check in the actual `downloader.download(..)` method for adding it into the map if is necessary 🍎

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerDownloader.java
```diff
@@ -71,7 +71,11 @@ public void download(Batch batch, Map<DownloadBatchId, DownloadBatch> downloadBa
     }
 
     public void download(DownloadBatch downloadBatch, Map<DownloadBatchId, DownloadBatch> downloadBatchMap) {
-        downloadBatchMap.put(downloadBatch.getId(), downloadBatch);
+        DownloadBatchId downloadBatchId = downloadBatch.getId();
+        if (!downloadBatchMap.containsKey(downloadBatchId)) {
```
zegnus, line 75: 🍎

library/src/test/java/com/novoda/downloadmanager/DownloadManagerTest.java
```diff
@@ -222,15 +222,6 @@ public void doesNotResume_whenBatchIsAlreadyDownloading() {
         inOrder.verifyNoMoreInteractions();
     }
 
-    @Test
-    public void removesBatchFromInternalList_whenResuming() {
```
zegnus, line 226: because of 🍎 this is not needed

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -54,10 +55,10 @@ void setCallback(DownloadBatchStatusCallback callback) {
 
     void download() {
         String rawBatchId = downloadBatchStatus.getDownloadBatchId().rawId();
-        Log.v("start download " + rawBatchId + ", status: " + downloadBatchStatus.status());
+        Log.v("start sync download " + rawBatchId + ", " + STATUS + " " + downloadBatchStatus.status());
```
MechaRyRy, line 58: FYI for log I think you can just do `,` instead of concatenating with `+`

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -86,31 +91,34 @@ void download() {
...
     }
 
     private static boolean shouldAbortStartingBatch(ConnectionChecker connectionChecker,
                                                     DownloadBatchStatusCallback callback,
                                                     InternalDownloadBatchStatus downloadBatchStatus,
                                                     DownloadsBatchPersistence downloadsBatchPersistence) {
-        DownloadBatchStatus.Status status = downloadBatchStatus.status();
-
-        if (status == DELETED) {
+        // WARNING: do not extract downloadBatchStatus.status() as a local variable, this will
```
MechaRyRy, line 101: Happens in a `thread` worth highlighting with an annotation? I'm not a fan of comments 😄

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
MechaRyRy: It doesn't matter? What happens if the files aren't deleted?

library/src/main/java/com/novoda/downloadmanager/DownloadManager.java
MechaRyRy: > method for adding it into the map if is necessary.

but here you are removing it 😄

MechaRyRy (approved):

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
danybony: Yeah, with NoTils... but also with the new logger, so no problem 👍

danybony (approved):

PR #372, opened 2018-03-20
Custom logger

danybony:
**Problem**

Logs in the app don't have a common tag allowing a quick filter in logcat

**Solution**

Allow the client of the library to specify a custom `LogHandle`.
In the demo app for example all the logs have a common tag prepended.
Also in this PR: remove the Notils dependency from both library and demo

### DIFF
--- a/demo-simple/build.gradle
+++ b/demo-simple/build.gradle
@@ -33,7 +33,6 @@ android {
 
 dependencies {
     implementation fileTree(include: ['*.jar'], dir: 'libs')
-    implementation 'com.novoda:notils:2.2.16'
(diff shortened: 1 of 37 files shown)


MechaRyRy (commented):
Discussed on Slack recording here for everyone else. We should make sure that the `TAG` is composite, that way I can search by either `download-manager` or by the `class`.

We should also ensure that we are using a single log mechanism through the library, i.e. only using `Logger`.

MechaRyRy:
@danybony still using `Log` in some places?

danybony:
Logs are only used in the Demo app files

MechaRyRy (approved):

PR #374, opened 2018-03-21
File id per batch

MechaRyRy:
## Problem
- It's possible for a client app to specify the same `FileId` across batches. At the moment we just replace in the db when the `FileId` matches an already stored id.

- `PartialMigrations` are stored in the original file locations from v1. If batches share the same file then removing one will invalidate the other 😬 

- v1 files can be shared across batches 😬 so if we delete them in the partial downloads then they don't exist for the complete 😱 

## Solution
- Create a composite key for a `DownloadFile` that consists of the `BatchId` and the `FileId`, meaning that a `FileId` is only unique per batch rather than globally. 

- Use a base path specified by clients for v1. Strip the base path from the absolute path from v1 leaving the relative path. To this relative path we then prepend the batch identifier at which point we recreate the absolute path. We then move all existing files from v1 to this new v2 location.

- Move the complete downloads before removing all of the files from v1.


library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
```diff
@@ -22,17 +22,19 @@
     private static final int TITLE_COLUMN = 1;
     private static final int MODIFIED_TIMESTAMP_COLUMN = 2;
 
-    private static final String DOWNLOADS_QUERY = "SELECT uri, _data, notificationextras FROM Downloads WHERE batch_id = ?";
+    private static final String DOWNLOADS_QUERY = "SELECT uri, hint, notificationextras FROM Downloads WHERE batch_id = ?";
```
MechaRyRy, line 25: `hint` is never empty from v1, it's populated at the time the request is made. `_data` by comparison is only populated when an individual file has completed download.

library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
```diff
@@ -78,12 +82,15 @@
                         }
                         newBatchBuilder.addFile(originalNetworkAddress).apply();
 
-                        FilePath filePath = new LiteFilePath(originalFileLocation);
-                        long rawFileSize = filePersistence.getCurrentSize(filePath);
+                        FilePath originalFilePath = new LiteFilePath(sanitizedOriginalFileLocation);
+                        FilePath newFilePath = MigrationPathExtractor.extractMigrationPath(basePath, originalFilePath.path(), downloadBatchId);
```
MechaRyRy, line 86: This creates a file path using the `base` and a `downloadBatchId` so that it appears in the form:
```
base/batchIdentifier/relativePathFromBase
```

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
```diff
@@ -218,4 +219,38 @@ private void deleteVersionOneDatabase(InternalMigrationStatus migrationStatus, S
         database.deleteDatabase();
     }
 
+    private void migrateV1FilesToV2Location(FilePersistence filePersistence, Migration migration) {
```
MechaRyRy, line 221: Turns out we still need to move the files 😂 we need them to appear per batch rather than globally, which is what v1 was doing.

library/src/main/java/com/novoda/downloadmanager/MigrationStoragePathSanitizer.java
```diff
@@ -0,0 +1,16 @@
...
+final class MigrationStoragePathSanitizer {
+
+    private static final String FILE_SCHEME_TO_REMOVE = "file:";
+    private static final String EMPTY = "";
+
+    private MigrationStoragePathSanitizer() {
+        // Uses static utility methods.
+    }
+
+    static String sanitize(String originalFilePath) {
```
MechaRyRy, line 12: This just removes the `file:` from the original v1 file location path.

library/src/main/java/com/novoda/downloadmanager/RoomFile.java
```diff
@@ -4,23 +4,23 @@
...
 import android.arch.persistence.room.ForeignKey;
 import android.arch.persistence.room.Index;
-import android.arch.persistence.room.PrimaryKey;
 import android.support.annotation.NonNull;
 
 import static android.arch.persistence.room.ForeignKey.CASCADE;
 
 @SuppressWarnings("checkstyle:visibilitymodifier") // Accessors will add a lot of boilerplate code.
 @Entity(
+        primaryKeys = {"file_id", "batch_id"},
```
MechaRyRy, line 13: Combination primary key so that a file identifier is unique per batch rather than globally.

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
```diff
@@ -73,8 +78,8 @@ public void run() {
         migrationStatus.markAsMigrating();
         onUpdate(migrationStatus);
 
-        migratePartialDownloads(migrationStatus, database, partialMigrations, downloadsPersistence, basePath);
-        migrateCompleteDownloads(migrationStatus, database, completeMigrations, downloadsPersistence, basePath);
+        migrateCompleteDownloads(migrationStatus, database, completeMigrations, downloadsPersistence, filePersistence);
```
MechaRyRy, line 80: `Complete` downloads need to be done first because otherwise we remove the files if they are shared by the partial download 😬

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -35,6 +36,10 @@
 public class MainActivity extends AppCompatActivity {
 
     private static final String TAG = MainActivity.class.getSimpleName();
+
+    @SuppressLint("SdCardPath")
+    private static final String V1_BASE_PATH = "/data/data/com.novoda.downloadmanager.demo.simple/files/Pictures/";
```
danybony, line 41: Worth making this dynamic using context, in case we refactor the package of the demo? (I can see us removing the `simple` part in the future)

library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
danybony: Weird but 👍

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
```diff
@@ -13,18 +16,21 @@
 
     private static final String TABLE_BATCHES = "batches";
     private static final String WHERE_CLAUSE_ID = "_id = ?";
+    private static final int RANDOMLY_CHOSEN_BUFFER_SIZE_THAT_SEEMS_TO_WORK = 4096;
```
danybony, line 18: 😂

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
```diff
@@ -218,4 +219,38 @@ private void deleteVersionOneDatabase(InternalMigrationStatus migrationStatus, S
...
+                byte[] bytes = new byte[RANDOMLY_CHOSEN_BUFFER_SIZE_THAT_SEEMS_TO_WORK];
+
+                // read the v1 file
+                int readLast = 0;
+                while (readLast != -1) {
+                    readLast = inputStream.read(bytes);
+                    if (readLast != 0 && readLast != -1) {
+                        // write the v1 file to the v2 location
+                        filePersistence.write(bytes, 0, readLast);
+                        bytes = new byte[RANDOMLY_CHOSEN_BUFFER_SIZE_THAT_SEEMS_TO_WORK];
```
danybony: I might be wrong, but you shouldn't need to recreate this every time.
Its content is overwritten anyway in `inputStream.read(bytes)`, and you are writing only the freshest content (the new `readLast` bytes)

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
```diff
@@ -218,4 +219,38 @@ private void deleteVersionOneDatabase(InternalMigrationStatus migrationStatus, S
...
+                    }
+                }
+            } catch (IOException e) {
+                Log.e(getClass().getSimpleName(), e.getMessage());
```
danybony: Logger

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
```diff
@@ -218,4 +219,38 @@ private void deleteVersionOneDatabase(InternalMigrationStatus migrationStatus, S
...
+            } finally {
+                try {
+                    filePersistence.close();
+                    if (inputStream != null) {
+                        inputStream.close();
+                    }
+                } catch (IOException e) {
+                    Log.e(getClass().getSimpleName(), e.getMessage());
```
danybony: Logger

library/src/main/java/com/novoda/downloadmanager/MigrationPathExtractor.java
```diff
@@ -0,0 +1,46 @@
...
+    private MigrationPathExtractor() {
+        // Uses static utility methods.
+    }
+
+    static FilePath extractMigrationPath(String basePath, String assetPath, DownloadBatchId downloadBatchId) {
+        String relativePath = extractRelativePath(basePath, assetPath);
+        String relativePathWithBatchId = prependBatchIdTo(relativePath, downloadBatchId);
+        String fileName = extractFileName(assetPath);
+        String absolutePath = basePath + PATH_SEPARATOR + relativePathWithBatchId + fileName;
+        String sanitizedAbsolutePath = absolutePath.replaceAll("//", "/");
```
danybony: 👍

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
MechaRyRy: Good question but this actually has nothing to do with the context, lint flags this because of the use of `data`. This `V1_BASE_PATH` is the base path that is used in the v1 database, we would only need to change this if we change the v1 database.

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
MechaRyRy: I think this is done to avoid the scenario where:

write one -> 200 bytes
write two -> 100 bytes

but I have 100 bytes that were written previously, so it will do the `100 bytes` PLUS whatever is the difference of the first, the missing `100 bytes`. Does that make sense?

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
MechaRyRy: d07b48f

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
MechaRyRy: d07b48f

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
danybony: Ahh right, I put the comment before reading the rest of the PR... it's OK to have it like this 👍

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
danybony: I don't think that will happen, because let's say 200 bytes is your buffer size.
Then you have 

1. readLast = 200 -> write bytes[0..199]
2. readlast = 100 -> bytes[100..199] contains outdated dat but we only write bytes[0..99]

MechaRyRy: 😄 on second look I agree.

library/src/main/java/com/novoda/downloadmanager/MigrationPathExtractor.java
```diff
@@ -0,0 +1,46 @@
+package com.novoda.downloadmanager;
+
+import java.io.File;
+
+final class MigrationPathExtractor {
+
+    private static final String PATH_SEPARATOR = "/";
```
zegnus: `File.pathSeparatorChar`

MechaRyRy: Errr this replaces with `:`

danybony (approved):

library/src/main/java/com/novoda/downloadmanager/MigrationPathExtractor.java
MechaRyRy: f743353

zegnus (approved):

PR #376, opened 2018-03-22
Duplicated urls in migration

danybony:
### Problem
Some client app might have the same URL multiple times in a single batch, with different file ids.
At the moment we filter these, resulting in missing download files.

### Solution
Check for only the pair (URL, FileId) and allow duplicated urls with different id on migration

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
+++ b/library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
@@ -83,12 +83,12 @@ class MigrationExtractor {
                             fileIds.add(originalFileId);
                         }
 
-                        if (originalFileId != null) {
+                        if (originalFileId == null) {
                             newBatchBuilder.addFile(originalNetworkAddress)
-                                    .withDownloadFileId(DownloadFileIdCreator.createFrom(originalFileId))
                                     .apply();
                         } else {
                             newBatchBuilder.addFile(originalNetworkAddress)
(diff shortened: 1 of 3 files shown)


library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
```diff
@@ -75,12 +76,21 @@
...
 
-                        if (uris.contains(originalNetworkAddress)) {
+                        if (uris.contains(originalNetworkAddress) && fileIds.contains(originalFileId)) {
                             continue;
                         } else {
                             uris.add(originalNetworkAddress);
+                            fileIds.add(originalFileId);
+                        }
+
+                        if (originalFileId != null) {
```
zegnus: invert the if ;)

library/src/main/java/com/novoda/downloadmanager/PartialDownloadMigrationExtractor.java
```diff
@@ -60,12 +61,21 @@
...
 
-                if (uris.contains(uri)) {
+                if (uris.contains(uri) && fileIds.contains(originalFileId)) {
                     continue;
                 } else {
                     uris.add(uri);
+                    fileIds.add(originalFileId);
+                }
+
+                if (originalFileId != null) {
```
zegnus: invert the if

MechaRyRy (approved):

zegnus (approved):

PR #383, opened 2018-04-03
Fix v1 to v2 complete migrations

zegnus:
**Problem**

When migrating batches that contain the same files, the original path is wrong and the migrated file cannot be found.

**Solution**

We are going to use the original path from the `_data` column and the new path from the `hint` column, in this way we can always read the original file from the original path and then we will construct a new path combining the `batchId` and `hint` providing a unique new file path that can handle duplicates

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
+++ b/library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
@@ -22,10 +22,11 @@ class MigrationExtractor {
     private static final int TITLE_COLUMN = 1;
     private static final int MODIFIED_TIMESTAMP_COLUMN = 2;
 
-    private static final String DOWNLOADS_QUERY = "SELECT uri, hint, notificationextras FROM Downloads WHERE batch_id = ?";
+    private static final String DOWNLOADS_QUERY = "SELECT uri, _data, hint, notificationextras FROM Downloads WHERE batch_id = ?";
     private static final int NETWORK_ADDRESS_COLUMN = 0;
-    private static final int FILE_LOCATION_COLUMN = 1;
-    private static final int FILE_ID_COLUMN = 2;
+    private static final int FILE_ORIGINAL_LOCATION_COLUMN = 1;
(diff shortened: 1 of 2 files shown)


library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
```diff
@@ -68,8 +70,13 @@
...
                         String originalNetworkAddress = downloadsCursor.getString(NETWORK_ADDRESS_COLUMN);
-                        String originalFileLocation = downloadsCursor.getString(FILE_LOCATION_COLUMN);
-                        String sanitizedOriginalFileLocation = MigrationStoragePathSanitizer.sanitize(originalFileLocation);
+
+                        // hint: file:/data/user/0/com.channel4.ondemand/files/all4/23241337
+                        String originalUniqueFileLocation = downloadsCursor.getString(FILE_UNIQUE_LOCATION_COLUMN);
+                        String sanitizedOriginalUniqueFileLocation = MigrationStoragePathSanitizer.sanitize(originalUniqueFileLocation);
+
+                        // _data: /data/user/0/com.channel4.ondemand/files/all4/23241337-1
+                        String originalFileLocation = downloadsCursor.getString(FILE_ORIGINAL_LOCATION_COLUMN);
```
danybony, line 79: Did you double check that this one is always populated? I seem to remember that we had issues because sometimes it was empty (and that's the reason why we moved to `hint`)

danybony (commented):
LGTM, just one question

library/src/main/java/com/novoda/downloadmanager/MigrationExtractor.java
MechaRyRy: This should be fine for **Completed Downloads** the `_data` is only populated when a download is complete, from what I saw of version one. 😄 I did hesitate changing this originally but decided to so that it matched the `PartialMigrationExtractor`, now I see I shouldn't have bothered 😬

MechaRyRy (approved):

danybony (approved):

PR #384, opened 2018-04-05
Load batch size before download

danybony:
### Problem
The batch size was always 0 from the moment it was added by the client application up to the moment that batch started actually downloading (it was always 0 when queued)

### Solution
Try to load the total batch size during queueing. If it's not possible (because of missing network connection) the size is fetched when the download starts, as before.

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -15,6 +15,7 @@ android {
     }
 
     lintOptions {
+        disable 'GradleDependency'
         lintConfig teamPropsFile('static-analysis/lint-config.xml')
         abortOnError true
         warningsAsErrors true
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -36,7 +36,7 @@ dependencies {
     annotationProcessor 'android.arch.persistence.room:compiler:1.0.0'
     implementation 'com.novoda:merlin:1.1.6'
     implementation 'com.facebook.stetho:stetho:1.5.0'
-    implementation 'com.android.support:support-v4:27.1.0'
+    implementation 'com.android.support:support-v4:27.1.1'
     implementation 'com.squareup.okhttp3:okhttp:3.10.0'
     implementation 'android.arch.persistence.room:runtime:1.0.0'
     implementation('com.evernote:android-job:1.2.5') {
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -15,7 +15,6 @@ android {
     }
 
     lintOptions {
-        disable 'GradleDependency'
         lintConfig teamPropsFile('static-analysis/lint-config.xml')
         abortOnError true
         warningsAsErrors true
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -36,7 +36,6 @@ dependencies {
     annotationProcessor 'android.arch.persistence.room:compiler:1.0.0'
     implementation 'com.novoda:merlin:1.1.6'
     implementation 'com.facebook.stetho:stetho:1.5.0'
-    implementation 'com.android.support:support-v4:27.1.1'
     implementation 'com.squareup.okhttp3:okhttp:3.10.0'
     implementation 'android.arch.persistence.room:runtime:1.0.0'
     implementation('com.evernote:android-job:1.2.5') {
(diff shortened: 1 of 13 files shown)


library/src/main/java/com/novoda/downloadmanager/DownloadBatchFactory.java
```diff
@@ -89,6 +89,22 @@ static DownloadBatch newInstance(Batch batch,
         );
     }
 
+    private static long getTotalSize(List<DownloadFile> downloadFiles, DownloadBatchId downloadBatchId) {
+        long totalBatchSize = 0;
+        for (DownloadFile downloadFile : downloadFiles) {
```
MechaRyRy: Isn't this bit of code used somewhere else too?

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerDownloader.java
```diff
@@ -54,28 +54,40 @@
...
 
-        downloadBatchMap.put(downloadBatch.getId(), downloadBatch);
-        download(downloadBatch, downloadBatchMap);
+            downloadBatchMap.put(downloadBatch.getId(), downloadBatch);
+            downloadSync(downloadBatch, downloadBatchMap);
+        });
     }
 
-    public void download(DownloadBatch downloadBatch, Map<DownloadBatchId, DownloadBatch> downloadBatchMap) {
+    private void downloadSync(DownloadBatch downloadBatch, Map<DownloadBatchId, DownloadBatch> downloadBatchMap) {
```
MechaRyRy: Not sure I agree with this name. I would avoid naming something as synchronous unless it is exposed publicly and clients of the application need to handle the threading themselves. We know when something is synchronous or not by whether we have added it to an executor and whether it is annotated with the `@WorkerThread`

library/src/main/java/com/novoda/downloadmanager/DownloadBatchFactory.java
MechaRyRy: You could probably extract the whole thing from the `DownloadBatch` into an external collaborator, I can't imagine that the status checks would be a problem here too.

danybony: It shouldn't be a problem, I avoided reusing that just because the status checks are not relevant here. But can extract anyway, to avoid duplication in logic

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerDownloader.java
danybony: I agree, and I added the `@WorkerThread` annotation. However, I'd need another name for to distinguish this and the `download()` method submitting to the executor

ouchadam: `submitDownload`?

For the sync naming, they do this a lot in nodejs land where there's a 
```
readFile(callback)
readFileSync()
```

library/src/main/java/com/novoda/downloadmanager/DownloadBatchFactory.java
```diff
@@ -68,13 +67,15 @@ static DownloadBatch newInstance(Batch batch,
             downloadFiles.add(downloadFile);
         }
 
+        DownloadBatchStatus.Status status = DownloadBatchStatus.Status.UNKNOWN;
```
ouchadam: What makes this status UNKNOWN instead of DOWNLOAD_ERROR?

danybony: It's UNKNOWN because it's just being created, so the status is not defined yet. It will be added to queue in the downloader later on

ouchadam (approved):

library/build.gradle
```diff
@@ -15,6 +15,7 @@ android {
     }
 
     lintOptions {
+        disable 'GradleDependency'
```
MechaRyRy: What is this?

MechaRyRy (changes requested):
Would avoid disabling checks, seems we need to change the lint configuration to suit our needs.

danybony:
@zegnus pointed me to a possible concurrency issue, so will revert some of the changes and try another path

danybony:
* @zegnus now I submit a dedicated runnable to the executor to fetch the total batch size, so that the creation of the batch and insertion to the map is sequential
* @Mecharyry updated the obsolete gradle dependency check severity to `informational`. It's still displayed in the Lint report without making the build fail, like this:
image

* @ouchadam hello there 👋

MechaRyRy (approved):

zegnus (approved):

PR #385, opened 2018-04-06
Insufficient Storage Space

MechaRyRy:
## Problem
When a device becomes low on storage space we start to see crashes especially around the SQLite database. In an attempt to prevent this from happening we want to ensure a certain `buffer` is free to the system to ensure that this does not happen because of us 😄

## Solution
Add a buffer amount, default 10%, to which the system must have free in order to download a particular file. This is a simple check when creating the file to see if the available space is below this calculated threshold.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/ExternalFilePersistence.java
+++ b/library/src/main/java/com/novoda/downloadmanager/ExternalFilePersistence.java
@@ -2,8 +2,6 @@ package com.novoda.downloadmanager;
 
 import android.content.Context;
 import android.os.Environment;
-import android.os.StatFs;
-import android.support.annotation.FloatRange;
 import android.support.annotation.Nullable;
 import android.support.v4.content.ContextCompat;
 
@@ -19,15 +17,15 @@ class ExternalFilePersistence implements FilePersistence {
     private static final boolean APPEND = true;
 
     private Context context;
-    private float percentageStorageRemaining;
+    private StorageRequirementsRule storageRequirementsRule;
 
     @Nullable
     private FileOutputStream fileOutputStream;
 
     @Override
-    public void initialiseWith(Context context, @FloatRange(from = 0.0, to = 0.5) float percentageOfStorageRemaining) {
+    public void initialiseWith(Context context, StorageRequirementsRule storageRequirementsRule) {
         this.context = context.getApplicationContext();
-        this.percentageStorageRemaining = percentageOfStorageRemaining;
+        this.storageRequirementsRule = storageRequirementsRule;
     }
 
     @Override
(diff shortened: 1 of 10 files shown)


library/src/main/java/com/novoda/downloadmanager/StorageCapacityReader.java
```diff
@@ -0,0 +1,20 @@
...
+import android.os.Build;
+import android.os.StatFs;
+
+final class StorageCapacityReader {
+
+    private StorageCapacityReader() {
+        // Uses static factory method.
+    }
+
+    static long storageCapacityInBytes(StatFs statFs) {
```
zegnus: How come this method is the same for calculating the storage capacity for both internal and external?

Like, if I have a device with 16GB internal storage with a 64GB with an SD card, those are the different numbers that this method should return right?

MechaRyRy: 😂 let me try again.

`external` is not `sd` cards, let's get that cleared up first 😄 

`external` -> publicly available space on the device. Not guaranteed to be an sd card. Not guaranteed to always be present.
`internal` -> root / app only access.

When looking for the `external` size it does not differentiate between the `primary` and `secondary` storage, so it will include both. 

To be quite honest, this is the first time I have spent any great deal of time looking at it. Perhaps @Hutch4 would be best place to explain this. He can also verify if I am barking up the wrong tree 😄

MechaRyRy:
test this

library/src/main/java/com/novoda/downloadmanager/ExternalFilePersistence.java
```diff
@@ -43,14 +47,26 @@ public FilePersistenceResult create(FilePath absoluteFilePath, FileSize fileSize
...
+    private boolean hasViolatedStorageRequirements(File storageDirectory, FileSize downloadFileSize) {
+        StatFs statFs = new StatFs(storageDirectory.getPath());
+        long storageCapacityInBytes = StorageCapacityReader.storageCapacityInBytes(statFs);
+        long minimumStorageRequiredInBytes = (long) (storageCapacityInBytes * percentageStorageRemaining);
+        long usableStorageInBytes = storageDirectory.getUsableSpace();
+        long remainingStorageAfterDownloadInBytes = usableStorageInBytes - downloadFileSize.totalSize();
+
+        Logger.d("Storage capacity in bytes: ", storageCapacityInBytes);
+        Logger.d("Usable storage in bytes: ", usableStorageInBytes);
+        Logger.d("Minimum required storage in bytes: ", minimumStorageRequiredInBytes);
```
danybony: I'm ok to keep thees, but I'd use verbose level

library/src/main/java/com/novoda/downloadmanager/InternalFilePersistence.java
```diff
@@ -33,14 +37,26 @@ public FilePersistenceResult create(FilePath absoluteFilePath, FileSize fileSize
...
-        long usableSpace = context.getFilesDir().getUsableSpace();
-        if (usableSpace < fileSize.totalSize()) {
+        if (hasViolatedStorageRequirements(context.getFilesDir(), fileSize)) {
             return FilePersistenceResult.ERROR_INSUFFICIENT_SPACE;
         }
 
         return create(absoluteFilePath);
     }
 
+    private boolean hasViolatedStorageRequirements(File storageDirectory, FileSize downloadFileSize) {
```
danybony: This is duplicated, can you extract?

danybony (commented):
Just a couple of minor comments

danybony (approved):

zegnus (approved):

PR #387, opened 2018-04-26
Remove file name

MechaRyRy:
## Problem
We discovered when setting a `FileName` via the `Batch.Builder` in a client application that we don't actually use this to change anything internally in the `download-manager` 😬 This makes sense really as the `download-manager` is only concerned with being able to identify downloads with an `Identifier` and supply the location with the absolute path.

## Solution
Remove the unused `FileName` from the codebase. This includes performing a migration of the Room database to drop the `FileName` column. 

## Follow-up PR 
Fix the fluent api on the `Batch.Builder` so that we can supply an optional filename but it will create the relative path right there and then, rather than delaying to the internals of the `download-manager`.


library/src/main/java/com/novoda/downloadmanager/RoomAppDatabase.java
```diff
@@ -31,7 +34,20 @@ static RoomAppDatabase newInstance(Context context) {
...
+        )
+                .addMigrations(MIGRATION_V1_TO_V2)
+                .build();
     }
 
+    private static final Migration MIGRATION_V1_TO_V2 = new Migration(1, 2) {
+        @Override
+        public void migrate(@NonNull SupportSQLiteDatabase database) {
+            database.execSQL("CREATE TABLE IF NOT EXISTS `RoomFileTemp` (`file_id` TEXT NOT NULL, `batch_id` TEXT NOT NULL, `file_path` TEXT, `total_size` INTEGER NOT NULL, `url` TEXT, `persistence_type` TEXT, PRIMARY KEY(`file_id`, `batch_id`), FOREIGN KEY(`batch_id`) REFERENCES `RoomBatch`(`batch_id`) ON UPDATE NO ACTION ON DELETE CASCADE )");
+            database.execSQL("CREATE  INDEX `index_RoomFileTemp_batch_id` ON `RoomFileTemp` (`batch_id`)");
```
danybony: Double space between `create` and `index`

danybony (approved):
Nice one 👍

zegnus (approved):
Up to @danybony

PR #388, opened 2018-04-26
Rename methods on Batch.Builder fluent api

MechaRyRy:
## Problem
Some of the methods on the `Batch.Builder` are easily misinterpreted 😬 

## Solution
- Rename `addFile(networkAddress` to `downloadFrom(networkAddress)`
- Rename `withRelativePath(path)` to `saveTo(path)`
- Rename `withDownloadFileId(downloadFileId)` to `withIdentifier(downloadFileId)`, this could probably be more explicit but I couldn't think of a good name.

- `saveTo` has two implementations, one that takes just a path and another that takes a path and a file name.

## Next PR
Will make the `Optional<String> path` no longer `Optional`, moving the manipulation of the path from the `DownloadBatchFactory` to the `Batch.Builder` instead.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/BatchFile.java
+++ b/library/src/main/java/com/novoda/downloadmanager/BatchFile.java
@@ -70,6 +70,8 @@ public class BatchFile {
 
         Builder saveTo(String path);
 
+        Builder saveTo(String path, String fileName);
+
         Batch.Builder apply();
     }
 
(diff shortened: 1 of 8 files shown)


danybony (commented):
I like most of the changes, I'm only not sure about the `downloadFrom` one:  I don't think it reflects the fact that each URL correspond to a separate file. Reading it might be interpreted that each `downloadFrom` call is overwriting the previous one, instead of adding a new one

library/src/main/java/com/novoda/downloadmanager/BatchFile.java
```diff
@@ -61,14 +61,16 @@ public String toString() {
...
     }
 
     public interface Builder {
-        Builder withDownloadFileId(DownloadFileId downloadFileId);
+        Builder withIdentifier(DownloadFileId downloadFileId);
 
-        Builder withRelativePath(String relativePath);
+        Builder saveTo(String path);
+
+        Builder saveTo(String path, String fileName);
```
ouchadam, line 73: :100:

ouchadam (approved):
Way easier on the brain!

MechaRyRy:
@danybony I understand your concern. I would say that because these methods come from a `batch` and you need to state `apply` that we shouldn't have clients thinking that these will overwrite each other. In a later PR I have added a javadoc to these builders to make it super explicit what these methods do. Hopefully that alleviates your concerns 😃

danybony:
Yeah, probably it's because I'm used to the old naming... let's go with this then 👍

PR #389, opened 2018-04-26
Move path manipulation to Batch / File Builder

MechaRyRy:
## Problem 
Currently in the `DownloadBatchFactory` create a path from the `relativePath` or the `network` address based on the parameters that are specified in the `Batch.Builder`. Additionally we sanitise and prepend the `DownloadBatchId` to the this path. All of this is hidden in the factory and essentially means that the `Batch` we create from the `Builder` is passing partially invalid data. Ideally when we reach the factory we already want the `Batch` to contain only valid data.

## Solution
- Move the sanitisation of the `DownloadBatchId` to the `DownloadBatchIdCreator` and make it explicit in the naming, to avoid confusion for clients.

- Move the manipulation of the `path` to the `BatchFileBuilder`, making `BatchFile.path` mandatory.


danybony (approved):
💯

PR #390, opened 2018-04-26
Extract builders and javadoc

MechaRyRy:
## Problem
I discovered it was quite confusing working with the `Batch.Builders` because of the nested classes and interfaces 😬

## Solution
Pull up the builders and javadoc to avoid confusion.


library/src/main/java/com/novoda/downloadmanager/BatchFileBuilder.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Adds {@link BatchFile} to the parent {@link Batch} using the {@link BatchBuilder}
+ * and {@link BatchFileBuilder} fluent APIs.
+ */
+public interface BatchFileBuilder {
+
+    /**
+     * Sets {@link BatchFileBuilder} to build a {@link BatchFile} with a corresponding
+     * {@param downloadFileId} that can be used to lookup a download file belonging to
```
danybony: What about replacing 
`that can be used to lookup a download file`
with
`that can be later used to retrieve a downloaded file`?

library/src/main/java/com/novoda/downloadmanager/BatchFileBuilder.java
```diff
@@ -0,0 +1,46 @@
...
+     * a given {@link Batch}.
+     *
+     * @param downloadFileId to flag a {@link BatchFile} with.
+     * @return {@link BatchFileBuilder}.
+     */
+    BatchFileBuilder withIdentifier(DownloadFileId downloadFileId);
+
+    /**
+     * Sets {@link BatchFileBuilder} to build a {@link BatchFile} that will be saved
+     * to the given {@param path}.
```
danybony: Maybe specify here that the path is only referring to the directory that will contain the file.
And also, is it relative or absolute?

library/src/main/java/com/novoda/downloadmanager/LiteBatchBuilder.java
```diff
@@ -0,0 +1,49 @@
...
+        this.title = title;
+        this.batchFiles = batchFiles;
+    }
+
+    @Override
+    public void withFile(BatchFile batchFile) {
+        batchFiles.add(batchFile);
+    }
+
+    private BatchFileBuilder fileBuilder;
```
danybony: Field before the constructor

zegnus (approved):

danybony (commented):
💯

PR #391, opened 2018-05-01
Move MigrationExtractors to Activity

MechaRyRy:
## Problem
We are unable to do a release because our `Migration` implementation is very specific and does not allow clients to specify where they retrieve their V1 data from. 

## Solution 
Move the `MigrationExtractor`s to the `MainActivity` and have the `MigrationJob` read through the `List<Migration>` and add them to the `v2` database. When the `COMPLETE` status is emitted, delete the v1 database.

## Follow-up PRs **discussion needed**
This completes the whole flow but in a very simplistic manner. We no longer emit the deleting v1 database and it might be more appropriate to return a list of completed migrations so the clients can act on them if necessary.  We need to remember to remove the files for the partial migrations, this can be done either internally or left to the client.


demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -119,12 +127,37 @@ protected void onCreate(Bundle savedInstanceState) {
 
     private final MigrationCallback migrationCallback = migrationStatus -> {
         if (migrationStatus.status() == MigrationStatus.Status.COMPLETE) {
+            File databasePath = getDatabasePath("downloads.db");
```
danybony: Extract this to a constant, given it's used twice

danybony: "downloads.db" I mean

danybony (commented):
Ok, now it's definitely more flexible for the client, maybe even too much? Don'y you think it's possible to enable some sort of default mechanism?

MechaRyRy:
@danybony what do you mean by a default mechanism? The `v1` db could be used in so many different ways it would be very difficult to predict what fields the clients are using to represent our `v2` concepts 😕 of course, if you have ideas then I am happy to try them 😄

danybony:
No @Mecharyry you are right, I was just discussing the possibility of having something that could be used to cover the most common scenario.
But as you stated, it's not possible to predict how the library was used in v1 by clients, so good with giving them flexibility

danybony (approved):

MechaRyRy:
Going for a different approach.

PR #392, opened 2018-05-01
Feature file names

MechaRyRy:
This is a feature branch merge request consisting of the following PRs:

#387 Remove FileName, redundant and used for deleting files that should never be present in the first place.
#388 Rename methods on the BatchBuilder to avoid confusion
#389 Move path manipulation methods to builder, ensuring that the model is not immediately invalid when being given to the download-manager
#390 Extract the builders and javadoc


demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -40,8 +40,8 @@
     @SuppressLint("SdCardPath")
     private static final String V1_BASE_PATH = "/data/data/com.novoda.downloadmanager.demo.simple/files/Pictures/";
 
-    private static final DownloadBatchId BATCH_ID_1 = DownloadBatchIdCreator.createFrom("batch_id_1");
-    private static final DownloadBatchId BATCH_ID_2 = DownloadBatchIdCreator.createFrom("batch_id_2");
+    private static final DownloadBatchId BATCH_ID_1 = DownloadBatchIdCreator.createSanitizedFrom("batch_id_1");
```
zegnus: is there an `unsanitized` alternative?

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -40,8 +40,8 @@
...
+    private static final DownloadBatchId BATCH_ID_2 = DownloadBatchIdCreator.createSanitizedFrom("batch_id_2");
```
zegnus, line 44: is there an `unsanitized` alternative?

library/src/main/java/com/novoda/downloadmanager/DownloadBatchIdCreator.java
```diff
@@ -6,7 +6,12 @@ private DownloadBatchIdCreator() {
         // non-instantiable class
     }
 
-    public static DownloadBatchId createFrom(String rawId) {
-        return new LiteDownloadBatchId(rawId);
+    public static DownloadBatchId createSanitizedFrom(String rawId) {
```
zegnus: add Javadoc with what this does to the param

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
MechaRyRy: No, there was never one, we just hid this in the library a few layers deeper 😄

library/src/main/java/com/novoda/downloadmanager/DownloadBatchIdCreator.java
MechaRyRy: 604b41f

zegnus (approved):

PR #393, opened 2018-05-01
GZIP response content length

MechaRyRy:
## Problem
`OkHttp` automatically adds the `Accept-Encoding` header to any request that we make. In the case of `GZIP` okhttp will automatically decompress the file when downloading, this means that the `Content-Length` (compressed size) does not match the actual size when it is decompressed automatically. 

## Solution
Add the header `Accept-Encoding: Identity` -> Indicates the identity function (i.e. no compression, nor modification). This will give the correct `Content-Length` for our file size request.

## Screen Capture

Before | After
--- | --- 
gzip_before mp4 | gzip_after mp4

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/NetworkFileSizeRequester.java
+++ b/library/src/main/java/com/novoda/downloadmanager/NetworkFileSizeRequester.java
@@ -46,7 +46,7 @@ class NetworkFileSizeRequester implements FileSizeRequester {
     }
 
     private long requestFileSizeThroughHeaderRequest(String url) throws IOException {
-        NetworkRequest fileSizeRequest = requestCreator.createFileSizeRequest(url);
+        NetworkRequest fileSizeRequest = requestCreator.createFileSizeHeadRequest(url);
         HttpClient.NetworkResponse response = httpClient.execute(fileSizeRequest);
         long fileSize = ZERO_FILE_SIZE;
         if (response.isSuccessful()) {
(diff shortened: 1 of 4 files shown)


library/src/main/java/com/novoda/downloadmanager/WrappedOkHttpClient.java
```diff
@@ -24,6 +24,7 @@ public NetworkResponse execute(NetworkRequest request) throws IOException {
             requestBuilder = requestBuilder.head();
         }
 
+        requestBuilder.addHeader("Accept-Encoding", "identity");
```
MechaRyRy: This is the change that fixes all the problems 😄 saying we do not accept compression on our responses, we want them just the way they were meant to be 😄

library/src/main/java/com/novoda/downloadmanager/NetworkFileSizeRequester.java
```diff
@@ -34,11 +34,11 @@ public FileSize requestFileSize(String url) {
 
     private long executeRequestFileSize(String url) throws IOException {
         long fileSize = requestFileSizeThroughHeaderRequest(url);
-        if (fileSize == 0) {
-            Logger.w("filesize request through header returned zero, we'll try again " + url);
+        if (fileSize == UNKNOWN_CONTENT_LENGTH || fileSize == ZERO_FILE_SIZE) {
```
MechaRyRy, line 37: Using the constants and ensuring that we use both `0` and `-1` when determining something as invalid. The header request will return `-1` in the event that the `content-length` is not found.

library/src/main/java/com/novoda/downloadmanager/NetworkFileSizeRequester.java
```diff
@@ -34,11 +34,11 @@ public FileSize requestFileSize(String url) {
...
+            Logger.w(String.format("file size header request '%s' returned %s, we'll try with a body request", url, fileSize));
```
MechaRyRy, line 38: Reformatted the message now that it is triggered by more than `0` 😄

joetimmins (approved):
EZPZ 🍋

MechaRyRy:
Closing whilst I make some changes.

zegnus (approved):

danybony (commented):
💯 awesome!

PR #394, opened 2018-05-11
Alternative migration partials

MechaRyRy:
## Problem
We want to be able to expose a mechanism for migrating data from version one to version two of the `download-manager`. At the moment the implementation is heavily constructed around one usecase and buried within a migration job that cannot be customised by clients of the library. 

## Solution
Avoid doing all of the heavy lifting in the library. The idea now is to only go through the `LiteDownloadManagerCommands` interface. This means we will trigger `download` for partials, we won't perform the extraction or any notifications from within the library itself, this will be down to the client. 

This PR publicly exposes the `PartialDownloadMigrationExtractor` (will be moved to demo module later), and shows how the `partial` downloads can be extracted and queued on the new `download-manager`.

## Follow-up
Do the same for the complete downloads from version one. Remove the migration job and wrap everything in something that can be executed off of the main thread.

### DIFF
--- a/demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
+++ b/demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
@@ -1,6 +1,7 @@
 package com.novoda.downloadmanager.demo;
 
 import android.annotation.SuppressLint;
+import android.database.sqlite.SQLiteDatabase;
 import android.os.Bundle;
 import android.support.v4.app.NotificationManagerCompat;
 import android.support.v7.app.AppCompatActivity;
@@ -24,10 +25,14 @@ import com.novoda.downloadmanager.DownloadFileIdCreator;
 import com.novoda.downloadmanager.DownloadMigrator;
 import com.novoda.downloadmanager.DownloadMigratorBuilder;
 import com.novoda.downloadmanager.LiteDownloadManagerCommands;
+import com.novoda.downloadmanager.Migration;
 import com.novoda.downloadmanager.MigrationCallback;
 import com.novoda.downloadmanager.MigrationStatus;
+import com.novoda.downloadmanager.PartialDownloadMigrationExtractor;
+import com.novoda.downloadmanager.SqlDatabaseWrapper;
 
 import java.io.File;
+import java.util.List;
 
 import static com.novoda.downloadmanager.DownloadBatchStatus.Status.ERROR;
 
@@ -124,7 +129,19 @@ public class MainActivity extends AppCompatActivity {
         databaseMigrationUpdates.setText(migrationStatus.status().toRawValue());
     };
 
-    private final View.OnClickListener startMigrationOnClick = v -> downloadMigrator.startMigration("migrationJob", getDatabasePath("downloads.db"), V1_BASE_PATH);
+    private final View.OnClickListener startMigrationOnClick = v -> {
+        File databasePath = getDatabasePath("downloads.db");
+
+        SQLiteDatabase sqLiteDatabase = SQLiteDatabase.openDatabase(databasePath.getAbsolutePath(), null, 0);
+        SqlDatabaseWrapper database = new SqlDatabaseWrapper(sqLiteDatabase);
+        PartialDownloadMigrationExtractor partialDownloadMigrationExtractor = new PartialDownloadMigrationExtractor(database, V1_BASE_PATH);
+
+        List<Migration> migrations = partialDownloadMigrationExtractor.extractMigrations();
+
+        for (Migration migration : migrations) {
+            liteDownloadManagerCommands.download(migration.batch());
+        }
+    };
 
     private final CompoundButton.OnCheckedChangeListener wifiOnlyOnCheckedChange = (buttonView, isChecked) -> {
         LiteDownloadManagerCommands downloadManagerCommands = ((DemoApplication) getApplication()).getLiteDownloadManagerCommands();
--- a/demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
+++ b/demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
@@ -140,8 +140,22 @@ public class MainActivity extends AppCompatActivity {
 
         for (VersionOnePartialDownloadBatch partialDownloadBatch : partialDownloadBatches) {
             liteDownloadManagerCommands.download(partialDownloadBatch.batch());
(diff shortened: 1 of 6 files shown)


library/src/main/java/com/novoda/downloadmanager/Migration.java
```diff
@@ -3,9 +3,9 @@
 import java.util.Collections;
 import java.util.List;
 
-class Migration {
+public class Migration {
```
MechaRyRy, line 6: I was originally going to use `Migration` for both the `complete` and `partial` downloads BUT I found that the `partials` didn't need most of the information that was here, so opted for using a new model.

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
```diff
@@ -58,13 +58,11 @@ public void run() {
...
         FilePersistence filePersistence = filePersistenceCreator.create();
 
-        PartialDownloadMigrationExtractor partialDownloadMigrationExtractor = new PartialDownloadMigrationExtractor(database, basePath);
         MigrationExtractor migrationExtractor = new MigrationExtractor(database, filePersistence, basePath);
-        List<Migration> partialMigrations = partialDownloadMigrationExtractor.extractMigrations();
         List<Migration> completeMigrations = migrationExtractor.extractMigrations();
         DownloadsPersistence downloadsPersistence = RoomDownloadsPersistence.newInstance(context);
 
-        int totalNumberOfMigrations = partialMigrations.size() + completeMigrations.size();
+        int totalNumberOfMigrations = completeMigrations.size();
```
MechaRyRy, line 65: This job will be removed so this code doesn't matter.

library/src/main/java/com/novoda/downloadmanager/MigrationJob.java
```diff
@@ -92,27 +89,6 @@ private void onUpdate(InternalMigrationStatus migrationStatus) {
         }
     }
 
-    private void migratePartialDownloads(InternalMigrationStatus migrationStatus,
```
MechaRyRy, line 95: Mostly exists on the `activity` now.

library/src/main/java/com/novoda/downloadmanager/PartialDownloadMigrationExtractor.java
```diff
@@ -7,54 +7,50 @@
 import java.util.List;
 import java.util.Set;
 
-class PartialDownloadMigrationExtractor {
+public class PartialDownloadMigrationExtractor {
```
MechaRyRy: So that it can be used by the activity (will move module later).

library/src/main/java/com/novoda/downloadmanager/VersionOnePartialDownloadBatch.java
```diff
@@ -0,0 +1,54 @@
+package com.novoda.downloadmanager;
+
+import java.util.List;
+
+public class VersionOnePartialDownloadBatch {
```
MechaRyRy, line 5: `Partials` only needed a subset of the data that a `Migration` captures so created it's own dedicated object.

demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -124,7 +129,33 @@ protected void onCreate(Bundle savedInstanceState) {
         databaseMigrationUpdates.setText(migrationStatus.status().toRawValue());
     };
 
-    private final View.OnClickListener startMigrationOnClick = v -> downloadMigrator.startMigration("migrationJob", getDatabasePath("downloads.db"), V1_BASE_PATH);
+    private final View.OnClickListener startMigrationOnClick = v -> {
```
MechaRyRy, line 132: Where all the magic happens. We extract the migrations from the v1 database and add to the wrapper object `VersionOnePartialDownloadBatch` which we use to perform the `download` and then delete the records from the v1 database. This will be wrapped in a runnable or some other thing later and moved off of the main-thread.

library/src/main/java/com/novoda/downloadmanager/PartialDownloadMigrationExtractor.java
ouchadam: +1 for moving into the demo.

Should this be called a `VersionOnePartialMigrationExtractor`?

library/src/main/java/com/novoda/downloadmanager/PartialDownloadMigrationExtractor.java
```diff
@@ -76,25 +72,14 @@
                             .withIdentifier(DownloadFileIdCreator.createFrom(originalFileId))
                             .apply();
                 }
-
-                FilePath newFilePath = MigrationPathExtractor.extractMigrationPath(basePath, originalFilePath.path(), downloadBatchId);
```
ouchadam, line 80: is this no longer needed because we're reusing the original file?

library/src/main/java/com/novoda/downloadmanager/Migration.java
ouchadam: is the concept of `Migration` still relevant in the library?

library/src/main/java/com/novoda/downloadmanager/PartialDownloadMigrationExtractor.java
MechaRyRy: I'm not even sure why this was here, we would ignore it anyway 😄 it's downloading them from scratch so it only needs the network address.

library/src/main/java/com/novoda/downloadmanager/Migration.java
MechaRyRy: This class will be removed when I do the `complete` downloads.

library/src/main/java/com/novoda/downloadmanager/PartialDownloadMigrationExtractor.java
MechaRyRy: More of a `VersionOnePartialDownloadBatchesExtractor` I guess 😄

ouchadam (approved):
LGMT :100:

danybony (approved):

library/src/main/java/com/novoda/downloadmanager/PartialDownloadMigrationExtractor.java
danybony: I think the original idea was to resume that partial download... 🤷‍♂️

PR #395, opened 2018-06-06
Alternative migration complete

MechaRyRy:
## Problem
We want to be able to expose a mechanism for migrating data from version one to version two of the `download-manager`. At the moment the implementation is heavily constructed around one usecase and buried within a migration job that exists internally inside the library, which cannot be customised by clients of the library. 

## Solution
Avoid doing all of the heavy lifting in the library. The idea now is to only go through the `LiteDownloadManagerCommands` interface. Create a new method on the `LiteDownloadManagerCommands` that is responsible for inserting already complete downloads into the database.

Removed the `Migration` object and replaced it with `VersionOnePartialDownloadBatch` and `CompletedDownloadBatch`. 

## Follow-up
As the `Migration` is now the responsibility of the client, the migration code needs to be moved to the `demo` module rather than existing in the `library`. Moving the migration code will most likely lead to some classes changing their access from package to public, but we should be able to go through the `Creator`s to avoid that happening to a certain extent.

Need to ensure that the `CompletedDownloadBatches` are added to the in-memory map that the `download-manager` uses to perform actions. This will involve checking that we can perform actions on these downloads after a migration is complete.


demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -122,40 +110,7 @@ protected void onCreate(Bundle savedInstanceState) {
...
     };
 
-    private final MigrationCallback migrationCallback = migrationStatus -> {
-        if (migrationStatus.status() == MigrationStatus.Status.COMPLETE) {
-            liteDownloadManagerCommands.submitAllStoredDownloads(() -> Log.d(TAG, "Migration completed, submitting all downloads"));
-        }
-        databaseMigrationUpdates.setText(migrationStatus.status().toRawValue());
-    };
-
-    private final View.OnClickListener startMigrationOnClick = v -> {
```
MechaRyRy, line 132: All of the following has been moved to the `MigrationJob`.

library/src/main/AndroidManifest.xml
```diff
@@ -12,7 +12,6 @@
   <application>
 
     <service android:name="com.novoda.downloadmanager.LiteDownloadService" />
-    <service android:name="com.novoda.downloadmanager.LiteDownloadMigrationService" />
```
MechaRyRy, line 15: 💥 one less service. Now it's the clients responsibility to extract and determine the threading mechanism necessary. We just provide an example through the `MainActivity` of the demo and the `MigrationJob`.

library/src/main/java/com/novoda/downloadmanager/CompletedDownloadBatch.java
```diff
@@ -1,41 +1,38 @@
 package com.novoda.downloadmanager;
 
-import java.util.Collections;
 import java.util.List;
 
-public class Migration {
+public class CompletedDownloadBatch {
```
MechaRyRy, line 5: Only used for `Completed` downloads that we just need to move the files for and insert into the database.

MechaRyRy: Felt better than having a generic `Migration` object where it was capturing information that may not be required for the two different types of migration.

library/src/main/java/com/novoda/downloadmanager/CompletedDownloadBatchesExtractor.java
```diff
@@ -8,7 +8,7 @@
 import java.util.List;
 import java.util.Set;
 
-class MigrationExtractor {
+public class CompletedDownloadBatchesExtractor {
```
MechaRyRy, line 11: As this is now the responsibility of the client application this class will be moved to the `demo` in the next PR.

library/src/main/java/com/novoda/downloadmanager/CompletedDownloadBatchesExtractor.java
```diff
@@ -90,39 +87,36 @@
                             fileIds.add(originalFileId);
                         }
 
-                        if (originalFileId == null) {
-                            newBatchBuilder.downloadFrom(originalNetworkAddress)
-                                    .apply();
-                        } else {
-                            newBatchBuilder.downloadFrom(originalNetworkAddress)
```
MechaRyRy, line 97: Confusing to use the `BatchBuilder` which is used for queueing downloads, instead we have a dedicated object.

library/src/main/java/com/novoda/downloadmanager/CompletedDownloadBatchesExtractor.java
```diff
@@ -90,39 +87,36 @@
...
-                                    .withIdentifier(DownloadFileIdCreator.createFrom(originalFileId))
-                                    .apply();
-                        }
-
                         String rawNewFilePath = new LiteFilePath(sanitizedOriginalUniqueFileLocation).path();
                         FilePath newFilePath = MigrationPathExtractor.extractMigrationPath(basePath, rawNewFilePath, downloadBatchId);
 
-                        FilePath originalFilePath = new LiteFilePath(originalFileLocation);
-                        long rawFileSize = filePersistence.getCurrentSize(originalFilePath);
+                        long rawFileSize = fileSizeExtractor.fileSizeFor(originalFileLocation);
```
MechaRyRy, line 93: `fileSizeExtractor` was created to avoid breaking the tests so that we can mock the dependency.

library/src/main/java/com/novoda/downloadmanager/DownloadMigratorBuilder.java
```diff
@@ -1,181 +0,0 @@
...
-import android.os.Looper;
-import android.support.annotation.DrawableRes;
-import android.support.annotation.RequiresApi;
-import android.support.v4.app.NotificationCompat;
-import android.support.v4.app.NotificationManagerCompat;
-
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-
-public final class DownloadMigratorBuilder {
```
MechaRyRy, line 22: Notifications, Threading and Extraction are now the responsibility of the client application.

library/src/main/java/com/novoda/downloadmanager/FileSizeExtractor.java
```diff
@@ -0,0 +1,11 @@
+package com.novoda.downloadmanager;
+
+import java.io.File;
+
+public class FileSizeExtractor {
```
MechaRyRy, line 5: Created for mocking purposes. Will be moved to the demo in the next PR.

library/src/main/java/com/novoda/downloadmanager/InternalMigrationStatus.java
```diff
@@ -1,17 +0,0 @@
-package com.novoda.downloadmanager;
-
-interface InternalMigrationStatus extends MigrationStatus {
```
MechaRyRy, line 3: Clients have full control of the migration process now except passing to the `download-manager` for storage or download which will go through the normal notification system.

library/src/main/java/com/novoda/downloadmanager/RoomDownloadsPersistence.java
```diff
@@ -139,4 +139,46 @@ public boolean update(DownloadBatchId downloadBatchId, boolean notificationSeen)
         database.roomBatchDao().update(roomBatch);
         return true;
     }
+
+    @Override
+    public void persistCompletedBatch(CompletedDownloadBatch completedDownloadBatch) {
```
MechaRyRy, line 145: All of the above code was originally exposed inside the `MigrationJob` but this meant client applications were unable to specify there own migration process. So now the `Migration` is exposed to clients through the custom object `CompletedDownloadBatch` which is then mapped back to how the library stores download batches.

library/src/main/java/com/novoda/downloadmanager/RoomDownloadsPersistence.java
```diff
@@ -139,4 +139,46 @@ public boolean update(DownloadBatchId downloadBatchId, boolean notificationSeen)
...
+                downloadedDateTimeInMillis,
+                true
+        );
+        persistBatch(persistedBatch);
+
+        for (CompletedDownloadBatch.CompletedDownloadFile completedDownloadFile : completedDownloadBatch.completedDownloadFiles()) {
+            String url = completedDownloadFile.originalNetworkAddress();
+
+            String rawDownloadFileId = rawFileIdFrom(completedDownloadBatch.downloadBatchTitle(), completedDownloadFile);
+            DownloadFileId downloadFileId = DownloadFileIdCreator.createFrom(rawDownloadFileId);
```
danybony, line 164: Should we do a check for duplicates here, throwing in case two files have the same id?

library/src/main/java/com/novoda/downloadmanager/RoomDownloadsPersistence.java
```diff
@@ -139,4 +139,46 @@ public boolean update(DownloadBatchId downloadBatchId, boolean notificationSeen)
...
+        long downloadedDateTimeInMillis = completedDownloadBatch.downloadedDateTimeInMillis();
+
+        DownloadsBatchPersisted persistedBatch = new LiteDownloadsBatchPersisted(
+                downloadBatchTitle,
+                downloadBatchId,
+                downloadBatchStatus,
+                downloadedDateTimeInMillis,
+                true
+        );
+        persistBatch(persistedBatch);
```
danybony: Do we need to check for duplicated batch ids? Maybe it's worth failing fast, maybe exposing an explicit Exception in the method signature, so that clients are aware of avoiding duplicates

danybony (commented):
Looks 💯, I like how this is going, delegating (giving more freedom) to the clients.
I have just a couple of questions/suggestions

library/src/main/java/com/novoda/downloadmanager/RoomDownloadsPersistence.java
```diff
@@ -139,4 +139,46 @@ public boolean update(DownloadBatchId downloadBatchId, boolean notificationSeen)
...
+        DownloadBatchTitle downloadBatchTitle = completedDownloadBatch.downloadBatchTitle();
+        DownloadBatchStatus.Status downloadBatchStatus = DownloadBatchStatus.Status.DOWNLOADED;
+        long downloadedDateTimeInMillis = completedDownloadBatch.downloadedDateTimeInMillis();
+
+        DownloadsBatchPersisted persistedBatch = new LiteDownloadsBatchPersisted(
+                downloadBatchTitle,
+                downloadBatchId,
+                downloadBatchStatus,
+                downloadedDateTimeInMillis,
+                true
```
zegnus: what is `true`?

library/src/main/java/com/novoda/downloadmanager/RoomDownloadsPersistence.java
zegnus: I'd avoid failing here as it looks like very late, could we instead fail at creation time? as, fail-fast

MechaRyRy: Fail when calling `LiteDownloadManagerCommands.addCompletedBatch`? So do a check to see if this identifier exists and then fail? 

I'd leave that for the next PR because I still need to add these to the internal map that is used on the `download-manager`, then I could add this throws.

library/src/main/java/com/novoda/downloadmanager/RoomDownloadsPersistence.java
MechaRyRy: See above comment, it's all related.

library/src/main/java/com/novoda/downloadmanager/RoomDownloadsPersistence.java
MechaRyRy: it's the notification 😂 I hate adding these now because the IDE shows what the param is anyway, it's just GH doesn't. 599d855

library/src/main/java/com/novoda/downloadmanager/RoomDownloadsPersistence.java
danybony: Yeah, I'd fail there, and OK to wait for next PR for me 👍

danybony (approved):

zegnus (approved):

PR #396, opened 2018-06-06
DownloadError message

MechaRyRy:
## Problem
The information provided by the `DownloadError` is a little sparse and as such it is super difficult to debug potential problems in the `download-manager`, when used in a production application. 

## Solution
Provide additional information in the form of a `message` that can be supplied to each of the `DownloadErrors` exposed in the download manager. This message should include information that can help us to diagnose issues.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
@@ -221,6 +221,8 @@ class DownloadBatch {
             if (currentBytesDownloaded > totalBatchSizeBytes) {
                 DownloadError downloadError = DownloadErrorFactory.createSizeMismatchError(downloadFileStatus);
                 downloadBatchStatus.markAsError(Optional.of(downloadError), downloadsBatchPersistence);
+                callbackThrottle.update(downloadBatchStatus);
+                return;
             }
 
             if (currentBytesDownloaded == totalBatchSizeBytes && totalBatchSizeBytes != ZERO_BYTES) {
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
@@ -219,7 +219,13 @@ class DownloadBatch {
             downloadBatchStatus.updateDownloaded(currentBytesDownloaded);
 
             if (currentBytesDownloaded > totalBatchSizeBytes) {
-                DownloadError downloadError = new DownloadError(DownloadError.Type.FILE_CURRENT_AND_TOTAL_SIZE_MISMATCH);
+                String sizeMismatchMessage = String.format(
+                        "Download File with ID: %s has a greater current size: %s than the total size: ",
+                        downloadFileStatus.downloadBatchId().rawId(),
+                        downloadFileStatus.bytesDownloaded(),
+                        downloadFileStatus.totalBytes()
+                );
+                DownloadError downloadError = new DownloadError(DownloadError.Type.FILE_CURRENT_AND_TOTAL_SIZE_MISMATCH, sizeMismatchMessage);
                 downloadBatchStatus.markAsError(Optional.of(downloadError), downloadsBatchPersistence);
             }
 
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
@@ -219,7 +219,7 @@ class DownloadBatch {
             downloadBatchStatus.updateDownloaded(currentBytesDownloaded);
 
             if (currentBytesDownloaded > totalBatchSizeBytes) {
-                DownloadError downloadError = new DownloadError(DownloadError.Error.FILE_CURRENT_AND_TOTAL_SIZE_MISMATCH);
+                DownloadError downloadError = new DownloadError(DownloadError.Type.FILE_CURRENT_AND_TOTAL_SIZE_MISMATCH);
                 downloadBatchStatus.markAsError(Optional.of(downloadError), downloadsBatchPersistence);
             }
 
@@ -256,8 +256,8 @@ class DownloadBatch {
         if (status == WAITING_FOR_NETWORK) {
             return true;
         } else if (status == ERROR) {
-            DownloadError.Error downloadErrorType = downloadBatchStatus.getDownloadErrorType();
-            if (downloadErrorType == DownloadError.Error.NETWORK_ERROR_CANNOT_DOWNLOAD_FILE) {
+            DownloadError.Type downloadErrorType = downloadBatchStatus.getDownloadErrorType();
+            if (downloadErrorType == DownloadError.Type.NETWORK_ERROR_CANNOT_DOWNLOAD_FILE) {
                 return true;
             }
         }
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
@@ -219,13 +219,7 @@ class DownloadBatch {
             downloadBatchStatus.updateDownloaded(currentBytesDownloaded);
 
             if (currentBytesDownloaded > totalBatchSizeBytes) {
-                String sizeMismatchMessage = String.format(
-                        "Download File with ID: %s has a greater current size: %s than the total size: ",
-                        downloadFileStatus.downloadBatchId().rawId(),
-                        downloadFileStatus.bytesDownloaded(),
-                        downloadFileStatus.totalBytes()
-                );
-                DownloadError downloadError = new DownloadError(DownloadError.Type.FILE_CURRENT_AND_TOTAL_SIZE_MISMATCH, sizeMismatchMessage);
+                DownloadError downloadError = DownloadErrorFactory.createSizeMismatchError(downloadFileStatus);
                 downloadBatchStatus.markAsError(Optional.of(downloadError), downloadsBatchPersistence);
             }
 
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
@@ -218,6 +218,11 @@ class DownloadBatch {
             long currentBytesDownloaded = getBytesDownloadedFrom(fileBytesDownloadedMap);
             downloadBatchStatus.updateDownloaded(currentBytesDownloaded);
 
+            if (currentBytesDownloaded > totalBatchSizeBytes) {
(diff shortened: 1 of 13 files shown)


library/src/main/java/com/novoda/downloadmanager/LiteDownloadFileStatus.java
```diff
@@ -98,9 +98,9 @@ public void markAsDeleted() {
     }
 
     @Override
-    public void markAsError(DownloadError.Error error) {
+    public void markAsError(DownloadError downloadError) {
```
suhtai, line 101: May be worth adding `@NonNull` to this method return as well as the parameter (kotlin interop)

library/src/main/java/com/novoda/downloadmanager/DownloadError.java
```diff
@@ -2,7 +2,8 @@
 
 public class DownloadError {
```
suhtai, line 3: Properties could do with `@NonNull` annotations (maybe the constructor too)

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -372,9 +377,9 @@ void persist() {
     void updateTotalSize() {
         if (totalBatchSizeBytes == 0) {
             totalBatchSizeBytes = DownloadBatchSizeCalculator.getTotalSize(
-                downloadFiles,
-                downloadBatchStatus.status(),
-                downloadBatchStatus.getDownloadBatchId()
+                    downloadFiles,
+                    downloadBatchStatus.status(),
+                    downloadBatchStatus.getDownloadBatchId()
```
suhtai, line 384: Extra tabulation? 🤔

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -218,6 +218,11 @@ public void onUpdate(InternalDownloadFileStatus downloadFileStatus) {
             long currentBytesDownloaded = getBytesDownloadedFrom(fileBytesDownloadedMap);
             downloadBatchStatus.updateDownloaded(currentBytesDownloaded);
 
+            if (currentBytesDownloaded > totalBatchSizeBytes) {
+                DownloadError downloadError = DownloadErrorFactory.createSizeMismatchError(downloadFileStatus);
+                downloadBatchStatus.markAsError(Optional.of(downloadError), downloadsBatchPersistence);
+            }
```
suhtai: What are the chances of these `if` clauses happening at the same time? Should it be `if-else`?

library/src/main/java/com/novoda/downloadmanager/DownloadErrorFactory.java
```diff
@@ -0,0 +1,66 @@
...
+        return new DownloadError(DownloadError.Type.STORAGE_UNAVAILABLE, storageUnavailableMessage);
+    }
+
+    static DownloadError createNetworkError(String networkErrorCause) {
+        String networkErrorMessage = String.format(
+                "Network error, cannot download file. Cause: %s",
+                networkErrorCause
+        );
+        return new DownloadError(DownloadError.Type.NETWORK_ERROR_CANNOT_DOWNLOAD_FILE, networkErrorMessage);
+    }
```
suhtai: Probably trivial as errors are not going to be produced in the millions (hopefully) but I thought to mention that format is several less efficient than standard concatenation: https://stackoverflow.com/a/2087703

library/src/test/java/com/novoda/downloadmanager/NetworkFileDownloaderTest.java
```diff
@@ -77,11 +86,13 @@ public void emitsDownloadFinished_whenResponseCodeIsInvalid() throws IOException
 
     @Test
     public void emitsError_whenRequestExecutionFails() throws IOException {
-        given(httpClient.execute(requestCreator.createDownloadRequest(ANY_RAW_URL))).willThrow(IOException.class);
+        IOException ioException = mock(IOException.class);
```
suhtai, line 85: I was gonna say that may not need to use a mock here, but I then wondered if actually, for an exception it's better to use a mock as it then it may not create a stacktrace 😅

library/src/test/java/com/novoda/downloadmanager/NetworkFileDownloaderTest.java
```diff
@@ -63,7 +67,12 @@ public void emitsError_whenResponseCodeIsInvalid() throws IOException {
 
         networkFileDownloader.startDownloading(ANY_RAW_URL, UNKNOWN_FILE_SIZE, callback);
 
-        verify(callback).onError();
+        String expectedCause = String.format(
+                "Request: %s with response code: %s failed.",
+                ANY_RAW_URL,
+                418
+        );
```
suhtai: Probably best to just define a constant string. Otherwise it feels like it's following the code from the test

library/src/main/java/com/novoda/downloadmanager/NetworkFileDownloader.java
```diff
@@ -59,7 +59,12 @@ private void processResponse(Callback callback, HttpClient.NetworkResponse respo
             }
         } else {
             Logger.e("Network response code is not ok, responseCode: " + responseCode);
-            callback.onError();
+            String networkErrorMessage = String.format(
+                    "Request: %s with response code: %s failed.",
+                    url,
+                    responseCode
+            );
```
suhtai, line 66: Same as before, concatenation may be more efficient and potentially easier to read (Shame we can use Kotlin's template, as it works as appending and reads amazing)

library/src/main/java/com/novoda/downloadmanager/DownloadFile.java
```diff
@@ -114,26 +115,22 @@ public void onDownloadFinished() {
...
-                return DownloadError.Error.FILE_CANNOT_BE_WRITTEN;
+                return DownloadErrorFactory.createCannotWriteToFileError(downloadFileStatus);
             default:
                 Logger.e("Status " + status + " missing to be processed");
                 break;
 
         }
-
-        return DownloadError.Error.UNKNOWN;
+        return DownloadErrorFactory.createUnknownError();
```
suhtai: Shouldn't this be in the `default` branch of the switch?

library/src/main/java/com/novoda/downloadmanager/DownloadError.java
suhtai: This applies to more places in this PR

library/src/main/java/com/novoda/downloadmanager/DownloadError.java
```diff
@@ -11,13 +12,24 @@
...
+    DownloadError(Type type) {
+        this.type = type;
+        this.message = "";
+    }
+
+    Type error() {
+        return type;
+    }
+
+    public String message() {
```
suhtai: This one and `error` probably should be `getXXX` to make it a Kotlin property when used with it. Also, should the getter be `type` instead of error? We may need to leave the current methods for convenience or compatibility (Maybe deprecated?)

suhtai (commented):
Just a couple of comments. This would be like a quarter of the lines in Kotlin 😅

library/src/main/java/com/novoda/downloadmanager/DownloadError.java
MechaRyRy: We typically avoid `@NonNull`, assume everything is `@NonNull` unless explicitly labelled with `@Nullable`.

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
MechaRyRy: I think my settings might differ from others on the project, need to add the settings to the repo, will do at some point in the future.

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
MechaRyRy: Sorry I don't understand. What is the chance of this happening at the same time as the `currentBytesDownloaded == totalBatchSizeBytes`? If this one is triggered then the other won't be triggered, right? `> is not ==` I think I might have missed your point.

library/src/test/java/com/novoda/downloadmanager/NetworkFileDownloaderTest.java
MechaRyRy: True 👍

library/src/main/java/com/novoda/downloadmanager/DownloadError.java
MechaRyRy: Yea this should be `type` 👍 I'm not concerned about convenience of compatibility atm, version two hasn't been released. The app we are using this in I need to supply this extra information in anyway, so will be easily be able to update.

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
suhtai: I meant more for the ones after. Like, can it have the condition `currentBytesDownloaded > totalBatchSizeBytes` as well as being marked as an error or waiting for network? Also, it seems unnecessary to do the following checks if this one passes...

library/src/main/java/com/novoda/downloadmanager/DownloadError.java
suhtai: Kotlin I guess ¯\_(ツ)_/¯

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
MechaRyRy: Ah I see. Yea probably most of these could return 🤔 I'll update just this one for now b5e6439

suhtai (approved):

library/src/main/java/com/novoda/downloadmanager/DownloadFile.java
```diff
@@ -114,26 +115,20 @@ public void onDownloadFinished() {
...
-                return DownloadError.Error.STORAGE_UNAVAILABLE;
+                return DownloadErrorFactory.createStorageNotAvailableError(filePersistence);
             case ERROR_OPENING_FILE:
-                return DownloadError.Error.FILE_CANNOT_BE_WRITTEN;
+                return DownloadErrorFactory.createCannotWriteToFileError(downloadFileStatus);
             default:
                 Logger.e("Status " + status + " missing to be processed");
-                break;
-
+                return DownloadErrorFactory.createUnknownError();
```
zegnus: I'd add the unknown status here

MechaRyRy: I don't know what that means 😂 the `Type` is added in the factory.

zegnus (approved):

PR #397, opened 2018-06-07
Return Error not Type

MechaRyRy:
## Problem
We are returning `DownloadError.Type` from `DownloadBatchStatus` instead of `DownloadError` 😬 

## Solution 
Update the interface and all references to use `DownloadError` in place of type.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
@@ -258,8 +258,8 @@ class DownloadBatch {
         if (status == WAITING_FOR_NETWORK) {
             return true;
         } else if (status == ERROR) {
-            DownloadError.Type downloadErrorType = downloadBatchStatus.getDownloadErrorType();
-            if (downloadErrorType == DownloadError.Type.NETWORK_ERROR_CANNOT_DOWNLOAD_FILE) {
+            DownloadError downloadError = downloadBatchStatus.downloadError();
+            if (downloadError != null && downloadError.type() == DownloadError.Type.NETWORK_ERROR_CANNOT_DOWNLOAD_FILE) {
                 return true;
             }
         }
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
@@ -259,9 +259,7 @@ class DownloadBatch {
             return true;
         } else if (status == ERROR) {
             DownloadError downloadError = downloadBatchStatus.downloadError();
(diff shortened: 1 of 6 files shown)


library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -258,8 +258,8 @@ private static boolean networkError(InternalDownloadBatchStatus downloadBatchSta
         if (status == WAITING_FOR_NETWORK) {
             return true;
         } else if (status == ERROR) {
-            DownloadError.Type downloadErrorType = downloadBatchStatus.getDownloadErrorType();
-            if (downloadErrorType == DownloadError.Type.NETWORK_ERROR_CANNOT_DOWNLOAD_FILE) {
+            DownloadError downloadError = downloadBatchStatus.downloadError();
+            if (downloadError != null && downloadError.type() == DownloadError.Type.NETWORK_ERROR_CANNOT_DOWNLOAD_FILE) {
```
lgvalle: It can be inlined to:
`return DownloadError.Type.NETWORK_ERROR_CANNOT_DOWNLOAD_FILE == downloadBatchStatus.downloadError()`

danybony (approved):

library/src/main/java/com/novoda/downloadmanager/LiteDownloadBatchStatus.java
```diff
@@ -161,9 +161,9 @@ private void updateStatusAsync(Status status, DownloadsBatchStatusPersistence pe
 
     @Nullable
     @Override
-    public DownloadError.Type getDownloadErrorType() {
+    public DownloadError downloadError() {
         if (downloadError.isPresent()) {
-            return downloadError.get().type();
+            return downloadError.get();
```
lgvalle, line 166: I know is not the goal of this PR but I thought I'd ask (and perhaps amend in other PR) 
Why don't we return here something like:
 `return downloadError.or(new DownloadError(DownloadError.Type.UNKNOWN));` 
to avoid the `null`?

MechaRyRy: We could do but then you can have a `DOWNLOADING` status with an error type `UNKNOWN`, its valid to return null or optional in this case.

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
MechaRyRy: Very true 😄 ec56a36

library/src/main/java/com/novoda/downloadmanager/LiteDownloadBatchStatus.java
lgvalle: Ok conversation for another PR, let's get this one merged with the fixes it contains already 👍

lgvalle (approved):

PR #399, opened 2018-06-15
Migration binder casting

MechaRyRy:
## Problem
We introduced a guard against casting an invalid binder in #398 but we forgot the migration one 😱 

## Solution
Add guard against an invalid binder in the migration service connection.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadMigratorBuilder.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadMigratorBuilder.java
@@ -93,6 +93,9 @@ public final class DownloadMigratorBuilder {
         ServiceConnection serviceConnection = new ServiceConnection() {
             @Override
             public void onServiceConnected(ComponentName name, IBinder binder) {
+                if (!(binder instanceof LiteDownloadMigrationService.MigrationDownloadServiceBinder)) {
+                    return;
+                }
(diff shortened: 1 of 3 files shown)


zegnus (approved):

library/src/main/java/com/novoda/downloadmanager/DownloadMigratorBuilder.java
```diff
@@ -93,6 +93,9 @@ public DownloadMigrator build() {
         ServiceConnection serviceConnection = new ServiceConnection() {
             @Override
             public void onServiceConnected(ComponentName name, IBinder binder) {
+                if (!(binder instanceof LiteDownloadMigrationService.MigrationDownloadServiceBinder)) {
```
suhtai: Negative check on an early return? 🤔

suhtai: I thought the reason for using early returns was to avoid negative checks

MechaRyRy: I can invert both if you would prefer?

zegnus: @pablisco yup though I tend to favour small if blocks

lgvalle (approved):

PR #402, opened 2018-06-20
V2 alternative migration in memory map

MechaRyRy:
## Problem 
We added a completed batch to the download manager but a client cannot interact with these 😬 because they have not been added to the internal map on the download manager. 

In #395 "Alternative Migration Complete" we wanted to add a `throws` when attempting to add a duplicated completed batch. To do this we need to be able to add the completed batch to the internal map.

## Solution
Add to the internal map and add a throws on the download manager when attempting to add a duplicated completed download batch. 

## QUESTION❓❓ 
This `DownloadManager.addCompletedBatch` feels like it should be a synchronous process that the client should put inside their own threading mechanism. 

1. The `asBatch` could take sometime and ATM would occur on whatever thread  they call it on 😬 
2. It might be useful to return a status and have the client do something based on that status, like how we handle the other parts of the code.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadManager.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadManager.java
@@ -224,14 +224,16 @@ class DownloadManager implements LiteDownloadManagerCommands {
     }
 
     @Override
-    public void addCompletedBatch(CompletedDownloadBatch completedDownloadBatch) {
-        DownloadBatchId downloadBatchId = completedDownloadBatch.downloadBatchId();
-        DownloadBatch downloadBatch = downloadBatchMap.get(downloadBatchId);
-        if (downloadBatch != null) {
-            throw new IllegalStateException("CompletedDownloadBatch with id: " + downloadBatchId + " already exists.");
+    public void addCompletedBatch(CompletedDownloadBatch completedDownloadBatch) throws IllegalArgumentException {
+        if (alreadyContainsBatch(completedDownloadBatch)) {
+            throw new IllegalArgumentException("CompletedDownloadBatch with id: " + completedDownloadBatch.downloadBatchId() + " already exists.");
         }
 
-        downloader.addCompletedBatch(completedDownloadBatch);
+        downloader.addCompletedBatch(completedDownloadBatch, downloadBatchMap);
+    }
+
+    private boolean alreadyContainsBatch(CompletedDownloadBatch completedDownloadBatch) {
+        return downloadBatchMap.get(completedDownloadBatch.downloadBatchId()) != null;
     }
 
 }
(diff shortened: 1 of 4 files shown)


library/src/main/java/com/novoda/downloadmanager/DownloadManager.java
```diff
@@ -224,8 +224,16 @@ public File getDownloadsDir() {
     }
 
     @Override
-    public void addCompletedBatch(CompletedDownloadBatch completedDownloadBatch) {
-        downloader.addCompletedBatch(completedDownloadBatch);
+    public void addCompletedBatch(CompletedDownloadBatch completedDownloadBatch) throws IllegalArgumentException {
```
MechaRyRy, line 227: I think this should be synchronous now 🤔 

The `asBatch` can take sometime to create given that it needs to loop through files. It might be useful to return a status too. What do you think?

MechaRyRy: If we decide to make synchronous then I would merge this and I'll add another method to make the refactoring easier. I need this as a base anyway 😄

library/src/main/java/com/novoda/downloadmanager/DownloadManager.java
```diff
@@ -224,8 +224,16 @@ public File getDownloadsDir() {
...
+        if (alreadyContainsBatch(completedDownloadBatch)) {
+            throw new IllegalArgumentException("CompletedDownloadBatch with id: " + completedDownloadBatch.downloadBatchId() + " already exists.");
+        }
+
+        downloader.addCompletedBatch(completedDownloadBatch, downloadBatchMap);
+    }
+
+    private boolean alreadyContainsBatch(CompletedDownloadBatch completedDownloadBatch) {
+        return downloadBatchMap.get(completedDownloadBatch.downloadBatchId()) != null;
```
danybony: This is the same as `downloadBatchMap.containsKey(completedDownloadBatch.downloadBatchId())`

danybony (commented):
LGTM, to answer your question I don't think the `asBatch` method is gonna be that heavy: you already have all the information at that point, you are just creating new objects to contain these

library/src/main/java/com/novoda/downloadmanager/DownloadManager.java
danybony: A part from the `asBatch` method (which I don't think is that expensive, what other expensive operations we perform here?

library/src/main/java/com/novoda/downloadmanager/DownloadManager.java
MechaRyRy: Very true 😂 memory failed me then 👅

MechaRyRy:
@danybony this implementation suffers in another regard. The client doesn't know if something was successfully added, it's fire and forget. Currently they would:

- Need to wait for it to complete, which is a guess because it executes against an executor. 
- Then check each completed download manually to see if it was indeed added to the download manager 😬 

This means if they wanted to delete from the old DB after adding the item they can't 😬

MechaRyRy:
I've created a PR that uses `WorkerThread` instead, please see the other PR #403

MechaRyRy:
Going with the sync version #403

PR #404, opened 2018-06-20
Feature v2 alternative migration

MechaRyRy:
This is a feature PR into version two of the download-manager. Please look through all of the code, this is a large feature which has been worked on sporadically over the last few months.

## Problem
The current implementation for migrating version one of the download-manager to version two is heavily tied to a single use-case. This migration belongs to a job which is buried in the `download-manager` which is extremely hard to customise. 

## Solution 
Move the migration away from the `download-manager` pushing it towards the client. We now expose a `CompletedDownloadBatch` and `CompletedDownloadFile` which the client can pass to `LiteDownloadManagerCommands.addCompletedBatch`, to add these downloads to the `download-manager`. The client is expected to handle the extraction from version one and the threading.


demo-simple/src/main/java/com/novoda/downloadmanager/demo/MainActivity.java
```diff
@@ -117,14 +111,7 @@ protected void onCreate(Bundle savedInstanceState) {
...
 
-    private final MigrationCallback migrationCallback = migrationStatus -> {
-        if (migrationStatus.status() == MigrationStatus.Status.COMPLETE) {
-            liteDownloadManagerCommands.submitAllStoredDownloads(() -> Log.d(TAG, "Migration completed, submitting all downloads"));
-        }
-        databaseMigrationUpdates.setText(migrationStatus.status().toRawValue());
-    };
-
-    private final View.OnClickListener startMigrationOnClick = v -> downloadMigrator.startMigration("migrationJob", getDatabasePath("downloads.db"), V1_BASE_PATH);
+    private final View.OnClickListener startMigrationOnClick = v -> Executors.newSingleThreadExecutor().submit(migrationJob);
```
danybony: I think both this and `migrationJobCallback` are so small that we could inline them, but on the other hand it makes more clear, for the purpose of the demo, what they are, so good to keep them like this

library/src/main/java/com/novoda/downloadmanager/RoomDownloadsPersistence.java
```diff
@@ -139,4 +140,46 @@ public boolean update(DownloadBatchId downloadBatchId, boolean notificationSeen)
...
+                    url,
+                    FilePersistenceType.INTERNAL
+            );
+            persistFile(persistedFile);
+        }
+    }
+
+    private String rawFileIdFrom(DownloadBatchTitle batch, CompletedDownloadBatch.CompletedDownloadFile completedDownloadFile) {
+        if (completedDownloadFile.fileId() == null || completedDownloadFile.fileId().isEmpty()) {
+            return batch.asString() + System.nanoTime();
```
danybony: Has `nanoTime()` enough precision so that if two sequential batch files don't have a valid fileId, this will give them two distinct ones?

danybony: I'm thinking that this will bite us in the future, when we'll have phones with fast enough CPUs, so that _a lot_ of operations can be done in the same nanosecond.
Worth checking this approach maybe?
https://stackoverflow.com/a/15493695/1184680

danybony (commented):
Looking good, as reviewed into individual PRs.
I just found a point I'd like to discuss

library/src/main/java/com/novoda/downloadmanager/RoomDownloadsPersistence.java
MechaRyRy: The approach you have linked, whilst interesting, I'm not sure can be used in this situation. The problem in the issue is with creating temporary files, we aren't creating files but creating identifiers. If we wanted this to be guaranteed to be unique we could use a combination of `nanosecond` and a hashcode string. The hashcode is unlikely to be the same unless the batch and the files are identical.

MechaRyRy: Perhaps we should be using a `UUID` instead?

danybony: Right, a `UUID` sounds like a more suitable approach 👍

danybony (approved):

Mecharyry:
test this

zegnus (approved):

PR #407, opened 2018-06-20
V2 java doc

MechaRyRy:
## Problem
We are getting closer and closer to performing the major release of `download-manager` v2 and we have very little in the way of javadocs.

## Solution
Begin to add some javadocs.


library/src/main/java/com/novoda/downloadmanager/CallbackThrottle.java
```diff
@@ -1,10 +1,27 @@
 package com.novoda.downloadmanager;
 
+/**
+ * For defining the speed at which {@link DownloadBatchStatus} notifications are emitted.
```
danybony: I prefer `Defines` instead of `For defining`

library/src/main/java/com/novoda/downloadmanager/CallbackThrottle.java
```diff
@@ -1,10 +1,27 @@
...
+    /**
+     * The callback to pass each emission of {@link DownloadBatchStatus} to.
+     *
+     * @param callback to notify.
+     */
     void setCallback(DownloadBatchStatusCallback callback);
 
+    /**
+     * Called internally by the download manager each time a {@link DownloadBatchStatus} changes.
+     * Clients should control the throttling of this status.
```
danybony: `Clients should control the throttling of this status. and sequently notify the callback defined with setCallback()` with a link to the method, I never remember how to do it 😅

library/src/main/java/com/novoda/downloadmanager/CallbackThrottle.java
```diff
@@ -1,10 +1,27 @@
...
+     *
+     * @param downloadBatchStatus to throttle.
+     */
     void update(DownloadBatchStatus downloadBatchStatus);
 
+    /**
+     * Prevent additional {@link DownloadBatchStatus} from being emitted.
```
danybony: This is not true, this is more like "notify the callback that no more DownloadBatchStatus will be emitted". Should we also specify that the implementer of this class might want notify the final callback with the last emitted status?

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerCommands.java
```diff
@@ -8,35 +8,119 @@
 
 public interface LiteDownloadManagerCommands {
```
danybony, line 8: While we are at it, should we rename this?
It still sounds super weird to me... at least removing the `lite` part?

library/src/main/java/com/novoda/downloadmanager/CallbackThrottle.java
```diff
@@ -1,10 +1,27 @@
 package com.novoda.downloadmanager;
 
+/**
+ * For defining the speed at which {@link DownloadBatchStatus} notifications are emitted.
+ */
 public interface CallbackThrottle {
```
danybony: I don't remember, how can the client use this?

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerCommands.java
```diff
@@ -8,35 +8,119 @@
...
+     * Removes a {@link DownloadBatchStatusCallback} from the internal list of callbacks,
+     * removing notifications on {@link DownloadBatchStatus} changes.
+     *
+     * @param downloadBatchCallback to remove from the internal list of callbacks.
+     */
     void removeDownloadBatchCallback(DownloadBatchStatusCallback downloadBatchCallback);
 
+    /**
+     * Retrieves all stored batches from the persistence layer and queues them ready for download,
+     * repopulating the internal map.
```
danybony: Maybe add an example of when this should be called? like after migration completed, or after reboot (if needed)...

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerCommands.java
```diff
@@ -8,35 +8,119 @@
...
+     * Retrieves a list of {@link DownloadBatchStatus} synchronously. Clients should
+     * specify their own Threading mechanism.
+     *
+     * @return a list of {@link DownloadBatchStatus}.
+     */
     @WorkerThread
     List<DownloadBatchStatus> getAllDownloadBatchStatuses();
 
+    /**
+     * Retrieves a list of {@link DownloadBatchStatus} passing to the {@link AllBatchStatusesCallback}.
```
danybony, line 73: `Asynchronously retrieves a list of {@link DownloadBatchStatus} passing it to the {@link AllBatchStatusesCallback}.`

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerCommands.java
```diff
@@ -8,35 +8,119 @@
...
+     * @param downloadBatchId of the {@link DownloadFileStatus} to retrieve.
+     * @param downloadFileId  of the {@link DownloadFileStatus} to retrieve.
+     * @return {@link DownloadFileStatus} or null.
+     */
     @Nullable
     @WorkerThread
     DownloadFileStatus getDownloadFileStatusWithMatching(DownloadBatchId downloadBatchId, DownloadFileId downloadFileId);
 
+    /**
+     * Retrieves {@link DownloadFileStatus} with the associated {@link DownloadBatchId} and {@link DownloadFileId}.
```
danybony, line 93: Asynchronously retrieves

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerCommands.java
```diff
@@ -8,35 +8,119 @@
...
+     *
+     * @param downloadBatchId of the {@link DownloadFileStatus} to retrieve.
+     * @param downloadFileId  of the {@link DownloadFileStatus} to retrieve.
+     * @param callback        to pass the {@link DownloadFileStatus} to.
+     */
     void getDownloadFileStatusWithMatching(DownloadBatchId downloadBatchId, DownloadFileId downloadFileId, DownloadFileStatusCallback callback);
 
+    /**
+     * Changes the {@link ConnectionType} that must be satisfied in order to download.
+     * Will pause all current downloads that are violating this condition.
```
danybony: ...and resume paused ones once the condition is satisfied

library/src/main/java/com/novoda/downloadmanager/NotificationCustomizer.java
```diff
@@ -7,6 +7,13 @@
 
     NotificationDisplayState notificationDisplayState(T payload);
```
danybony: Specify that this should return the display state (with a link to the class) for the given payload, so that the user can decide if to show/hide the notification.
Also, missing javadoc for that class

library/src/main/java/com/novoda/downloadmanager/NotificationCustomizer.java
```diff
@@ -7,6 +7,13 @@
```
danybony, line 14: This class should have a description and link to `DownloadManagerBuilder` method using it

library/src/main/java/com/novoda/downloadmanager/CallbackThrottle.java
danybony: Ah yes, this class should have a link to `DownloadManagerBuilder` method using it

danybony (changes requested):

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerCommands.java
MechaRyRy: I would avoid saying that something does something asynchronously. We have the `WorkerThread` annotation to show when something is done synchronously, seems overkill to then say the same for async. What do you think @zegnus ?

library/src/main/java/com/novoda/downloadmanager/CallbackThrottle.java
MechaRyRy: Good idea! 💯

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerCommands.java
MechaRyRy: Yea I might do that actually, but it might be better left until the migration stuff is merged, this will have lots of conflicts probably 😄

library/src/main/java/com/novoda/downloadmanager/CallbackThrottle.java
MechaRyRy: I don't follow. This method prevents additional statuses from being emitted. Either with a guard statement or by removing a listener. It doesn't notify any callback to stop doing something.

library/src/main/java/com/novoda/downloadmanager/CallbackThrottle.java
MechaRyRy: Good idea 😄

library/src/main/java/com/novoda/downloadmanager/CallbackThrottle.java
MechaRyRy: 👍

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerCommands.java
danybony: Agreed 👍

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerCommands.java
MechaRyRy: I was confused about this because I couldn't see where it resumes once the condition is satisfied. It doesn't seem to schedule a job or anything to check.

library/src/main/java/com/novoda/downloadmanager/CallbackThrottle.java
danybony: What I meant is that given how it's phrased like now, as a client I'm implementing this interface with my concrete implementation. This method is gonna then be called to tell my implementation that no more batch statuses are gonna be emitted. It's a callback, not a method that actually prevents the emission (that I need to implemented myself). I don't know if it's clearer now

MechaRyRy: Oh I see what you mean. It's a method that is called by the `download-manager` to prevent the statuses from being emitted, as opposed to something that the client can call to stop them 👍 makes sense 😄

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerCommands.java
MechaRyRy: f0bbd450ec7e8c630f0006fce1a4ebf67612762b

library/src/main/AndroidManifest.xml
```diff
@@ -1,15 +1,19 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
+  xmlns:tools="http://schemas.android.com/tools"
   package="com.novoda.downloadmanager">
 
   <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
   <uses-permission android:name="android.permission.INTERNET" />
   <uses-permission android:name="android.permission.WAKE_LOCK" />
+  <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
```
zegnus, line 8: why is this part of a javadoc PR?

demo-simple/src/main/res/xml/network_security_config.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<network-security-config>
+  <base-config cleartextTrafficPermitted="true" />
```
MechaRyRy, line 3: Starting with Android 9.0 (API level 28), cleartext support is disabled by default. You need to specify this file or do a hack on the application in the manifest to workaround it
The idea being that the app should specify what endpoints it can interact with rather than just connecting to anything :joy:
I specify `true` so that it behaves as it did prior to API 28 and have added it to the demo to show how people can override it

library/src/main/java/com/novoda/downloadmanager/DownloadBatchStatus.java
```diff
@@ -33,27 +39,61 @@ public static Status from(String rawValue) {
 
     }
 
+    /**
+     * @return The title associated to a {@link DownloadBatch}.
+     * Specified when calling {@link Batch#with(StorageRoot, DownloadBatchId, String)}.
```
zegnus, line 44: Maybe we can add that this title is going to be used by the default notifications

library/src/main/java/com/novoda/downloadmanager/DownloadFileStatus.java
```diff
@@ -12,15 +18,33 @@
...
+     */
     long bytesDownloaded();
 
+    /**
+     * @return The total number of bytes to download.
+     */
     long totalBytes();
 
+    /**
+     * @return The local path to this download file.
```
zegnus, line 42: Worth adding an example of the format?

library/src/main/java/com/novoda/downloadmanager/FilePath.java
```diff
@@ -1,5 +1,8 @@
 package com.novoda.downloadmanager;
 
+/**
+ * Represents a path on the system to the asset.
```
zegnus: same as before, it would be interesting to know if this returns a relative or an absolute path for example

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerCommands.java
```diff
@@ -7,33 +7,113 @@
...
+     * Removes a {@link DownloadBatchStatusCallback} from the internal list of callbacks,
+     * removing notifications on {@link DownloadBatchStatus} changes.
+     *
+     * @param downloadBatchCallback to remove from the internal list of callbacks.
+     */
     void removeDownloadBatchCallback(DownloadBatchStatusCallback downloadBatchCallback);
 
+    /**
+     * Retrieves all stored batches from the persistence layer and queues them ready for download,
+     * repopulating the internal map. Normally called when first starting the app, after a migration or restart.
```
zegnus: >repopulating the internal map
might not be necessary to say

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerCommands.java
zegnus: Agree with @Mecharyry, by default calls are non-blocking and if the param is a callback I would understand that asynchronicity. Blocking calls though should be noted in the Javadoc.

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerCommands.java
```diff
@@ -7,33 +7,113 @@
...
+     * Will pause all current downloads that are violating this condition, resuming
+     * all those that satisfy the condition.
+     *
+     * @param allowedConnectionType that must be satisfied in order to download.
+     */
     void updateAllowedConnectionType(ConnectionType allowedConnectionType);
 
+    /**
+     * Adds a {@link CompletedDownloadBatch} to the download manager.
+     * This call is synchronous, clients should specify their own Threading mechanism.
```
zegnus: I would add something like, This is a long-running blocking operation, clients should call this method on their own Threading mechanism. Same for everywhere we use synchronous.

I'm mentioning this because that a method is synchronous is not a problem, the majority are. The problem is that this is a long running blocking operation.

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerCommands.java
```diff
@@ -7,33 +7,113 @@
...
+     *
+     * @param completedDownloadBatch to add to the download manager.
```
zegnus: when might this method be useful for a client?

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerCommands.java
danybony: I still think that we should add the Synchronously/Asynchronously words when there are two variations of the same method

danybony (commented):
Just one comment about sync/async to be added to some methods description, up to @zegnus

danybony (approved):

library/src/main/AndroidManifest.xml
MechaRyRy: 🤔 not sure. I think I updated to API 28 when the build was failing 😬

MechaRyRy:
@zegnus I've been through and addressed all of your comments.

zegnus (approved):

PR #408, opened 2018-06-21
storage requirement

MechaRyRy:
## Problem

Our percentage based storage requirement rule is a little excessive :joy: If a device has 32GB of memory then they cannot download a file when they have 3.2GB of space remaining on their device :grimacing:

## Solution
Implement an interface for `StorageRequirementRule` so that clients can specify their own rules for when a write / create of a file should fail. Create a wrapper `StorageRequirmentRules` that holds one or more rules to be checked. This is exposed through the `DownloadManagerBuilder..withStorageRequirementRules(StorageRequirementRule...)`. 

:warning: we do not add a rule by default :warning:

### DIFF
--- a/demo-simple/src/main/java/com/novoda/downloadmanager/demo/DemoApplication.java
+++ b/demo-simple/src/main/java/com/novoda/downloadmanager/demo/DemoApplication.java
@@ -5,12 +5,13 @@ import android.os.Handler;
 import android.os.Looper;
 
 import com.facebook.stetho.Stetho;
+import com.novoda.downloadmanager.ByteBasedStorageRequirementRule;
 import com.novoda.downloadmanager.DownloadManagerBuilder;
 import com.novoda.downloadmanager.LiteDownloadManagerCommands;
-import com.novoda.downloadmanager.PercentageBasedStorageRequirementRule;
 
 public class DemoApplication extends Application {
 
+    private static final int TWO_HUNDRED_MB_IN_BYTES = 200000000;
     private volatile LiteDownloadManagerCommands liteDownloadManagerCommands;
 
     @Override
@@ -26,7 +27,7 @@ public class DemoApplication extends Application {
         liteDownloadManagerCommands = DownloadManagerBuilder
                 .newInstance(this, handler, R.mipmap.ic_launcher_round)
                 .withLogHandle(new DemoLogHandle())
-                .withStorageRequirementRules(PercentageBasedStorageRequirementRule.withPercentageOfStorageRemaining(0.2f))
+                .withStorageRequirementRules(ByteBasedStorageRequirementRule.withBytesOfStorageRemaining(TWO_HUNDRED_MB_IN_BYTES))
                 .build();
     }
 
--- a/demo-simple/src/main/java/com/novoda/downloadmanager/demo/DemoApplication.java
+++ b/demo-simple/src/main/java/com/novoda/downloadmanager/demo/DemoApplication.java
@@ -7,6 +7,7 @@ import android.os.Looper;
 import com.facebook.stetho.Stetho;
 import com.novoda.downloadmanager.DownloadManagerBuilder;
 import com.novoda.downloadmanager.LiteDownloadManagerCommands;
+import com.novoda.downloadmanager.PercentageBasedStorageRequirementRule;
 
 public class DemoApplication extends Application {
 
@@ -25,6 +26,7 @@ public class DemoApplication extends Application {
         liteDownloadManagerCommands = DownloadManagerBuilder
                 .newInstance(this, handler, R.mipmap.ic_launcher_round)
                 .withLogHandle(new DemoLogHandle())
+                .withStorageRequirementRules(PercentageBasedStorageRequirementRule.withPercentageOfStorageRemaining(0.2f))
                 .build();
     }
 
--- a/demo-simple/src/main/java/com/novoda/downloadmanager/demo/DemoApplication.java
+++ b/demo-simple/src/main/java/com/novoda/downloadmanager/demo/DemoApplication.java
@@ -11,7 +11,7 @@ import com.novoda.downloadmanager.StorageRequirementRuleFactory;
 
 public class DemoApplication extends Application {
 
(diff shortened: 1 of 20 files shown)


library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
```diff
@@ -60,12 +59,12 @@
...
     private long frequency;
     private Optional<LogHandle> logHandle;
-    private StorageRequirementsRule storageRequirementsRule;
+    private StorageRequirementRules storageRequirementRules;
 
     public static DownloadManagerBuilder newInstance(Context context, Handler callbackHandler, @DrawableRes final int notificationIcon) {
         Context applicationContext = context.getApplicationContext();
 
-        StorageRequirementsRule storageRequirementsRule = StorageRequirementsRule.withPercentageOfStorageRemaining(TEN_PERCENT);
+        StorageRequirementRules storageRequirementRule = StorageRequirementRules.newInstance();
```
MechaRyRy, line 66: We do not add a rule by default, just create the empty list. It is the responsibility of the client to add a rule if they feel one is necessary.

library/src/main/java/com/novoda/downloadmanager/FileSize.java
```diff
@@ -6,6 +6,8 @@
 
     long totalSize();
 
+    long remainingSize();
```
MechaRyRy, line 9: Seemed prudent to also have a method for remaining size.

library/src/main/java/com/novoda/downloadmanager/PercentageBasedStorageRequirementRule.java
```diff
@@ -1,32 +1,31 @@
...
-        StatFs statFs = new StatFs(storageDirectory.getPath());
-        long storageCapacityInBytes = storageCapacityReader.storageCapacityInBytes(statFs);
+    @Override
+    public boolean hasViolatedRule(File storageDirectory,
+                                   FileSize downloadFileSize) {
+        long storageCapacityInBytes = storageCapacityReader.storageCapacityInBytes(storageDirectory.getPath());
         long minimumStorageRequiredInBytes = (long) (storageCapacityInBytes * percentageOfStorageRemaining);
         long usableStorageInBytes = storageDirectory.getUsableSpace();
-        long remainingStorageAfterDownloadInBytes = usableStorageInBytes - downloadFileSize.totalSize();
+        long remainingStorageAfterDownloadInBytes = usableStorageInBytes - downloadFileSize.remainingSize();
```
MechaRyRy, line 24: It was using the `totalSize` to determine whether it could continue to download :man_facepalming: this is constant! It should be using the remaining size of the file :smile:

library/src/main/java/com/novoda/downloadmanager/StorageCapacityReader.java
```diff
@@ -5,7 +5,8 @@
 
 class StorageCapacityReader {
 
-    long storageCapacityInBytes(StatFs statFs) {
+    long storageCapacityInBytes(String path) {
+        StatFs statFs = new StatFs(path);
```
MechaRyRy, line 9: Takes `String` as a param so that this class can be mocked :smile:

library/src/main/java/com/novoda/downloadmanager/StorageRequirementRules.java
```diff
@@ -0,0 +1,33 @@
+package com.novoda.downloadmanager;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
+
+final class StorageRequirementRules implements StorageRequirementRule {
```
MechaRyRy, line 7: Wrapper to hold multiple rules.

library/src/main/java/com/novoda/downloadmanager/ByteBasedStorageRequirementRule.java
```diff
@@ -0,0 +1,32 @@
+package com.novoda.downloadmanager;
+
+import java.io.File;
+
+public final class ByteBasedStorageRequirementRule implements StorageRequirementRule {
+
+    private final StorageCapacityReader storageCapacityReader;
+    private final long bytesRemainingAfterDownload;
+
+    public static ByteBasedStorageRequirementRule withBytesOfStorageRemaining(long bytesRemainingAfterDownload) {
```
zegnus: I'd hide this public constructor and the class itself, then from the usage API point of view you can create a unique `StorageRequirementRuleFactory` that returns an implementation of `StorageRequirementRule`.

This has two advantatges:
- All the internal implementation of the download manager keeps being private to the library
- Only the interface is public to the client
- The user has an intuitive way to find out if we provide implementations for the required interface at `. withStorageRequirementRules`

MechaRyRy: fc19d23

zegnus (approved):

demo-simple/src/main/java/com/novoda/downloadmanager/demo/DemoApplication.java
```diff
@@ -7,9 +7,11 @@
 import com.facebook.stetho.Stetho;
 import com.novoda.downloadmanager.DownloadManagerBuilder;
 import com.novoda.downloadmanager.LiteDownloadManagerCommands;
+import com.novoda.downloadmanager.StorageRequirementRuleFactory;
 
 public class DemoApplication extends Application {
 
+    private static final int TWO_HUNDRED_MB_IN_BYTES = 200000000;
```
joetimmins: shouldn't this be 209,715,200?

MechaRyRy: oh very true 😂 I was using decimal version to make it easier in the tests but shouldn't be this here 😂

joetimmins (commented):
The logic all seems fine, but are you sure the numbers are correct? shouldn't they be powers of 2?

MechaRyRy:
@joetimmins sorry, I was using decimal for the tests and forgot to change back for the main application 😂

joetimmins (approved):

PR #409, opened 2018-06-21
Use UUID in place of system.nanoTime.

MechaRyRy:
## Problem
We currently use `System.nanoTime` when creating a `rawFileId` for a given batch. 😬 There is no guarantee that this time will be unique each time this method is called.

## Solution
Use a `UUID` instead.


danybony (approved):
Beautiful 💯

PR #410, opened 2018-06-26
CallbackThrottle & delete on error

MechaRyRy:
## Problem
- The `CallbackThrottleByProgressIncrease` emits `DownloadBatchStatus` updates on progress change updates. It would also emit two instances of `ERROR` 😬 because we do not check for status changes or `DownloadError.type`.

- If receiving an error such as `insufficient storage` then it is impossible to delete that batch 😱 it seems that the `delete` is dependent on the batch still running in some way 😬 

## Solution
- Make the `CallbackThrottleByProgressIncrease` emit `DownloadBatchStatus` for only unique values.

- When performing a delete and the batch is in an `ERROR` state, immediately delete the asset, rather than waiting for the next tick of the `download-manager`.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/CallbackThrottleByProgressIncrease.java
+++ b/library/src/main/java/com/novoda/downloadmanager/CallbackThrottleByProgressIncrease.java
@@ -33,7 +33,7 @@ class CallbackThrottleByProgressIncrease implements CallbackThrottle {
 
     private boolean statusHasChanged(DownloadBatchStatus currentDownloadBatchStatus) {
         DownloadBatchStatus.Status newStatus = currentDownloadBatchStatus.status();
-        return currentStatus != newStatus;
+        return !currentStatus.equals(newStatus);
     }
 
     private boolean progressHasChanged(DownloadBatchStatus currentDownloadBatchStatus) {
@@ -43,7 +43,7 @@ class CallbackThrottleByProgressIncrease implements CallbackThrottle {
 
     private boolean errorHasChanged(DownloadBatchStatus currentDownloadBatchStatus) {
         DownloadError newDownloadError = currentDownloadBatchStatus.downloadError();
-        return currentDownloadError != newDownloadError;
+        return !currentDownloadError.equals(newDownloadError);
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


zegnus (approved):
we might also want to do that in the other callback by time

library/src/main/java/com/novoda/downloadmanager/CallbackThrottleByProgressIncrease.java
```diff
@@ -3,40 +3,51 @@
...
-        }
+    private boolean errorHasChanged(DownloadBatchStatus currentDownloadBatchStatus) {
+        DownloadError newDownloadError = currentDownloadBatchStatus.downloadError();
+        return newDownloadError != null && !newDownloadError.equals(currentDownloadError);
     }
 
     @Override
     public void stopUpdates() {
-        callback.onUpdate(downloadBatchStatus);
+        callback = null;
```
danybony, line 51: Are you sure about this? are we not risking that the last status is not emitted?

MechaRyRy: I ran through the app and this doesn't seem to cause any issues, it was strange that we were saying to stop updates and yet it was spitting out an additional anyway.

danybony: Ok, here's a scenario that illustrates why it's useful to send another update:
- file is 100 MB
- the previous update forwarded to the client was at 99.000 KBytes downloaded (99%)
- we download 999 kBytes, but given that the percentage didn't change (99%) we don't send a new update to the client
- download stops (for any reason, say network issue)
- `stopUpdates()` is invoked
- the client will never know that we actually download almost another 1MB (1%) of the file

Now the numbers might not be entirely allright in my example, but I hope I managed to explain why we need to send one last update here (maybe with some additional logic, about the downloaded data?)

MechaRyRy: The important part from this is `download stops (for any reason, say network issue)` this will be notified because there was a state change, then the `stopUpdates` is invoked.

danybony: Mmm good point 👍  That should cover all the edge cases where before we needed to send this extra update

danybony (approved):

PR #411, opened 2018-06-26
Path based persistence

MechaRyRy:
## Problem
We currently implement two `FilePersistence` classes that are very similar, `InternalFilePersistence` and `ExternalFilePersistence`. `basePath()` is the only "real" difference between the two, we could easily create a generic `FilePersistence` that operates using only an `absolutePath`. 

There are a couple of considerations that need to be made for this `PathBasedFilePersistence`.
- The `PathBasedFilePersistence` needs to be created when loading files from the database 😬 this means we can't keep the `FilePersistence.basePath` because it could potentially change value on each run of the application (switching between primary and secondary storage).
- If we remove the `FilePersistence.basePath` where do we add this now? It's prepended to the  `BatchFile.path` to create the `absolutePath` when hitting the `FilePersistence`.
- If we need to add this `basePath` somewhere else it needs to be easy enough to use for the client applications without making them specify it for every file.

## Solution
- Create a `StorageRoot` interface which represents the `basePath` to download a file to. Provide a primary storage root which would expose a base path similar to: `/data/user/0/com.novoda.downloadmanager.demo.simple/files`.

- Clients supply this `StorageRoot` to the `Batch` through `Batch.with(StorageRoot storageRoot, DownloadBatchId downloadBatchId, String title)`.

- Remove the `FilePersistence.basePath` and use the absolute path that is supplied by each `BatchFile`. 

#### Batch Table 

Before | screen shot 2018-06-26 at 17 02 01
--- | ---
**After** | screen shot 2018-06-26 at 16 39 18

#### File Table
Before | screen shot 2018-06-26 at 17 03 09
--- | ---
**After** | screen shot 2018-06-26 at 16 42 19

#### Batch Table 

Before | 
screen shot 2018-06-26 at 17 05 54
--- | ---
**After** | screen shot 2018-06-26 at 16 34 21

#### File Table
Before | screen shot 2018-06-26 at 17 06 34
--- | ---
**After** | screen shot 2018-06-26 at 16 34 11


zegnus (commented):
This `StorageRootFactory` is not used at all in the demo, right? If this is something that's available, showcase it ;)

MechaRyRy:
I'm happy to use the `Factory` if you can accept that the `/downloads/` part of the path will now be missing from the demo.

library/src/main/java/com/novoda/downloadmanager/DownloadErrorFactory.java
```diff
@@ -41,8 +41,9 @@ static DownloadError createCannotWriteToFileError(DownloadFileStatus downloadFil
         return new DownloadError(DownloadError.Type.FILE_CANNOT_BE_WRITTEN, cannotWriteToFileMessage);
     }
 
-    static DownloadError createStorageNotAvailableError(FilePersistence filePersistence) {
-        String storageUnavailableMessage = "Storage with base path: " + filePersistence.basePath() + " is not available";
+    static DownloadError createStorageNotAvailableError(FilePath filePath) {
+        // QUESTION: is it enough to swap out the basePath with the filePath?
```
zegnus: yeah I guess, can we test this?

library/src/main/java/com/novoda/downloadmanager/PrimaryStorageRoot.java
```diff
@@ -0,0 +1,18 @@
...
+
+    private final Context context;
+
+    PrimaryStorageRoot(Context context) {
+        this.context = context;
+    }
+
+    @Override
+    public String path() {
+        return context.getApplicationContext().getFilesDir().getAbsolutePath();
```
zegnus: we won't use `private static final String DOWNLOADS_DIR = "/downloads/";` anymore?

zegnus (commented):
minor comments

library/src/main/java/com/novoda/downloadmanager/DownloadErrorFactory.java
MechaRyRy: Ah! I just realised that this isn't actually exposed in the `download-manager` anymore 🤔 We are using absolute paths and not checking if the storage is mounted so we'll see an error when attempting to write instead. Is it ok to remove this @zegnus ?

library/src/main/java/com/novoda/downloadmanager/PrimaryStorageRoot.java
MechaRyRy: Well I wasn't sure about this tbh. I figured we should just use the internal directory `files` if the client wants to add to the root then they can by supplying their own `StorageRoot`.

MechaRyRy:
@zegnus I've removed the `UNAVAILABLE_STORAGE` error and used the `downloads directory` on the main activity.

library/src/main/java/com/novoda/downloadmanager/DownloadErrorFactory.java
zegnus: ok 👍

library/src/main/java/com/novoda/downloadmanager/PrimaryStorageRoot.java
zegnus: Yeah I know but this is our implementation, just to align it with what we had before I'd keep the `downloads`. The client can provide whatever then in their own implementation

zegnus (approved):

library/src/main/java/com/novoda/downloadmanager/PrimaryStorageDownloadsDirectoryRoot.java
```diff
@@ -0,0 +1,20 @@
...
+    PrimaryStorageDownloadsDirectoryRoot(Context context) {
+        this.context = context;
+    }
+
+    @Override
+    public String path() {
+        return context.getApplicationContext().getFilesDir().getAbsolutePath() + DOWNLOADS_DIR;
+    }
+}
+
```
jszmltr: Extra line

library/src/test/java/com/novoda/downloadmanager/MigrationExtractorTest.java
```diff
@@ -48,21 +48,21 @@
...
     private final SqlDatabaseWrapper database = mock(SqlDatabaseWrapper.class);
-    private final InternalFilePersistence internalFilePersistence = mock(InternalFilePersistence.class);
+    private final FilePersistence filePersistence = mock(FilePersistence.class);
 
     private MigrationExtractor migrationExtractor;
 
     @Before
     public void setUp() {
-        migrationExtractor = new MigrationExtractor(database, internalFilePersistence, "base");
+        migrationExtractor = new MigrationExtractor(TestStorageRootFactory.create(), database, filePersistence, "base");
```
jszmltr: seems like we use `TestStorageRootFactory.create()` in quite few places. Should we extract it to be final field ?

jszmltr (commented):
Two teeny-tiny comments.

PR #413, opened 2018-06-29
Concurrent modification exception

zegnus:
**Problem**

A client has reported `concurrent modification exceptions` on the `.add` and `.remove` callbacks

**Solution**

Change the data structure of the callbacks from the current `ArrayList` to `https://developer.android.com/reference/java/util/concurrent/CopyOnWriteArraySet` as the latter is thread-safe and fail-safe

Official documentation: https://developer.android.com/reference/java/util/concurrent/CopyOnWriteArraySet

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadManager.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadManager.java
@@ -7,6 +7,7 @@ import android.support.annotation.WorkerThread;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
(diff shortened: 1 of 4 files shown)


MechaRyRy:
@zegnus were you able to reproduce the issue and prove that this fixes it? We had a similar issue before and switching these out made very little difference 😄

zegnus:
@Mecharyry I wasn't able to reproduce the issue using the application but I did a small java app with some concurrent threads adding, removing and looping through elements and changing the data structure fixed the issue...

MechaRyRy (approved):

danybony (approved):

PR #414, opened 2018-07-03
Multiple error events

MechaRyRy:
## Problem
The #410 Callback throttle and delete on error PR spoke about how the `download-manager` is emitting multiple times for some instances of `DownloadBatchStatus`. After checking into the code it seems that we only throttle callbacks that come from `DownloadFile.Callback`, this means we can emit duplicate events from anywhere else in the `DownloadBatch`.

## Solution
Have `DownloadBatch` only emit `DownloadBatchStatus` through the `CallbackThrottle`. Also add some logs to the default `CallbackThrottle` for debugging purposes.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
@@ -30,7 +30,6 @@ class DownloadBatch {
     private final ConnectionChecker connectionChecker;
 
     private long totalBatchSizeBytes;
-    private DownloadBatchStatusCallback callback;
 
     DownloadBatch(InternalDownloadBatchStatus internalDownloadBatchStatus,
                   List<DownloadFile> downloadFiles,
@@ -47,7 +46,6 @@ class DownloadBatch {
     }
 
     void setCallback(DownloadBatchStatusCallback callback) {
-        this.callback = callback;
         callbackThrottle.setCallback(callback);
     }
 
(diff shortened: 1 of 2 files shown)


library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -47,51 +39,49 @@
...
         }
 
         if (networkError(downloadBatchStatus)) {
-            processNetworkError(downloadBatchStatus, callback, downloadsBatchPersistence);
+            processNetworkError(downloadBatchStatus, callbackThrottle, downloadsBatchPersistence);
         }
 
-        deleteBatchIfNeeded(downloadBatchStatus, downloadsBatchPersistence, callback);
-        notifyCallback(callback, downloadBatchStatus);
-        callbackThrottle.stopUpdates();
```
MechaRyRy, line 89: ⚠️ ❓ I don't think this method if required anymore on the `CallbackThrottle`. The `CallbackThrottle` will always be able to callback to the `Downloader` for a given `Batch`. If the batch is removed then it is removed from the map, when it is removed from the map then this throttle will be eligible for GC and will not exist further. Let me know what you think please 😄 

I haven't removed it because this is a discussion before I remove a potentially controversial piece of code.

zegnus: mmm for every single download there is a new callback being created, this also shuts down any call inside the implementation of the callback that should be called when that download finishes, check for example `callbackThrottleByTime.stopUpdates`. I think we have to call it

MechaRyRy:
Opting for #415

PR #415, opened 2018-07-03
Multiple events

MechaRyRy:
## Problem
After some discussion with @zegnus we would rather not mess with the callback flow that currently exists on the `DownloadBatch`.

## Solution
- We've opted with introducing a variation of the filter that was used on the `CallbackThrottleByProgressIncrease` for `DownloadBatchStatus` to the `LiteDownloadManagerDownloader`. When the `LiteDownloadManagerDownloader` receives a `DownloadBatchStatus` change, it will only allow it to be emitted to the clients if it wasn't previously emitted. 

- The `CallbackThrottleByProgressIncrease` has been reverted to its previous state and now only handles `progress increases` again.


zegnus:
Let's go with this for now, it might require another thought when we have some more time

zegnus (approved):

juankysoriano (approved):

PR #416, opened 2018-07-04
Callback throttle rename

MechaRyRy:
## Problem
When working on #415 I found the `CallbackThrottle` a little misleading. From a clients point of view it implies, to me, that it throttles all of the callbacks in the library that are emitted to the client but this is not the case. This `CallbackThrottle` handles the emissions from the `DownloadFile` to the `DownloadBatch`, essentially the emissions that tick when a file is downloading.

## Solution
Rename the interface to show that it is explicitly for the file and not for anything else. Add a javadoc to the interface 😄

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/FileCallbackThrottle.java
+++ b/library/src/main/java/com/novoda/downloadmanager/FileCallbackThrottle.java
@@ -1,6 +1,6 @@
 package com.novoda.downloadmanager;
 
-public interface CallbackThrottle {
+public interface FileCallbackThrottle {
 
     void setCallback(DownloadBatchStatusCallback callback);
 
--- a/library/src/main/java/com/novoda/downloadmanager/FileCallbackThrottle.java
+++ b/library/src/main/java/com/novoda/downloadmanager/FileCallbackThrottle.java
@@ -1,10 +1,30 @@
 package com.novoda.downloadmanager;
 
+/**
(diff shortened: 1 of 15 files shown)


juankysoriano (approved):

zegnus (approved):

library/src/main/java/com/novoda/downloadmanager/FileCallbackThrottle.java
```diff
@@ -0,0 +1,32 @@
+package com.novoda.downloadmanager;
+
+/**
+ * Controls the rate at which a {@link DownloadFile} calls back to
+ * a {@link DownloadBatch}. Essentially controls the rate of notifications
+ * during a download of a file.
+ * <p>
+ * Note: This does not control the overall emissions of {@link DownloadBatchStatus} between the client and this library.
+ */
+public interface FileCallbackThrottle {
```
danybony, line 10: Sorry I'm late I just managed to check this PR, but what about adding a link in this class' Javadoc to the method the client can use to specify a custom Throttle?

MechaRyRy: Sure thing, here you go 😄 https://github.com/novoda/download-manager/blob/v2_java_doc/library/src/main/java/com/novoda/downloadmanager/FileCallbackThrottle.java#L6

PR #417, opened 2018-07-09
Demo add storage root selection

MechaRyRy:
## Problem
I need to add some more views when re-queuing downloads but the current `MainActivity` and the associated views are a little messy, they need some TLC.

## Solution
Split the `Migration` and the `Downloads` into two activities that can be navigated to from a `LandingActivity` splitting some of the logic present in the `MainActivity`.

Introduce a `DownloadBatchStatusView` to encapsulate the commands that can be executed against a given batch.

## Screen Capture

Landing | Migration | Downloads
--- | --- | ---
landing | migration | downloads

## Follow-up 
Add a view component for changing between storage locations and a `re-queue` button to each download batch.


demo-simple/src/main/res/layout/merge_download_batch_status.xml
```diff
@@ -0,0 +1,47 @@
...
+
+  <TextView
+    android:id="@+id/download_batch_status"
+    android:layout_width="match_parent"
+    android:layout_height="0dp"
+    android:layout_marginEnd="8dp"
+    android:layout_marginLeft="8dp"
+    android:layout_marginRight="8dp"
+    android:layout_marginStart="8dp"
+    android:layout_marginTop="8dp"
```
danybony, line 17: These margins are always the same... worth extracting these (8 and 16) to res? 
Not a blocking eh, it's good already for the demo

danybony (approved):
Nice one, `MainActivity` really needed some love (and to be more focused).
Just one idea to make the layout more clear: what if we put each `DownloadBatchStatusView` inside a `CardView`, to better group each batch operations?

demo-simple/src/main/java/com/novoda/downloadmanager/demo/DownloadBatchStatusView.java
```diff
@@ -0,0 +1,82 @@
...
+        super(context, attrs);
+    }
+
+    @Override
+    protected void onFinishInflate() {
+        super.onFinishInflate();
+        View.inflate(getContext(), R.layout.merge_download_batch_status, this);
+
+        downloadBatchStatusView = findViewById(R.id.download_batch_status);
+        View pauseDownloadView = findViewById(R.id.button_pause_downloading);
```
zegnus: I'd avoid the local variables because if you create a `NO_OP` implementation for the `DownloadBatchStatusListener` then you can collapse all this code in just two lines

demo-simple/src/main/java/com/novoda/downloadmanager/demo/LandingActivity.java
```diff
@@ -0,0 +1,29 @@
...
+import android.view.View;
+
+public class LandingActivity extends AppCompatActivity {
+
+    @Override
+    protected void onCreate(@Nullable Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_landing);
+
+        View migrationView = findViewById(R.id.migration);
```
zegnus: same, would inline here and extract a common method with just the Activity class that you want to start, two lines of code.

demo-simple/src/main/java/com/novoda/downloadmanager/demo/DownloadBatchStatusView.java
MechaRyRy: 979d0f9

demo-simple/src/main/java/com/novoda/downloadmanager/demo/LandingActivity.java
MechaRyRy: 7bfea1a

zegnus (approved):

PR #418, opened 2018-07-12
Add Storage Root to Batch

MechaRyRy:
## Problem
In a client application we want to prevent downloads from being shown when the SD card is removed. This means we need a way to filter the downloads on the client side. We don't want to iterate through all of the files as this is an expensive operation.

## Solution
Add the `storage root` to the `Batch`. This storage root can then be used to filter the downloads in the client application.

## Screen Capture
This shows that the default value is now the internal storage.
screen shot 2018-07-12 at 20 42 43

### DIFF
--- a/library/roomSchemas/com.novoda.downloadmanager.RoomAppDatabase/3.json
+++ b/library/roomSchemas/com.novoda.downloadmanager.RoomAppDatabase/3.json
@@ -1,0 +1,144 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 3,
+    "identityHash": "cea0758d42e212b6739e5db5ac2080f1",
+    "entities": [
+      {
+        "tableName": "RoomBatch",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`batch_id` TEXT NOT NULL, `batch_title` TEXT, `batch_status` TEXT, `batch_downloaded_date_time_in_millis` INTEGER NOT NULL, `notification_seen` INTEGER NOT NULL, `storage_root` TEXT, PRIMARY KEY(`batch_id`))",
+        "fields": [
+          {
+            "fieldPath": "id",
+            "columnName": "batch_id",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "title",
+            "columnName": "batch_title",
+            "affinity": "TEXT",
+            "notNull": false
+          },
+          {
+            "fieldPath": "status",
+            "columnName": "batch_status",
+            "affinity": "TEXT",
+            "notNull": false
+          },
+          {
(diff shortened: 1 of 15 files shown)


library/src/main/java/com/novoda/downloadmanager/RoomDownloadsPersistence.java
```diff
@@ -56,7 +57,8 @@ public void persistBatch(DownloadsBatchPersisted batchPersisted) {
                     DownloadBatchIdCreator.createSanitizedFrom(roomBatch.id),
                     DownloadBatchStatus.Status.from(roomBatch.status),
                     roomBatch.downloadedDateTimeInMillis,
-                    roomBatch.notificationSeen
+                    roomBatch.notificationSeen,
+                    roomBatch.storageRoot
```
MechaRyRy, line 61: This line right here! I forgot to add this one line and it seemed that none of the SQL statements were working for me. After a break and thinking I realised that the SQL statements were workings, it was just being overwritten when the app ran with `empty` because this was missing 😿

library/roomSchemas/com.novoda.downloadmanager.RoomAppDatabase/3.json
```diff
@@ -0,0 +1,144 @@
+{
```
zegnus, line 1: I assume this file is autogenerated right?

library/roomSchemas/com.novoda.downloadmanager.RoomAppDatabase/3.json
```diff
@@ -0,0 +1,144 @@
...
+            "referencedColumns": [
+              "batch_id"
+            ]
+          }
+        ]
+      }
+    ],
+    "setupQueries": [
+      "CREATE TABLE IF NOT EXISTS room_master_table (id INTEGER PRIMARY KEY,identity_hash TEXT)",
+      "INSERT OR REPLACE INTO room_master_table (id,identity_hash) VALUES(42, \"cea0758d42e212b6739e5db5ac2080f1\")"
```
danybony, line 141: What's the reason for this?

zegnus (approved):

danybony (approved):
Looking good, just one question

library/roomSchemas/com.novoda.downloadmanager.RoomAppDatabase/3.json
MechaRyRy: This is autogenerated. It's the schema for the database, it's generated using the `Room` files 😄

library/roomSchemas/com.novoda.downloadmanager.RoomAppDatabase/3.json
MechaRyRy: Yep. It's the migration file.

PR #420, opened 2018-07-17
Remove custom file persistence

MechaRyRy:
## Problem 
In a client application we are receiving a crash because the `FilePersistenceType` is attempting to parse `Internal` which no longer exists now that we use `PATH` and `CUSTOM` only 😬 

## Solution
After some discussion we have decided to remove the `FilePersistenceType` and the ability to create a custom `FilePersistence`. We reasoned that now that we use the absolute path for creating files it is extremely unlikely that a client application would need to introduce their own file persistence variant.


library/roomSchemas/com.novoda.downloadmanager.RoomAppDatabase/4.json
```diff
@@ -0,0 +1,138 @@
+{
```
MechaRyRy, line 1: Auto-generated schema

zegnus (approved):

danybony (approved):

PR #422, opened 2018-07-25
Add transaction annotation to Room queries.

MechaRyRy:
## Problem

We're seeing the following error on some client applications:
```
Fatal Exception: android.database.sqlite.SQLiteDatabaseCorruptException
database disk image is malformed (code 11): , while compiling: PRAGMA journal_mode
```
The stacktrace points towards the load batches query. After reading some blog posts by the Room team I stumbled across this post: https://medium.com/google-developers/7-pro-tips-for-room-fbadea4bfbd1 point two suggests implementing transactions for large queries to prevent corruption of the database.

## Solution
Use the `Transaction` annotation for `Dao` query methods.


zegnus (approved):
Brilliant, I thought that those operations were already under the transaction contract!

PR #428, opened 2018-09-13
Add snapshot and release build-script to the project

MechaRyRy:
## Problem
Now that we are preparing to release v2 we need to still be able to release snapshots, but in a more manageable way. 

## Solution
Add the snapshot and release build-script that is used on some of our other projects. CI work still to be done.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,7 @@
+allprojects {
+    version = '2.0.0'
+}
+
 def teamPropsFile(propsFile) {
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -1,3 +1,7 @@
+allprojects {
+    version = '2.0.0'
```
zegnus, line 2: ❤️

zegnus (approved):

danybony (approved):

PR #435, opened 2018-09-25
Java doc storage requirement rule factory 

MechaRyRy:
## Problem
It seems that the storage requirement rules factory does not include any java doc, which might make it confusing. 

## Solution
JavaDoc! Rename the class so that it more explicitly mentions that this is the remaining space. This is a small rename and is super easy for clients as they are protected by our interface and factory.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/StorageRequirementRuleFactory.java
+++ b/library/src/main/java/com/novoda/downloadmanager/StorageRequirementRuleFactory.java
@@ -8,11 +8,30 @@ public final class StorageRequirementRuleFactory {
         // Uses static factory methods.
     }
 
+    /**
+     * Creates a storage requirement rule, where the storage bytes must be greater than the
+     * given bytes after a download completes. e.g. if specifying 100mb then the rule will
+     * be violated if the system memory remaining after the download is 99mb, preventing the
+     * download from being started.
+     *
+     * @param bytesRemainingAfterDownload the amount of storage in bytes that must be remaining after a download completes.
+     * @return the storage requirement rule to be evaluated when creating a file.
+     */
     public static StorageRequirementRule createByteBasedRule(long bytesRemainingAfterDownload) {
-        return new ByteBasedStorageRequirementRule(new StorageCapacityReader(), bytesRemainingAfterDownload);
+        return new ByteBasedRemainingStorageRequirementRule(new StorageCapacityReader(), bytesRemainingAfterDownload);
     }
 
+    /**
+     * Creates a storage requirement rule, where the storage bytes must be greater than the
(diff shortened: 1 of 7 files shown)


library/src/main/java/com/novoda/downloadmanager/StorageRequirementRuleFactory.java
```diff
@@ -8,11 +8,30 @@ private StorageRequirementRuleFactory() {
         // Uses static factory methods.
     }
 
+    /**
+     * Creates a storage requirement rule, where the storage bytes must be greater than the
+     * given bytes after a download completes. e.g. if specifying 100mb then the rule will
```
zegnus: should be 100MB (b -> bits, B -> bytes)

rock3r: Nit: `completes. e.g. if` -> `completes. E.g., if` 💃

danybony (dismissed):

library/src/main/java/com/novoda/downloadmanager/StorageRequirementRuleFactory.java
```diff
@@ -8,11 +8,30 @@ private StorageRequirementRuleFactory() {
         // Uses static factory methods.
     }
 
+    /**
+     * Creates a storage requirement rule, where the storage bytes must be greater than the
+     * given bytes after a download completes. E.g. if specifying 100MB then the rule will
+     * be violated if the system memory remaining after the download is 99MB, preventing the
```
zegnus: `system memory` isn't the memory, is the storage right?

library/src/main/java/com/novoda/downloadmanager/StorageRequirementRuleFactory.java
```diff
@@ -8,11 +8,30 @@ private StorageRequirementRuleFactory() {
...
+     */
     public static StorageRequirementRule createByteBasedRule(long bytesRemainingAfterDownload) {
-        return new ByteBasedStorageRequirementRule(new StorageCapacityReader(), bytesRemainingAfterDownload);
+        return new ByteBasedRemainingStorageRequirementRule(new StorageCapacityReader(), bytesRemainingAfterDownload);
     }
 
+    /**
+     * Creates a storage requirement rule, where the storage bytes must be greater than the
+     * a given percentage of the storage bytes after a download completes. E.g. if you specify 10% then the rule
+     * will be violated if the remaining system memory is at 9% of the total system memory, preventing the download
```
zegnus: `system memory` isn't the memory, is the storage right?

zegnus (approved):

danybony (approved):

PR #452, opened 2018-10-23
Migration fixes

MechaRyRy:
### Problem
The migration demo is a little broken. It did not take into account:
- current and total bytes do not match, resolving in a partial. 
- one file matches another from v1, forcing an empty file for the second usage as the first is deleted
- We didn't move the completed batch files

### Solution
Update the database and fix the above issues.


demo-simple/src/main/java/com/novoda/downloadmanager/demo/migration/CompletedDownloadBatchesExtractor.java
```diff
@@ -27,7 +27,8 @@
             + "batches INNER JOIN DownloadsByBatch ON DownloadsByBatch.batch_id = batches._id "
             + "WHERE batches._id NOT IN (SELECT DownloadsByBatch.batch_id FROM DownloadsByBatch "
             + "INNER JOIN batches ON batches._id = DownloadsByBatch.batch_id "
-            + "WHERE DownloadsByBatch._data IS NULL "
+            + "WHERE DownloadsByBatch.batch_total_bytes != DownloadsByBatch.batch_current_bytes "
+            + "OR DownloadsByBatch._data IS NULL "
```
danybony, line 31: 💯

demo-simple/src/main/java/com/novoda/downloadmanager/demo/migration/MigrationJob.java
```diff
@@ -66,24 +72,72 @@ public void run() {
...
+            CompletedDownloadBatch completedDownloadBatch = completeDownloadBatches.get(i);
+            downloadManager.addCompletedBatch(completedDownloadBatch);
+            completeDownloadBatches.remove(i);
+
+            for (CompletedDownloadFile completedDownloadFile : completedDownloadBatch.completedDownloadFiles()) {
+                migrateV1FileToV2Location(completedDownloadFile);
+
+                boolean usedInFurtherBatches = originalFileUsedInFurtherBatches(
+                        completedDownloadFile.originalFileLocation(),
+                        partialDownloadBatches,
```
danybony, line 104: At this point `partialDownloadBatches` is empty, right?

demo-simple/src/main/java/com/novoda/downloadmanager/demo/migration/MigrationJob.java
```diff
@@ -66,24 +72,72 @@ public void run() {
...
         database.deleteDatabase();
         onUpdate("Completed Migration");
     }
 
+    private boolean originalFileUsedInFurtherBatches(String originalFile,
+                                                     List<VersionOnePartialDownloadBatch> partialDownloadBatches,
+                                                     List<CompletedDownloadBatch> completeDownloadBatches) {
+
+        for (VersionOnePartialDownloadBatch partialDownloadBatch : partialDownloadBatches) {
+            for (String originalFileLocation : partialDownloadBatch.originalFileLocations()) {
```
danybony: here you can simply do 
```
 if (partialDownloadBatch.originalFileLocations().contains(originalFile)) {
       return true;
}
```

demo-simple/src/main/java/com/novoda/downloadmanager/demo/migration/MigrationJob.java
MechaRyRy: Yea should be, I didn't want to create another method that did the same thing but had one list less 😂

danybony (approved):

Issue #454, opened 2018-10-24
DownloadManager.submitAllStoredDownloads swallowing events


MechaRyRy:
## Problem 
It seems like that any call to the `submitAllStoredDownloads` is preventing further event calls, pause etc, from going through. I'm assuming that the reference we store in the internal map is being nuked on subsequent calls which leads to unexpected behaviour. 

#### Steps to reproduce
- Open the demo application and start downloading
- Spam the "Download only over wifi button"
- Try to pause and resume downloads / delete them

**Expected:** actions work as expected
**Actual:** It deletes downloads and then immediately re-queues them (this is probably due to adding the same reference to the executor multiple times)
**Actual:** It doesn't pause / resume the download.

## Solution
- Add additional logs around this `submitAllStoredDownloads` and the references accessed when performing events. 
- Find out where this problem is originating and patch the issue.
- I think we should drop `submitAllStoredDownloads` from the public API, calls to this method should only be handled internally. 

## Impact
This is a serious issue, we could be downloading multiple times, not removing downloads etc, this is a huge impact to our client applications.

MechaRyRy:
This doesn't seem to be tied to the job that reschedules because I can reproduce if I always remain in mobile network (making it never execute).

Lingviston:
Today I encountered the same bug, while playing with queue management.
My steps to reproduce (I used modified demo with 5 files instead of 2):
1. Start downloading of several files;
2. Verify that they're queued;
3. Verify that first file started downloading;
4. Fail downloading of the first file;
5. Verify that second file started downloading;
6. Pause or delete second file

**Expected**: Second file gets paused or deleted;
**Actual**: Second file continues downloading.

This happens because Novoda supports auto-restoration after connection lost. However ANY connection error is treated by Novoda as connection lost. The mechanism is the following: whenever connection error happens Novoda fetches the downloads from persistence storage and submits them to download. Those fetches downloads are represented by Java instances which are different from the initial ones. In the described scenario when user tries to pause or delete the second download (which is ongoing) after the failed first one it tries to manipulate with the "old" instance, while internally Novoda has already rewritten link to it with the new one.

I managed to fix it by removing
`downloadBatchMap.put(downloadBatch.getId(), downloadBatch);`
line from LiteDownloadManager#loadBatchesCallback and modifying the LiteDownloadManagerDownloader#download in the following way:
```
void download(DownloadBatch downloadBatch, Map<DownloadBatchId, DownloadBatch> downloadBatchMap) {
[...]

PR #456, opened 2018-10-29
DLM-454/Swallowing Events

MechaRyRy:
## Problem 
As per #454 there are inconsistencies when emitting events when queuing multiple times against the executor. After some investigation by @Lingviston and myself we have discovered that it is because of the references we are storing in the `map` and passing to the executor. Let's break it down with a table 

#### Before fix
Batch & Reference | Map | Executor 
--- | --- | ---
Batch 1 - ref 1 | Added (ref 1) | Batch 1 - ref 1
Batch 1 - ref 2 | Replaced (ref 2) | Batch 1 - ref 2

#### After fix
Batch & Reference | Map | Executor 
--- | --- | ---
Batch 1 - ref 1 | Added (ref 1) | Batch 1 - ref 1
Batch 1 - ref 2 | Not replaced (ref 1) | Batch 1 - ref 1

When we perform an action (pause), we are mutating the original ref (ref1) which will pause the download and eventually remove from the executor. If we then force a re-queue, using `submitAll` we queue a new ref (ref2), which replaces the original in the map and queues again on the executor. Essentially wiping the `pause` and saying we want to download again.

## Solution 
When the `map` contains a `DownloadBatch` we should use that to queue against the executor, it represents our mutated `DownloadBatch`, rather than using the new `DownloadBatch` as it is considered `fresh` and doesn't represent the state changes that have occurred in the `download-manager`. This means that we only perform a `put` in two locations, on `download` and on `addCompletedBatch`. 

## Additional
We should also remove a `DownloadBatch` from the internal map when it has been deleted. 

## Drawbacks?
The only thing I can think of is if a client adds the same download batch id for different batches, but that isn't supported anyway 

### Big thanks
Big thank you to @Lingviston, for taking the time to investigate this issue with me, you were a great help!

### DIFF
--- a/library/src/test/java/com/novoda/downloadmanager/LiteDownloadManagerTest.java
+++ b/library/src/test/java/com/novoda/downloadmanager/LiteDownloadManagerTest.java
@@ -192,6 +192,13 @@ public class LiteDownloadManagerTest {
         }
 
         @Test
+        public void removesBatchFromMap_whenDeletingBatch() {
+            liteDownloadManager.delete(DOWNLOAD_BATCH_ID);
+
+            assertThat(downloadingBatches).doesNotContainKey(DOWNLOAD_BATCH_ID);
+        }
+
+        @Test
         public void addsCallbackToInternalList() {
             DownloadBatchStatusCallback additionalDownloadBatchCallback = mock(DownloadBatchStatusCallback.class);
 
--- a/library/src/test/java/com/novoda/downloadmanager/LiteDownloadManagerTest.java
+++ b/library/src/test/java/com/novoda/downloadmanager/LiteDownloadManagerTest.java
@@ -105,6 +105,15 @@ public class LiteDownloadManagerTest {
         }
 
         @Test
+        public void doesNotPutBatchesIntoDownloadsMap_whenSubmittingAllStoredDownloads() {
+            downloadingBatches.clear();
+
+            liteDownloadManager.submitAllStoredDownloads(allStoredDownloadsSubmittedCallback);
+
+            assertThat(downloadingBatches).isEmpty();
+        }
+
+        @Test
         public void notifies_whenSubmittingAllStoredDownloads() {
             liteDownloadManager.submitAllStoredDownloads(allStoredDownloadsSubmittedCallback);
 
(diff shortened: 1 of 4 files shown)


zegnus (commented):
Did you try to smash pause/resume/delete buttons?

MechaRyRy:
@zegnus yes, I spammed all the buttons.

Lingviston:
Unit tests are missing.

MechaRyRy:
@Lingviston we can block this PR if you would like to wait until some unit tests are created. I unfortunately do not have much time to add them until later today.

Lingviston:
> @Lingviston we can block this PR if you would like to wait until some unit tests are created. I unfortunately do not have much time to add them until later today.

I'll add them.

Lingviston:
@Mecharyry as agreed I created unit tests for these scenarios. But I have no permissions to push into the branch.

MechaRyRy:
@Lingviston I've given you temporary contributor status 😄 normally you would need to fork this library and create a separate PR pushing to this branch.

Lingviston:
@Mecharyry got it. Thanks! Tests were pushed.

library/src/test/java/com/novoda/downloadmanager/LiteDownloadManagerTest.java
```diff
@@ -104,6 +104,15 @@ public void triggersDownloadOfBatches_whenSubmittingAllStoredDownloads() {
             inOrder.verify(downloadManagerDownloader).download(additionalDownloadBatch, downloadingBatches);
         }
 
+        @Test
+        public void doesNotPutBatchesIntoDownloadsMap_whenSubmittingAllStoredDownloads() {
```
MechaRyRy, line 108: I find this one very misleading. It doesn't put it into the map in this class true, **but** it does do it when calling `download` on the downloader, which is called from `submitAllStoredDownloads`.

Lingviston: Hmm, but that's implementation detail of the downloader. How would you call it?

MechaRyRy: I agree, it is an implementation detail. I guess we'll leave it, I just hope people don't make assumptions on how this all works 😬

MechaRyRy:
@Lingviston @zegnus are you both happy if we merge this now?

zegnus (approved):

PR #459, opened 2018-11-06
Implemented DownloadBatchRequirementRule

pglebocki:
**The problem.**
Need to have possibility to dynamically change storage size limit and block queued batches witch are oversized. Cannot use StorageRequirementRule because it is related to single file check, not all batch.

**The solution**
Implemented DownloadBatchRequirementRule (analogically to StorageRequirementRule) which validates batch total size before starting download.


hal9002:

library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
```diff
@@ -277,8 +289,8 @@ public void onServiceDisconnected(ComponentName name) {
                 downloadsFilePersistence,
                 downloadsPersistence,
                 callbackThrottleCreator,
-                connectionChecker
-        );
+                connectionChecker,
+                downloadBatchRequirementRules);
```
MechaRyRy: `);` on a newline to keep the formatting the same.

library/src/main/java/com/novoda/downloadmanager/DownloadBatchRequirementRule.java
```diff
@@ -0,0 +1,17 @@
+package com.novoda.downloadmanager;
+
+/**
+ * Used to define the rules that need to be satisfied to download an batch.
+ * Clients of this library can create their own custom implementations and
+ * pass them to {@link DownloadManagerBuilder#withStorageRequirementRules(StorageRequirementRule...)}.
```
MechaRyRy: Should be linking to -> https://github.com/novoda/download-manager/pull/459/files#diff-bee109b165737f88fe1d3425310bdd9bR171

pglebocki: nice catch :)

library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
pglebocki: sure

library/src/main/java/com/novoda/downloadmanager/DownloadBatchRequirementRule.java
```diff
@@ -0,0 +1,17 @@
...
+ */
+public interface DownloadBatchRequirementRule {
+
+    /**
+     * Determines whether a {@link DownloadBatch} has violated a size requirement rule.
+     *
+     * @param batchSize of the batch to check.
+     * @return whether the {@link DownloadBatch} has violated the rule.
+     */
+    boolean hasViolatedRule(Long batchSize);
```
Lingviston: Maybe provide some more information about the batch? Like batch id for example.

pglebocki: ok, good idea. I think I can pass full DownloadBatchStatus class

MechaRyRy:
@pglebocki would you be able to supply a usecase for this feature using the demo application? We would really like to see a gif of this in action using the demo if possible.

library/src/main/java/com/novoda/downloadmanager/DownloadsBatchPersistence.java
```diff
@@ -19,17 +19,19 @@
...
     private final CallbackThrottleCreator callbackThrottleCreator;
     private final ConnectionChecker connectionChecker;
+    private final DownloadBatchRequirementRule downloadBatchRequirementRule;
 
     DownloadsBatchPersistence(Executor executor,
                               DownloadsFilePersistence downloadsFilePersistence,
                               DownloadsPersistence downloadsPersistence,
                               CallbackThrottleCreator callbackThrottleCreator,
-                              ConnectionChecker connectionChecker) {
+                              ConnectionChecker connectionChecker, DownloadBatchRequirementRule downloadBatchRequirementRule) {
```
MechaRyRy: One parameter per line 😄

library/src/main/java/com/novoda/downloadmanager/DownloadsBatchPersistence.java
```diff
@@ -146,8 +148,8 @@ private DownloadBatch getDownloadBatch(FileOperations fileOperations, DownloadsB
                 downloadedFileSizeMap,
                 DownloadsBatchPersistence.this,
                 fileCallbackThrottle,
-                connectionChecker
-        );
+                connectionChecker,
+                downloadBatchRequirementRule);
```
MechaRyRy: `);` on a newline

library/src/main/java/com/novoda/downloadmanager/DownloadBatchFactory.java
```diff
@@ -83,8 +84,8 @@ static DownloadBatch newInstance(Batch batch,
                 new HashMap<>(),
                 downloadsBatchPersistence,
                 fileCallbackThrottle,
-                connectionChecker
-        );
+                connectionChecker,
+                downloadBatchRequirementRule);
```
MechaRyRy: `);` on a newline

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerDownloader.java
```diff
@@ -81,12 +85,9 @@ void download(DownloadBatch downloadBatch, Map<DownloadBatchId, DownloadBatch> d
...
         DownloadBatch batchToDownload = downloadBatchMap.get(downloadBatchId);
 
-        executor.submit(new Runnable() {
-            @Override
-            public void run() {
-                Wait.<Void>waitFor(serviceCriteria, waitForDownloadService)
-                        .thenPerform(executeDownload(batchToDownload, downloadBatchMap));
-            }
+        executor.submit(() -> {
+            Wait.<Void>waitFor(serviceCriteria, waitForDownloadService)
```
MechaRyRy: Did this convert it to a `Callable` or a `Runnable`? I believe this is why it wasn't inlined before.

library/src/main/java/com/novoda/downloadmanager/DownloadBatchRequirementRule.java
```diff
@@ -0,0 +1,18 @@
...
+public interface DownloadBatchRequirementRule {
+
+    /**
+     * Determines whether a {@link DownloadBatch} has violated a size requirement rule.
+     *
+     * @param batchSize of the batch to check.
+     * @param downloadBatchStatus of batch to check.
+     * @return whether the {@link DownloadBatch} has violated the rule.
+     */
+    boolean hasViolatedRule(DownloadBatchStatus downloadBatchStatus, Long batchSize);
```
MechaRyRy: You should only need `DownloadBatchStatus` as it already contains the batch size.

MechaRyRy:
test this

library/src/main/java/com/novoda/downloadmanager/DownloadsBatchPersistence.java
pglebocki: Ok, I will correct all formatting stuff :) . Btw. I saw your formatting rules are defined https://github.com/novoda/novoda/tree/master/ide-settings. I will apply it to my IDE

library/src/main/java/com/novoda/downloadmanager/DownloadBatchRequirementRule.java
pglebocki: Good point. I haven't noticed that before

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerDownloader.java
pglebocki: My IDE tells me that it is still runnable as it was before, however I will revert this change, because it is not part of this PR

pglebocki:
@Mecharyry it looks that this PR was automatically closed, can it be reopened or should I create new one will all corrections ?

MechaRyRy:
https://ci.novoda.com/view/PRB/job/download-manager-prb/909/

MechaRyRy:
@pglebocki it can be reopened when the above issue is resolved. It's an update to the merlin dependency. It should be `1.2.0`

pglebocki:
@Mecharyry from logs i understand that's it about publishing report failure. Is is something which I can fix in my PR or have to wait for some update on your site ?

MechaRyRy:
@pglebocki this is the issue
screenshot 2018-11-07 at 09 17 08

pglebocki:
@Mecharyry I've updated to 1.2.0 and addressed other comment. Can this PR be reopened ?

zegnus:
@pglebocki if you fix this:

> DownloadBatch.java:67, LineLengthCheck, Priority: Low
> --
> Line is longer than 150 characters (found 161).

I think you'll make the `ci` happy

pglebocki:
@zegnus Fixed. I just wonder why ./gradlew check doesn't indicate this issue. Is there different setup for CI ?

MechaRyRy:
@pglebocki if you run `./gradlew clean build` you should get the same issues locally. If not, then it might be the thresholds are not configured for local builds correctly 😬

pglebocki:
@Mecharyry It looks that ./gradlew clean build works succesfully for me, when run it locally. 
Max line length treshold is set to 150. For latest build logs I cannot figure out what happen

pglebocki:
@Mecharyry I saw that CI fail on PMD  _warnings_ but locally warnings are passed by static analysis. I think this should be consistent. However I fixed them.

MechaRyRy:
> If not, then it might be the thresholds are not configured for local builds correctly

I'm looking into the issue @pglebocki, sorry for the delay.

pglebocki:
@Mecharyry No problem. But it is not related to threshold setup, I think they are set properly. It is just about changing severity between warning/error. If CI doesn't pass warning level issue the local task should does the same ?

MechaRyRy:
@pglebocki There's a local file that deals with the thresholds when building locally, here's a PR that fixes the issue https://github.com/novoda/download-manager/pull/460

MechaRyRy:
@pglebocki if you merge develop into your branch it should now show the errors locally.

MechaRyRy:
> @pglebocki would you be able to supply a usecase for this feature using the demo application? We would really like to see a gif of this in action using the demo if possible.

We are awaiting this action I believe.

pglebocki:
> > @pglebocki would you be able to supply a usecase for this feature using the demo application? We would really like to see a gif of this in action using the demo if possible.
> 
> We are awaiting this action I believe.

@Mecharyry No, it is already done. Please open DemoActivity and use SeekBar.

MechaRyRy:
@pglebocki I have to respectively disagree. The reason we ask for `gif` is for future reference and to remove the need for anyone who is reviewing the PR from having to `checkout` the branch, install the apk and run the app to check the functionality. This is the reason why we requested it in gif form.

pglebocki:
@Mecharyry Ok, I got your point :) Please find attached gif below.

There is following usacase:
1. Batch size limit is set to max (should pass 2 batches)
2. Start downloading 2 batches sequentially.
3. When first batch is in progress then changing size limit.
4. Second download is not started because of not passing size condition.

dowbnload_batch_size

MechaRyRy:
@pglebocki I've noticed something that might be an issue from the gif. I assumed in the case of a violation we would return an error but I see in the UI that the download is still marked as `DOWNLOADING`. I find this a little misleading, can you explain the rationale behind this @pglebocki ? @zegnus what do you think also?

pglebocki:
> @pglebocki I've noticed something that might be an issue from the gif. I assumed in the case of a violation we would return an error but I see in the UI that the download is still marked as `DOWNLOADING`. I find this a little misleading, can you explain the rationale behind this @pglebocki ? @zegnus what do you think also?

@Mecharyry I agree that DOWNLOADING status is misleading here. I just forgot to change this to ERROR (exactly like in case of StorageRequirementRule for files) when rule is violated. I will updated this soon and send new movie.

pglebocki:
@Mecharyry  Error code added.

requirement_rule_for_batch

MechaRyRy:
@pglebocki Something doesn't look right in this gif. You change the slider and the first download still completes. Is the intention for this requirement to be evaluated when first triggering a download, or constantly as the batch is being downloaded?

pglebocki:
@Mecharyry My intention is to evaluate requirement only once before triggering download of each batch (not constantly while downloading).

MechaRyRy (approved):
@zegnus I am happy with this. Do you have any comments you would like to add?

pglebocki:
@zegnus @Mecharyry any update on this PR ? Just asking, because currently using my own fork, but it would be nice to use release build.

zegnus:
test this please

PR #475, opened 2019-02-21
Add support for root base directory

suhtai:
## Problem

When providing `"/"` as the path to use with `saveTo()` we end up with a final location that contains `///` on it.

## Solution

Use `Optional` to make sure we fall back to `absent` when providing a null or root path (using `filterNot`).

Then, when we compile the final path, we map to `path + '/'` or an empty string if absent.

## Rational for other changes

- Removed the use of `StringBuilder` as the compiler replaces `+` with it's own `StringBuilder` and since we are not looping (only case where there is a performance gain) it removes noice. 
- Use `File.separator` instead of `File.separatorChar`because it's shorter and the effective code of `append` for char and String is the same with the diference of calling `length()` on the string. Minimal impact, with easier readability.
- Added a couple of basic `Optional` operators. Why not.
- Also added a test to cover this use case.

### DIFF
--- a/demo-simple/build.gradle
+++ b/demo-simple/build.gradle
@@ -36,8 +36,8 @@ dependencies {
     implementation 'com.facebook.stetho:stetho:1.5.0'
     implementation 'com.android.support:appcompat-v7:28.0.0'
     implementation 'com.android.support.constraint:constraint-layout:1.1.3'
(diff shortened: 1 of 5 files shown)


MechaRyRy (dismissed):
LGTM, would be good to have @zegnus take a look too.

zegnus (approved):

Issue #472, opened 2019-02-08
Create a BatchBuilder from an existing batch


MechaRyRy:
### Problem
Sometimes clients may wish to create a `BatchBuilder` from a `Batch` so that they can add additional files etc to a batch prior to submitting it to the download manager.

### Solution
Expose a method on the `BatchBuilder` to create a new builder from a `Batch`.

PR #478, opened 2019-02-25
DLM-472/Create BatchBuilder from Batch

MechaRyRy:
## Problem
As mentioned in #472, clients may want to be able to add additional files to a download batch builder outside a single flow. At the moment `Batch` is immutable, which is good, but a builder can only be created and used in a single step. 

## Solution 
Create a mechanism where clients can create a `BatchBuilder` from a given `batch`, allowing them to append additional files before posting to the `download-manager`.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/BatchBuilder.java
+++ b/library/src/main/java/com/novoda/downloadmanager/BatchBuilder.java
@@ -21,6 +21,12 @@ public interface BatchBuilder {
      */
     Batch build();
 
+    /**
+     * Creates a {@link BatchBuilder} from a {@link Batch}.
+     *
+     * @param batch to create the {@link BatchBuilder} from.
+     * @return {@link BatchBuilder}.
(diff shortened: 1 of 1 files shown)


danybony (dismissed):
👍

library/src/main/java/com/novoda/downloadmanager/BatchBuilder.java
```diff
@@ -20,4 +20,12 @@
      * @return an instance of {@link Batch}.
      */
     Batch build();
+
+    static BatchBuilder from(Batch batch) {
```
danybony, line 30: Maybe do you want to add a javadoc description?

MechaRyRy: Ah true 😂

mr-archano (approved):
💯 thanks @Mecharyry for taking care of this!

Issue #253, opened 2017-11-14
Quick return warning


Mecharyry:
## Problem
We have a few quick return statements in `v2` that can lead to silent failures, no notifications to dev or client application that the flow is incorrect.

## Solution
As a minimum we should surface a warning log when this happens, or if we want to be more aggressive throw an exception. 

## Example
In `CallbackThrottleByTime` we have:

```
        if (timerTask == null) {
            return;
        }
```

so if we don't call `setCallback` when we trigger an update nothing will happen, no warnings, it will just silently do nothing.

Aiuspaktyn:
Same problem, on Android 5.1.1

zegnus:
Yes, this would be a `DeveloperError`

MechaRyRy:
Addressed in the above.

PR #479, opened 2019-02-26
DLM-253/Quick Returns

MechaRyRy:
## Problem 
As stated in #253, we have a number of `quick return` statements in the library that have little to no feedback when they are triggered. This means we could have errors in the flow but they are undiscoverable because we receive no notifications. 

## Solution
For the `CallbackThrottle` we now throw `IllegalStateException` when a callback is not present but we try to emit a `DownloadBatchStatus`. 

For the remainder of the `quick return`s, we now log messages to make debugging easier for maintainers and client applications. These are hidden behind the `Logger` so they need to be enabled through the `builder` to be shown in the `logcat`.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
@@ -22,6 +22,7 @@ class DownloadBatch {
 
     private static final int ZERO_BYTES = 0;
     private static final String STATUS = ", status ";
+    private static final String BATCH = "batch ";
 
     private final Map<DownloadFileId, Long> fileBytesDownloadedMap;
     private final InternalDownloadBatchStatus downloadBatchStatus;
@@ -68,7 +69,7 @@ class DownloadBatch {
 
         updateTotalSize();
 
-        Logger.v("batch " + downloadBatchStatus.getDownloadBatchId().rawId()
+        Logger.v(BATCH + downloadBatchStatus.getDownloadBatchId().rawId()
                          + STATUS + downloadBatchStatus.status()
                          + " totalBatchSize " + totalBatchSizeBytes);
 
@@ -79,7 +80,7 @@ class DownloadBatch {
                 downloadBatchRequirementRule,
                 totalBatchSizeBytes
         )) {
-            Logger.v("abort after getting total batch size download " + rawBatchId + STATUS + downloadBatchStatus.status());
+            Logger.v("abort after getting total " + BATCH + "size download " + rawBatchId + STATUS + downloadBatchStatus.status());
             return;
         }
 
@@ -142,7 +143,7 @@ class DownloadBatch {
                                             DownloadsBatchPersistence downloadsBatchPersistence,
                                             DownloadBatchStatusCallback callback) {
         if (downloadBatchStatus.status() == DELETING) {
-            Logger.v("sync delete and mark as deleted batch " + downloadBatchStatus.getDownloadBatchId().rawId());
+            Logger.v("sync delete and mark as deleted " + BATCH + downloadBatchStatus.getDownloadBatchId().rawId());
             downloadBatchStatus.markAsDeleted();
             downloadsBatchPersistence.deleteSync(downloadBatchStatus);
             notifyCallback(callback, downloadBatchStatus);
@@ -163,13 +164,13 @@ class DownloadBatch {
                                             DownloadBatchStatusCallback callback,
                                             DownloadsBatchPersistence downloadsBatchPersistence) {
         if (downloadBatchStatus.status() == DELETING) {
-            Logger.v("abort processNetworkError, the batch " + downloadBatchStatus.getDownloadBatchId().rawId() + " is deleting");
+            Logger.v("abort processNetworkError, the " + BATCH + downloadBatchStatus.getDownloadBatchId().rawId() + " is deleting");
             return;
         }
         downloadBatchStatus.markAsWaitingForNetwork(downloadsBatchPersistence);
         notifyCallback(callback, downloadBatchStatus);
         Logger.v(
-                "scheduleRecovery for batch "
+                "scheduleRecovery for " + BATCH
                         + downloadBatchStatus.getDownloadBatchId().rawId()
                         + STATUS
                         + downloadBatchStatus.status()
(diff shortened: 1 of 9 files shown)


library/src/main/java/com/novoda/downloadmanager/FileCallbackThrottleByProgressIncrease.java
```diff
@@ -14,7 +14,7 @@ public void setCallback(DownloadBatchStatusCallback callback) {
     @Override
     public void update(DownloadBatchStatus currentDownloadBatchStatus) {
         if (callback == null) {
-            return;
+            throw new IllegalStateException("A DownloadBatchStatusCallback must be set before an update is called.");
```
MechaRyRy: I'm a little worried about introducing this change but it is a valid change that we should make.

library/src/main/java/com/novoda/downloadmanager/FileCallbackThrottleByTime.java
```diff
@@ -19,7 +19,7 @@ public void setCallback(final DownloadBatchStatusCallback callback) {
     @Override
     public void update(DownloadBatchStatus downloadBatchStatus) {
         if (callback == null) {
-            return;
+            throw new IllegalStateException("A DownloadBatchStatusCallback must be set before an update is called.");
```
MechaRyRy: I'm a little worried about introducing this change but it is a valid change that we should make.

library/src/test/java/com/novoda/downloadmanager/FileCallbackThrottleByProgressIncreaseTest.java
```diff
@@ -42,16 +44,6 @@ public void doesNotEmit_whenPercentageIsUnchanged() {
         then(downloadBatchCallback).should(never()).onUpdate(any(DownloadBatchStatus.class));
     }
 
-    @Test
-    public void doesNotEmitsStatus_whenStoppingUpdates() {
```
MechaRyRy, line 46: Why did this get removed 🤔

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -289,9 +292,12 @@ private static boolean networkError(InternalDownloadBatchStatus downloadBatchSta
     }
 
     void pause() {
-        Logger.v("pause batch " + downloadBatchStatus.getDownloadBatchId().rawId() + ", " + STATUS + " " + downloadBatchStatus.status());
+        Logger.v("pause " + BATCH + downloadBatchStatus.getDownloadBatchId().rawId() + STATUS + downloadBatchStatus.status());
         DownloadBatchStatus.Status status = downloadBatchStatus.status();
```
danybony, line 296: I'm commenting it here, but this is valid for most of these methods:
if you move `downloadBatchStatus.status()` at the beginning of the method, then you can directly use `status` in all the logs and checks

library/src/main/java/com/novoda/downloadmanager/FileCallbackThrottleByProgressIncrease.java
danybony: On one hand I agree, but on the other then I'll refactor how this callback works: 
in `stopUldates()` we set the callback to null, so we are not implying that we are not forwarding anything anymore (callback is null) but also that we won't receive any further update from downstream (because otherwise we'll crash).
This second assumption is wrong in my opinion...

library/src/main/java/com/novoda/downloadmanager/FileCallbackThrottleByTime.java
danybony: same as before

danybony (commented):
Good update, but I have a few concerns

library/src/main/java/com/novoda/downloadmanager/FileCallbackThrottleByProgressIncrease.java
MechaRyRy: Sorry but I don't understand, it might be because of the number of negations in the paragraph 😂 You are saying you don't agree with this change? I'm not sure I follow the refactoring you propose.

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
MechaRyRy: ea465c4

library/src/main/java/com/novoda/downloadmanager/FileCallbackThrottleByProgressIncrease.java
danybony: Sorry, what I meant is that I don't agree with this change if it's alone because we are connecting two external factors which in my opinion should stay disconnected, from a user of this class point of view:

- **output**: it is controlled by the method `stopUpdates()`, where we stop forwarding events to the output callback by setting its reference to `null`
- **input**: it is this method `update(DownloadBatchStatus currentDownloadBatchStatus)`, where we might continue receiving updates, but before this PR we were just doing nothing if the output was stopped, while now we crash

So the issue for me is that the user of this class needs to take care about its state. It made more sense to me before, having source (emitting updates) and controller (stopping output) possibly done by two different entities.

MechaRyRy: Fair enough. How about I revert these two changes and replace with a warning log?

danybony: Sounds good 👍

MechaRyRy:
Removing the ❌ because the risk I was worried about no longer exists after some discussion. Thanks @danybony ❤️

library/src/test/java/com/novoda/downloadmanager/FileCallbackThrottleByProgressIncreaseTest.java
```diff
@@ -20,9 +20,11 @@
 
     private final FileCallbackThrottleByProgressIncrease callbackThrottleByProgressIncrease = new FileCallbackThrottleByProgressIncrease();
 
-    @Test(expected = IllegalStateException.class)
-    public void throws_whenCallbackUnset() {
+    @Test
+    public void doesNothing_whenCallbackUnset() {
         callbackThrottleByProgressIncrease.update(percentageIncreasedStatus);
+
+        then(downloadBatchCallback).should(never()).onUpdate(percentageIncreasedStatus);
```
danybony: 👍

danybony (approved):

PR #480, opened 2019-02-26
Normalise paths when building BatchFile

frapontillo:
## Problem

When building a `BatchFile` with `LiteBatchFileBuilder`, we were simply checking that `path` wasn't the file separator `/`, so we wouldn't add a duplicated separator after that. But that fix (#475) didn't work for cases where the `path` is something like `my/path/to/the/file/` (with a file separator at the end), so we were essentially still adding a double `/`, causing issues like failed downloads.

## Solution

The solution consists in treating each string we need to concatenate as the same kind of path segment:

1. If the first segment (`storageRoot.path()`) starts with a file separator, we add a file separator to the `StringBuilder`
2. Then, for each element
    1. If the element is empty, we ignore it
    1. We add a file separator to the `StringBuilder` if the last char of the `StringBuilder` is not a file separator
    1. We remove any leading or trailing file separators (`/`) from the segment
    1. We add the segment to the `StringBuilder`

## Tests

Tests have been added to prove that all edge and middle cases are covered for.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
+++ b/library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
@@ -70,7 +70,7 @@ final class LiteBatchFileBuilder implements InternalBatchFileBuilder {
             if (path.isEmpty()) {
                 continue; // ignore empty paths
             }
-            if (!isLastCharFileSeparator(stringBuilder)) {
+            if (isLastCharNotFileSeparator(stringBuilder)) {
                 stringBuilder.append(File.separatorChar);
             }
             stringBuilder.append(removeLeadingTrailingFileSeparator(path));
@@ -93,8 +93,8 @@ final class LiteBatchFileBuilder implements InternalBatchFileBuilder {
         return element.substring(beginIndex, endIndex);
     }
 
-    private boolean isLastCharFileSeparator(StringBuilder stringBuilder) {
-        return stringBuilder.length() <= 0 || stringBuilder.charAt(stringBuilder.length() - 1) == File.separatorChar;
+    private boolean isLastCharNotFileSeparator(StringBuilder stringBuilder) {
+        return stringBuilder.length() > 0 && stringBuilder.charAt(stringBuilder.length() - 1) != File.separatorChar;
     }
 
 }
--- a/library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
+++ b/library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
@@ -61,8 +61,9 @@ final class LiteBatchFileBuilder implements InternalBatchFileBuilder {
     private String buildPath(String... paths) {
         StringBuilder stringBuilder = new StringBuilder();
 
-        if (paths[0].startsWith(File.separator)) {
-            stringBuilder.append(File.separator); // if storage path starts with /, we honour it
+        String storagePath = paths[0];
+        if (storagePath.startsWith(File.separator)) {
+            stringBuilder.append(File.separator);
         }
 
         for (String path : paths) {
--- a/library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
+++ b/library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
@@ -75,22 +75,22 @@ final class LiteBatchFileBuilder implements InternalBatchFileBuilder {
     }
 
     private String sanitise(String path) {
-        String[] pathSegments = path.split(File.separator);
+        String[] pathSegments = path.split(File.separatorChar == '\\' ? "\\\\" : File.separator);
         CharSequence[] filteredPathSegments = filterEmptySegmentsOut(pathSegments);
         return joinWithFileSeparator(filteredPathSegments);
     }
 
-    private CharSequence[] filterEmptySegmentsOut(String[] pathSegments) {
+    private CharSequence[] filterEmptySegmentsOut(String... pathSegments) {
         List<CharSequence> filteredPathSegments = new ArrayList<>();
         for (String pathSegment : pathSegments) {
             if (!pathSegment.isEmpty()) {
                 filteredPathSegments.add(pathSegment);
             }
         }
-        return filteredPathSegments.toArray(new CharSequence[0]);
+        return filteredPathSegments.toArray(new CharSequence[filteredPathSegments.size()]);
     }
 
-    private String joinWithFileSeparator(CharSequence[] elements) {
+    private String joinWithFileSeparator(CharSequence... elements) {
         StringBuilder stringBuilder = new StringBuilder();
         for (int i = 0; i < elements.length; i++) {
             stringBuilder.append(elements[i]);
--- a/library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
+++ b/library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
@@ -75,7 +75,7 @@ final class LiteBatchFileBuilder implements InternalBatchFileBuilder {
     }
 
     private String sanitise(String path) {
-        String[] pathSegments = path.split("/");
+        String[] pathSegments = path.split(File.separator);
         CharSequence[] filteredPathSegments = filterEmptySegmentsOut(pathSegments);
         return joinWithFileSeparator(filteredPathSegments);
     }
--- a/library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
+++ b/library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
@@ -6,6 +6,8 @@ import java.util.List;
 
 final class LiteBatchFileBuilder implements InternalBatchFileBuilder {
 
+    private static final String SEPARATOR_AS_REGEX = File.separatorChar == '\\' ? "\\\\" : File.separator;
+
     private final StorageRoot storageRoot;
     private final DownloadBatchId downloadBatchId;
     private final String networkAddress;
@@ -75,7 +77,7 @@ final class LiteBatchFileBuilder implements InternalBatchFileBuilder {
     }
 
     private String sanitise(String path) {
-        String[] pathSegments = path.split(File.separatorChar == '\\' ? "\\\\" : File.separator);
+        String[] pathSegments = path.split(SEPARATOR_AS_REGEX);
         CharSequence[] filteredPathSegments = filterEmptySegmentsOut(pathSegments);
         return joinWithFileSeparator(filteredPathSegments);
     }
--- a/library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
+++ b/library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
@@ -40,20 +40,60 @@ final class LiteBatchFileBuilder implements InternalBatchFileBuilder {
 
     @Override
     public BatchFileBuilder saveTo(String path, String fileName) {
-        this.path = Optional.fromNullable(path).filterNot("/"::equals);
+        this.path = Optional.fromNullable(path);
         this.fileName = Optional.fromNullable(fileName);
         return this;
     }
 
     @Override
     public BatchBuilder apply() {
-        String absolutePath = storageRoot.path() + File.separator
-                + downloadBatchId.rawId() + File.separator
-                + path.map((it) -> it + File.separator).or("")
-                + fileName.getOrElse(() -> FileNameExtractor.extractFrom(networkAddress));
+        String absolutePath = buildPath(
+                storageRoot.path(),
+                downloadBatchId.rawId(),
+                path.or(""),
+                fileName.getOrElse(() -> FileNameExtractor.extractFrom(networkAddress))
+        );
 
         parentBuilder.withFile(new BatchFile(networkAddress, downloadFileId, absolutePath));
         return parentBuilder;
+    }
+
+    private String buildPath(String... elements) {
+        StringBuilder stringBuilder = new StringBuilder();
+
+        if (elements[0].startsWith(File.separator)) {
+            stringBuilder.append(File.separator); // if storage path starts with /, we honour it
+        }
+
+        for (String element : elements) {
+            if (element.isEmpty()) {
+                continue; // ignore empty paths
+            }
+            if (!isLastCharFileSeparator(stringBuilder)) {
+                stringBuilder.append(File.separatorChar);
+            }
+            stringBuilder.append(removeLeadingTrailingFileSeparator(element));
+        }
+
+        return stringBuilder.toString();
+    }
+
+    private String removeLeadingTrailingFileSeparator(String element) {
+        int beginIndex = 0;
+        int endIndex = element.length();
+
+        if (element.charAt(0) == File.separatorChar) {
+            beginIndex = 1;
+        }
+        int lastIndexOfSeparator = element.lastIndexOf(File.separatorChar);
+        if (lastIndexOfSeparator != 0 && lastIndexOfSeparator == element.length() - 1) {
+            endIndex = element.length() - 1;
+        }
+        return element.substring(beginIndex, endIndex);
+    }
+
+    private boolean isLastCharFileSeparator(StringBuilder stringBuilder) {
+        return stringBuilder.length() <= 0 || stringBuilder.charAt(stringBuilder.length() - 1) == File.separatorChar;
     }
 
 }
--- a/library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
+++ b/library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
@@ -58,21 +58,21 @@ final class LiteBatchFileBuilder implements InternalBatchFileBuilder {
         return parentBuilder;
     }
 
-    private String buildPath(String... elements) {
+    private String buildPath(String... paths) {
         StringBuilder stringBuilder = new StringBuilder();
 
-        if (elements[0].startsWith(File.separator)) {
+        if (paths[0].startsWith(File.separator)) {
             stringBuilder.append(File.separator); // if storage path starts with /, we honour it
         }
 
-        for (String element : elements) {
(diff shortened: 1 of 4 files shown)


library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
```diff
@@ -40,20 +40,60 @@ public BatchFileBuilder saveTo(String path) {
 
     @Override
     public BatchFileBuilder saveTo(String path, String fileName) {
-        this.path = Optional.fromNullable(path).filterNot("/"::equals);
```
frapontillo, line 43: We thought we fixed it by filtering out paths that were `/` only, but this wouldn't solve the issue with a `path` that **ends** in `/` (e.g. `my/path/`).

library/src/test/java/com/novoda/downloadmanager/LiteBatchFileBuilderTest.java
```diff
@@ -0,0 +1,99 @@
...
+    private final InternalBatchBuilder batchBuilder = mock(InternalBatchBuilder.class);
+    private final StorageRoot storageRoot = mock(StorageRoot.class);
+    private final DownloadBatchId downloadBatchId = mock(DownloadBatchId.class);
+
+    private LiteBatchFileBuilder liteBatchFileBuilder;
+
+    @Before
+    public void setUp() {
+        when(downloadBatchId.rawId()).thenReturn("my-movie");
+        liteBatchFileBuilder = new LiteBatchFileBuilder(
```
MechaRyRy: Could move this up to the field and make it `final`.

library/src/test/java/com/novoda/downloadmanager/LiteBatchFileBuilderTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+    private final InternalBatchBuilder batchBuilder = mock(InternalBatchBuilder.class);
+    private final StorageRoot storageRoot = mock(StorageRoot.class);
+    private final DownloadBatchId downloadBatchId = mock(DownloadBatchId.class);
+
+    private LiteBatchFileBuilder liteBatchFileBuilder;
+
+    @Before
+    public void setUp() {
+        when(downloadBatchId.rawId()).thenReturn("my-movie");
```
MechaRyRy: Can use `BDD` mockito. `given().willReturn`.

library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
```diff
@@ -40,20 +40,60 @@ public BatchFileBuilder saveTo(String path) {
...
+            if (element.isEmpty()) {
+                continue; // ignore empty paths
+            }
+            if (!isLastCharFileSeparator(stringBuilder)) {
+                stringBuilder.append(File.separatorChar);
+            }
+            stringBuilder.append(removeLeadingTrailingFileSeparator(element));
+        }
+
+        return stringBuilder.toString();
```
suhtai, line 76: To avoid using `StringBuilder` and simplify the logic, we could join all the elements (using something like `TextUtils` or similar) and then remove duplicated separators until none are found and finally any extra wrapping separators.

library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
```diff
@@ -40,20 +40,60 @@ public BatchFileBuilder saveTo(String path) {
...
+    private String buildPath(String... elements) {
+        StringBuilder stringBuilder = new StringBuilder();
+
+        if (elements[0].startsWith(File.separator)) {
+            stringBuilder.append(File.separator); // if storage path starts with /, we honour it
+        }
+
+        for (String element : elements) {
+            if (element.isEmpty()) {
+                continue; // ignore empty paths
```
MechaRyRy: Could call `elements` `paths`, thereby removing the comment?

library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
```diff
@@ -40,20 +40,60 @@ public BatchFileBuilder saveTo(String path) {
...
+        );
 
         parentBuilder.withFile(new BatchFile(networkAddress, downloadFileId, absolutePath));
         return parentBuilder;
     }
 
+    private String buildPath(String... elements) {
+        StringBuilder stringBuilder = new StringBuilder();
+
+        if (elements[0].startsWith(File.separator)) {
```
MechaRyRy: Could pull out the `elements[0]`, name it `storageRootPath` and remove the comment below.

library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
```diff
@@ -40,20 +40,60 @@ public BatchFileBuilder saveTo(String path) {
...
+            stringBuilder.append(File.separator); // if storage path starts with /, we honour it
+        }
+
+        for (String element : elements) {
+            if (element.isEmpty()) {
+                continue; // ignore empty paths
+            }
+            if (!isLastCharFileSeparator(stringBuilder)) {
```
MechaRyRy: Could invert the statement inside this method and rename to remove the `!` negation

library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
```diff
@@ -40,20 +40,60 @@ public BatchFileBuilder saveTo(String path) {
...
+        int beginIndex = 0;
+        int endIndex = element.length();
+
+        if (element.charAt(0) == File.separatorChar) {
+            beginIndex = 1;
+        }
+        int lastIndexOfSeparator = element.lastIndexOf(File.separatorChar);
+        if (lastIndexOfSeparator != 0 && lastIndexOfSeparator == element.length() - 1) {
+            endIndex = element.length() - 1;
+        }
```
suhtai, line 91: This check could just check the last character of the element:
```
if(element.charAt(endIndex - 1) == File.separatorChar) {
  endIndex++
}
```
Both start and end could also use a `while` to check multiple ones:
```
while(element.charAt(beginIndex) == File.separatorChar && beginIndex < endIndex) {
  beginIndex++;
}
while(element.charAt(endIndex - 1) == File.separatorChar && beginIndex < endIndex) {
  endIndex--;
}
```

library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
```diff
@@ -40,20 +40,60 @@ public BatchFileBuilder saveTo(String path) {
...
+                downloadBatchId.rawId(),
+                path.or(""),
+                fileName.getOrElse(() -> FileNameExtractor.extractFrom(networkAddress))
+        );
 
         parentBuilder.withFile(new BatchFile(networkAddress, downloadFileId, absolutePath));
         return parentBuilder;
     }
 
+    private String buildPath(String... elements) {
```
MechaRyRy: The only problem with `String... elements` is that you can change the order and everything breaks 😄 but I guess this is backed with tests so 🤷‍♂️ 😄

library/src/test/java/com/novoda/downloadmanager/LiteBatchFileBuilderTest.java
suhtai: Or move it locally with a `given` factory method which takes in the conditions of the subject

library/src/test/java/com/novoda/downloadmanager/LiteBatchFileBuilderTest.java
```diff
@@ -0,0 +1,99 @@
+package com.novoda.downloadmanager;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+public class LiteBatchFileBuilderTest {
```
MechaRyRy, line 10: Wondering if we should have a test / tests with a ludicrous number of Separators spaced throughout the `root` and the `paths`? At the moment the most extreme case we have is `/path/`.

library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
MechaRyRy: I like the idea, it might make it a bit simpler. I wonder how heavy `TextUtils` or similar is compared to this method 🤔

library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
```diff
@@ -40,20 +40,60 @@ public BatchFileBuilder saveTo(String path) {
...
+            beginIndex = 1;
+        }
+        int lastIndexOfSeparator = element.lastIndexOf(File.separatorChar);
+        if (lastIndexOfSeparator != 0 && lastIndexOfSeparator == element.length() - 1) {
+            endIndex = element.length() - 1;
+        }
+        return element.substring(beginIndex, endIndex);
+    }
+
+    private boolean isLastCharFileSeparator(StringBuilder stringBuilder) {
```
suhtai: This method can take a `CharSequence` Always use the most abstract form of a type possible (Effective Java)

library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
suhtai: `continue` == `goto` :trollface:

MechaRyRy (commented):
This is definitely heading in the right direction 💯 ❤️

library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
frapontillo: I personally find the current approach easier to understand, as we are building the string one step at a time, instead of doing backtracking on something already built.

library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
frapontillo: 😄 I would have used `filter` but you know, Java.

library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
frapontillo: In the first snippet, you probably meant `endIndex--`. Anyway, it would crash for the case of element being `/`, since `beginIndex` would be `1` and `endIndex` would end up being `0`.

library/src/test/java/com/novoda/downloadmanager/LiteBatchFileBuilderTest.java
frapontillo: So, I agree we could do that (sanitise all elements throughout) but I think that could also be the user's job. I think accommodating for leading and trailing separators is enough (for now) to fix some weird edge cases.

library/src/test/java/com/novoda/downloadmanager/LiteBatchFileBuilderTest.java
frapontillo: Done in 469aca79cb5739c28aa31d4338a2022f46d89ceb.

library/src/test/java/com/novoda/downloadmanager/LiteBatchFileBuilderTest.java
frapontillo: Done in d5a16cbdfb291c625e4d91fccd0c0d3d7bc269fc.

library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
frapontillo: Done in 9c2bd121b95df17de9e0497cf177497b106bf482.

library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
frapontillo: Done in 8998736eea676b0be82a8a1d7c4d4bed9eeff0ff.

library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
frapontillo: Done in 8c627c326481c8c56968bc71bab1ea49b4d5da92.

library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
frapontillo: Done in 3dca3ed8c1cca7699221c07a85aae2ed99525203.

library/src/test/java/com/novoda/downloadmanager/LiteBatchFileBuilderTest.java
frapontillo: Hmmm, I have slightly changed `doesNotAddDuplicateSeparators` in bd9a16a824a8926f8bb08bb9521220996417fe05 to test the case where every single path element has leading and trailing separators.

library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
suhtai: `beginIndex <= endIndex` then? 😁

library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
```diff
@@ -40,20 +40,61 @@ public BatchFileBuilder saveTo(String path) {
...
+        StringBuilder stringBuilder = new StringBuilder();
+
+        String storagePath = paths[0];
+        if (storagePath.startsWith(File.separator)) {
+            stringBuilder.append(File.separator);
+        }
+
+        for (String path : paths) {
+            if (path.isEmpty()) {
+                continue; // ignore empty paths
```
MechaRyRy: Didn't remove the comment 👅

library/src/test/java/com/novoda/downloadmanager/LiteBatchFileBuilderTest.java
MechaRyRy: Sorry, my comment was hinting at what happens if you have `//path//` for example.

library/src/test/java/com/novoda/downloadmanager/LiteBatchFileBuilderTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+        verify(batchBuilder).withFile(batchFileWithPath("root/my-movie/my/path/my-movie.mp4"));
+    }
+
+    @Test
+    public void doesNotAddDuplicateSeparators() {
+        given(storageRoot.path()).willReturn("/root/");
+        given(downloadBatchId.rawId()).willReturn("/my-movie/");
+
+        liteBatchFileBuilder
```
suhtai: I meant to have the builder as a local property in the test functions 😁

library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
frapontillo: Oh, i thought the comment was for the `storagePath`!

frapontillo:
Implementation changed in 0224efaef89fd01b8e8c951efaeff7cf3cf4b5da to clean up internal duplicated separators too 💯.

library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
```diff
@@ -40,20 +42,62 @@ public BatchFileBuilder saveTo(String path) {
...
 
+    private String buildPath(String... paths) {
+        StringBuilder stringBuilder = new StringBuilder();
+
+        String storagePath = paths[0];
+        if (storagePath.startsWith(File.separator)) {
+            stringBuilder.append(File.separator);
+        }
+
+        String joinedPaths = sanitise(joinWithFileSeparator(paths));
```
frapontillo, line 73: The new logic ✨

library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
```diff
@@ -40,20 +42,62 @@ public BatchFileBuilder saveTo(String path) {
...
+        List<CharSequence> filteredPathSegments = new ArrayList<>();
+        for (String pathSegment : pathSegments) {
+            if (!pathSegment.isEmpty()) {
+                filteredPathSegments.add(pathSegment);
+            }
+        }
+        return filteredPathSegments.toArray(new CharSequence[0]);
+    }
+
+    private String joinWithFileSeparator(CharSequence[] elements) {
```
frapontillo: I opted to re-implement the `join` method since the `TextUtils` one can't be used in tests and `String.join` is definitely not available for our min API (16).

library/src/test/java/com/novoda/downloadmanager/LiteBatchFileBuilderTest.java
```diff
@@ -0,0 +1,96 @@
...
+        given(storageRoot.path()).willReturn("/root/");
+        given(downloadBatchId.rawId()).willReturn("/my-movie/");
+
+        whenLiteBatchFileBuilderApply("/my/path/", "/my-movie.mp4");
+
+        verify(batchBuilder).withFile(batchFileWithPath("/root/my-movie/my/path/my-movie.mp4"));
+    }
+
+    @Test
+    public void doesNotAddDuplicateSeparatorsWhenInMiddleOfSegments() {
```
frapontillo, line 48: Here's your ludicrous test @Mecharyry 😆

frapontillo:
We leave the developer to specify the file path as best as possible. This would only provide some safety around what's been passed, which should be separated with a `File.separator`.

MechaRyRy (dismissed):
❤️ great stuff ❤️

suhtai (approved):

MechaRyRy (approved):

Issue #483, opened 2019-02-28
Network error treatment


MechaRyRy:
## Problem
Collapsing #458 and #466 into one issue so everything is easier to deal with. 

The crux of the issue is that all `network issues` are treated as the same when they are not. Loss of connectivity that is brief is different from a server always returning an error code that will never resolve in an asset being downloaded. We need to stop treating them like they are the same. 

## Potential Solution

These needs some discussion between the maintainers, @zegnus, @danybony and @Mecharyry, and the community that opened the original issues @matecode @Lingviston. We need to understand what level of control client applications need and how we can expose this in the library. 

## Notes
- Alignment on the callback API when we perform network calls. Currently the pulling of a file provides more information than the size request.

matecode:
For me as a developer a download manager should be a tool. This means that any errors that result in download can not being started or finished should be reported to the client so that it can make an informed decision about what to do. That means all network errors or server errors like SSLHandShakeExceptions or HTTP40X. On the other hand, this also means that all problems that make a download possible, but not now, for example that there is no network connection or similar, should be handled by the tool itself. Maybe different configuration levels are useful.

MechaRyRy:
We've spent a little bit of time talking about this today and here's what I took away from our meeting:

- `FileDownloader.onError(String cause)` needs to be more strongly typed so we return the `network error code` and an `associated message`. 
- `DownloadError` should be more strongly typed. Perhaps it should be `DownloadError<T>` so we type it to the particular scenario and the message becomes an instance of `T` instead of `String`. That way we provide tailed information based on what has broken. 
- Create an interface that can determine how individual network errors are handled for each batch. By default we would have an implementation that retries automatically every error. Clients can then override this information to customise how they would like each error to be treated. 
- We should follow this same process for the `FileSizeRequester` too. At the moment it returns just a `FileSize` which can be `0` in the event it fails. 

@danybony @zegnus can you let me know if I have missed anything from the above?

MechaRyRy:
I've created a [feature branch](https://github.com/novoda/download-manager/tree/DLM-483/feature_network_errors) where we can track the current progress. I'll be branching from it to tackle the list above, will update this issue with more information as I have it.

MechaRyRy:
`FileDownloader` strong type -> https://github.com/novoda/download-manager/tree/DLM-483/file_downloader_strong_type

PR #486, opened 2019-03-14
DLM-483/FileDownloader Error Strong Typing

MechaRyRy:
⚠️ Against a feature branch to ensure we have addressed all of the issues outlined in #483 ⚠️ 

## Problem 
We need to introduce a better retry and failure mechanism when attempting to download a file. At the moment we only return a `message` associated with the failure from the `FileDownloader` which is not enough information to allow us or clients to perform any meaningful retrial / error flow. 

## Solution 
Introduce a `FileDownloader.FileDownloadError` interface and an associated model that represents some additional data that is required in order to be able to perform meaningful error handling. 

Not currently hooked up to any error handling, just passes the new model back to the `DownloadBatch` who then transforms it back to the old error. This will be used when we introduce the new error handling interface. See #483 for more information.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadErrorFactory.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadErrorFactory.java
@@ -41,7 +41,7 @@ final class DownloadErrorFactory {
         return new DownloadError(DownloadError.Type.FILE_CANNOT_BE_WRITTEN, cannotWriteToFileMessage);
     }
 
-    static DownloadError createNetworkError(FileDownloader.FileDownloadError error) {
+    static DownloadError createNetworkError(FileDownloader.Error error) {
         String networkErrorMessage = "Network error, cannot download file. Cause: " + error.message();
         return new DownloadError(DownloadError.Type.NETWORK_ERROR_CANNOT_DOWNLOAD_FILE, networkErrorMessage);
     }
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadErrorFactory.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadErrorFactory.java
@@ -41,8 +41,8 @@ final class DownloadErrorFactory {
         return new DownloadError(DownloadError.Type.FILE_CANNOT_BE_WRITTEN, cannotWriteToFileMessage);
     }
 
-    static DownloadError createNetworkError(String networkErrorCause) {
-        String networkErrorMessage = "Network error, cannot download file. Cause: " + networkErrorCause;
+    static DownloadError createNetworkError(FileDownloader.FileDownloadError networkErrorCause) {
+        String networkErrorMessage = "Network error, cannot download file. Cause: " + networkErrorCause.message();
         return new DownloadError(DownloadError.Type.NETWORK_ERROR_CANNOT_DOWNLOAD_FILE, networkErrorMessage);
     }
 
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadErrorFactory.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadErrorFactory.java
@@ -41,8 +41,8 @@ final class DownloadErrorFactory {
         return new DownloadError(DownloadError.Type.FILE_CANNOT_BE_WRITTEN, cannotWriteToFileMessage);
     }
 
-    static DownloadError createNetworkError(FileDownloader.FileDownloadError networkErrorCause) {
-        String networkErrorMessage = "Network error, cannot download file. Cause: " + networkErrorCause.message();
+    static DownloadError createNetworkError(FileDownloader.FileDownloadError error) {
+        String networkErrorMessage = "Network error, cannot download file. Cause: " + error.message();
         return new DownloadError(DownloadError.Type.NETWORK_ERROR_CANNOT_DOWNLOAD_FILE, networkErrorMessage);
     }
 
--- a/library/src/main/java/com/novoda/downloadmanager/FileDownloader.java
+++ b/library/src/main/java/com/novoda/downloadmanager/FileDownloader.java
@@ -29,7 +29,7 @@ public interface FileDownloader {
     }
 
     /**
-     * Represents the information of a {@link LiteFileDownloadError} that is accessible to clients.
+     * Represents error information that is accessible to clients.
      */
     interface Error {
 
--- a/library/src/main/java/com/novoda/downloadmanager/FileDownloader.java
+++ b/library/src/main/java/com/novoda/downloadmanager/FileDownloader.java
@@ -33,6 +33,8 @@ public interface FileDownloader {
      */
     interface Error {
 
+        int UNEXPECTED_ERROR_CODE = -1;
+
         /**
          * Use to create instances of {@link LiteFileDownloadError}.
          *
--- a/library/src/main/java/com/novoda/downloadmanager/FileDownloader.java
+++ b/library/src/main/java/com/novoda/downloadmanager/FileDownloader.java
@@ -27,4 +27,14 @@ public interface FileDownloader {
 
         void onDownloadFinished();
     }
+
+    interface FileDownloadError {
+
+        String rawRequest();
+
+        String message();
+
+        int errorCode();
+
+    }
 }
--- a/library/src/main/java/com/novoda/downloadmanager/FileDownloader.java
+++ b/library/src/main/java/com/novoda/downloadmanager/FileDownloader.java
@@ -28,13 +28,36 @@ public interface FileDownloader {
         void onDownloadFinished();
     }
 
+    /**
+     * Represents the information of a {@link LiteFileDownloadError} that is accessible to clients.
+     */
     interface FileDownloadError {
 
+        /**
+         * Use to create instances of {@link LiteFileDownloadError}.
+         *
+         * @param rawRequest the raw request from which the error occurs.
+         * @param message    any message associated with the error.
+         * @param errorCode  the code associated with the error.
+         * @return an instance of {@link FileDownloadError}.
+         */
+        static FileDownloader.FileDownloadError createFrom(String rawRequest, String message, int errorCode) {
+            return new LiteFileDownloadError(rawRequest, message, errorCode);
+        }
+
+        /**
+         * @return the raw request performed when the error occurred.
+         */
         String rawRequest();
 
+        /**
+         * @return a message representing the error that occurred.
+         */
         String message();
 
+        /**
+         * @return a short code representing the error that occurred.
+         */
         int errorCode();
-
     }
 }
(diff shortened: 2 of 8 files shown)


library/src/main/java/com/novoda/downloadmanager/FileDownloader.java
```diff
@@ -23,8 +23,41 @@
...
+         * @return an instance of {@link FileDownloadError}.
+         */
+        static FileDownloader.FileDownloadError createFrom(String rawRequest, String message, int errorCode) {
+            return new LiteFileDownloadError(rawRequest, message, errorCode);
+        }
+
+        /**
+         * @return the raw request performed when the error occurred.
+         */
+        String rawRequest();
```
danybony: This could be something like `requestUrl`?
Raw request sounds like the full okhttp request (headers and ecc...)

library/src/main/java/com/novoda/downloadmanager/FileDownloader.java
```diff
@@ -23,8 +23,41 @@
 
         void onBytesRead(byte[] buffer, int bytesRead);
 
-        void onError(String cause);
+        void onError(FileDownloadError cause);
```
blundell: is it an error or a cause? :-)

```
void onError(FileDownloadError error);
```

```
void onError(Cause cause);
```

library/src/main/java/com/novoda/downloadmanager/FileDownloader.java
```diff
@@ -23,8 +23,41 @@
...
 
         void onDownloadFinished();
     }
+
+    /**
+     * Represents the information of a {@link LiteFileDownloadError} that is accessible to clients.
+     */
+    interface FileDownloadError {
```
blundell: what if you called this `Error`? it's known from the scope what the error is for, makes it a bit more intuitive and helps peoples line length plus removes duplication

```
interface Error {
```
Would be used like:

```
static DownloadError createNetworkError(FileDownloader.Error networkErrorCause) {
```

so then perhaps the `FileDownloader` method would also change:
```
void onError(Error cause);
```
or if you need to be explicit at this point
```
void onFileDownloaderError(Error cause);
```

library/src/main/java/com/novoda/downloadmanager/FileDownloader.java
MechaRyRy: The problem is that the `FileDownloader` may not necessarily call the network, I wasn't really sure about `rawRequest`, in my mind it should probably be `fileDownloadPath` to keep it agnostic.

library/src/main/java/com/novoda/downloadmanager/DownloadErrorFactory.java
```diff
@@ -41,8 +41,8 @@ static DownloadError createCannotWriteToFileError(DownloadFileStatus downloadFil
         return new DownloadError(DownloadError.Type.FILE_CANNOT_BE_WRITTEN, cannotWriteToFileMessage);
     }
 
-    static DownloadError createNetworkError(String networkErrorCause) {
-        String networkErrorMessage = "Network error, cannot download file. Cause: " + networkErrorCause;
+    static DownloadError createNetworkError(FileDownloader.FileDownloadError networkErrorCause) {
+        String networkErrorMessage = "Network error, cannot download file. Cause: " + networkErrorCause.message();
         return new DownloadError(DownloadError.Type.NETWORK_ERROR_CANNOT_DOWNLOAD_FILE, networkErrorMessage);
```
blundell, line 46: This method reads as so:
- we are passed an error
- we get the msg out the error
- we make a new error of type NETWORK_ERROR_CANNOT_DOWNLOAD_FILE using the mesage

With the new change, we are passed the actual error, is there any point in redirecting that to create another error? i.e. could this whole method now be inlined? It could be potentially confusing to have two public api facing error types, that from here seem to have the same function? Perhaps thats the next step

library/src/main/java/com/novoda/downloadmanager/NetworkFileDownloader.java
```diff
@@ -30,7 +30,7 @@ public void startDownloading(String url, FileSize fileSize, Callback callback) {
             processResponse(callback, response, responseCode, url);
         } catch (IOException e) {
             Logger.e(e, "Exception with http request");
-            callback.onError(e.getMessage());
+            callback.onError(FileDownloadError.createFrom(url, e.getMessage(), -1));
```
blundell: what does -1 mean?

library/src/main/java/com/novoda/downloadmanager/FileDownloader.java
danybony: Wearing the "I don't know this library" user hat here:
TBH `fileDownloadPath` sounds like the path where the file is gonna be downloaded.
And looking at the `FileDownloader.startDownloading(..)` signature, the first param is named url...

library/src/main/java/com/novoda/downloadmanager/LiteFileDownloadError.java
```diff
@@ -0,0 +1,66 @@
+package com.novoda.downloadmanager;
+
+class LiteFileDownloadError implements FileDownloader.FileDownloadError {
```
blundell: easiest kotlin class you could for a start of a change ;)

library/src/main/java/com/novoda/downloadmanager/LiteFileDownloadError.java
```diff
@@ -0,0 +1,66 @@
...
+
+    private final String rawRequest;
+    private final String message;
+    private final int errorCode;
+
+    LiteFileDownloadError(String rawRequest, String message, int errorCode) {
```
blundell: what is "Lite" about this? Is it saying its a "Lite File Download" that has Error'd? Or is it saying a File Download has a "Lite Error" which I'm not sure what that is either?

blundell (changes requested):

library/src/main/java/com/novoda/downloadmanager/DownloadErrorFactory.java
MechaRyRy: It's the next step 😄 Just trying to make small changes that I can do within a timeframe, otherwise I end up losing the context of what I am doing.

library/src/main/java/com/novoda/downloadmanager/NetworkFileDownloader.java
MechaRyRy: Nothing at the moment, this will be iterated on before this is released. At the moment it probably means `UNEXPECTED` I guess.

library/src/main/java/com/novoda/downloadmanager/LiteFileDownloadError.java
MechaRyRy: `Lite` is used to represent all of our internal classes. We always introduce an interface on this project which represents the publicly available information and then hide everything else. Clients do not have the opportunity to mess with these classes, they only deal with interfaces.

library/src/main/java/com/novoda/downloadmanager/NetworkFileDownloader.java
blundell: would be good to name it 👍

library/src/main/java/com/novoda/downloadmanager/LiteFileDownloadError.java
blundell: hmm strange name.  Two things then:

Shouldn't the interface be declared public:

```
    interface FileDownloadError {
```

and

The `FileDownloadError` javadoc is referencing the internal class, but clients shouldn't be aware of it?
```
    /**
     * Represents the information of a {@link LiteFileDownloadError} that is accessible to clients.
     */
    interface FileDownloadError {
```

library/src/main/java/com/novoda/downloadmanager/DownloadErrorFactory.java
blundell: next step, ok 👍 

not sure its fair to say a small PR helps keep you context, small commits will help and a small tick list for you to tick off does. But the PR is for other people to understand and review the changes, perhaps a different scope for the PR i.e. after a bit more work, would allow others to see more context of the change - or alternatively leaving a TODO in the code here so that you know what the next refactoring step is and reviewers can see this is going to change. Right now if all worked stopped after this merge, its a bad place.

not asking for changes here, just for thought 👍

MechaRyRy: I see your point, I'll try and breakdown the steps a bit more and try and document them somewhere 😄 Normally I spend a great deal of time experimenting before picking an action but it might be better to stop and think, rather than just diving in.

library/src/main/java/com/novoda/downloadmanager/FileDownloader.java
MechaRyRy: That's a fair point. I'll rename both so that they are `requestUrl` then 😄

library/src/main/java/com/novoda/downloadmanager/LiteFileDownloadError.java
MechaRyRy: Good point about referencing the internal class 👍 

The interface, `FileDownloadError` is public, it's enclosed by another interface which is public.

library/src/main/java/com/novoda/downloadmanager/FileDownloader.java
MechaRyRy: cb5db7cb07ca44459a0c4306690d5a6a3bc06e26

library/src/main/java/com/novoda/downloadmanager/FileDownloader.java
MechaRyRy: 64cd3cdec65cb619936870a8e0e2772a2f862051

library/src/main/java/com/novoda/downloadmanager/FileDownloader.java
MechaRyRy: 00c717333fe92669b0513da0a2c89cb3d8aa49e3

library/src/main/java/com/novoda/downloadmanager/NetworkFileDownloader.java
MechaRyRy: d99707fa4fb0d34c3a1e7de7c740f8a4d8e0b57f

library/src/main/java/com/novoda/downloadmanager/LiteFileDownloadError.java
MechaRyRy: 6d130c92186cf7b0ac9f52a629ed4d84c2de7cd6

blundell (approved):

danybony (approved):
💯

PR #488, opened 2019-04-02
Downloaded notification re-shown

MechaRyRy:
## Problem 
The `DOWNLOADED` notification is show again whenever a network recovery mechanism is selected. 

When a network recovery mechanism is selected it requeues downloads against the `download-manager`. If the underlying `Map<DownloadBatchId, DownloadBatch>` already contains the `Batch` to queue then it will not add the new `Batch`. Unfortunately this `Batch` is the original and it was never updated with `notificationSeen=true` 

## Solution 
Add a method to the `DownloadBatchStatus` to update the `notificationSeen` status. This update will be propagated to the `Map<DownloadBatchId, DownloadBatch>` solving the issue.

### Map<DownloadBatchId, DownloadBatch>` 
Someone is bound to ask the question "why not update the map each time"...

If someone were to trigger the network recovery multiple times then we end up queueing multiple downloads against the `ExecutorService`. Which in itself is not a problem, it only is if you allow each `DownloadBatchId` to have a different state, which could potentially happen. Let's say we trigger a download, a network error then occurs and then we decide to delete it. **Well** then the download would finish **and then** the delete would occur. Instead you supply a single consistent instance for multiple threads, that way when the first terminates, even with it queuing a duplicate it will have a consistent state and just go straight to the end state. 

## Future work
- It would be good to avoid the cast. I'd really like to be able to supply an interface to our internal classes where they could receive the `InternalDownloadBatchStatus` but supply the more limited version to other client interfaces that supplies the `DownloadBatchStatus`.

### DIFF
--- a/library/src/main/java/com/novoda/downloadmanager/DownloadBatchStatusNotificationDispatcher.java
+++ b/library/src/main/java/com/novoda/downloadmanager/DownloadBatchStatusNotificationDispatcher.java
@@ -21,7 +21,7 @@ class DownloadBatchStatusNotificationDispatcher {
         this.downloadBatchIdNotificationSeen = downloadBatchIdNotificationSeen;
     }
 
-    void updateNotification(DownloadBatchStatus downloadBatchStatus) {
+    void updateNotification(InternalDownloadBatchStatus downloadBatchStatus) {
         if (downloadBatchStatus.notificationSeen()) {
             Logger.v("DownloadBatchStatus:", downloadBatchStatus.getDownloadBatchId(), "notification has already been seen.");
             return;
@@ -38,7 +38,7 @@ class DownloadBatchStatusNotificationDispatcher {
             Logger.v("start updateNotificationSeenAsync " + rawDownloadBatchId
                              + ", seen: " + NOTIFICATION_SEEN
                              + ", status: " + downloadBatchStatus.status());
-            notificationSeenPersistence.updateNotificationSeenAsync(downloadBatchStatus, NOTIFICATION_SEEN);
+            downloadBatchStatus.markAsNotificationSeen(notificationSeenPersistence);
         }
 
         notificationDispatcher.updateNotification(downloadBatchStatus);
--- a/library/src/main/java/com/novoda/downloadmanager/LiteDownloadBatchStatus.java
+++ b/library/src/main/java/com/novoda/downloadmanager/LiteDownloadBatchStatus.java
@@ -6,6 +6,7 @@ class LiteDownloadBatchStatus implements InternalDownloadBatchStatus {
 
     private static final long ZERO_BYTES = 0;
     private static final int TOTAL_PERCENTAGE = 100;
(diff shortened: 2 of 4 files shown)


library/src/main/java/com/novoda/downloadmanager/DownloadBatchStatusNotificationDispatcher.java
```diff
@@ -21,7 +21,7 @@
         this.downloadBatchIdNotificationSeen = downloadBatchIdNotificationSeen;
     }
 
-    void updateNotification(DownloadBatchStatus downloadBatchStatus) {
+    void updateNotification(InternalDownloadBatchStatus downloadBatchStatus) {
```
MechaRyRy, line 24: This class is not available to clients so passing `InternalDownloadBatchStatus` is ok here.

library/src/main/java/com/novoda/downloadmanager/LiteDownloadBatchStatus.java
```diff
@@ -149,6 +150,12 @@ public void markAsWaitingForNetwork(DownloadsBatchPersistence persistence) {
         updateStatusAsync(status, persistence);
     }
 
+    @Override
+    public void markAsNotificationSeen(DownloadsNotificationSeenPersistence persistence) {
```
MechaRyRy, line 154: Following the scheme of the others in this class now that we need to update both this status and the underlying DB version.

library/src/main/java/com/novoda/downloadmanager/LiteDownloadManagerDownloader.java
```diff
@@ -129,7 +129,11 @@ private DownloadBatchStatusCallback downloadBatchCallback(Map<DownloadBatchId, D
                     for (DownloadBatchStatusCallback callback : callbacks) {
                         callback.onUpdate(downloadBatchStatus);
                     }
-                    notificationDispatcher.updateNotification(downloadBatchStatus);
+
+                    DownloadBatch downloadBatch = downloadBatchMap.get(downloadBatchId);
+                    if (downloadBatch != null) {
```
MechaRyRy, line 134: We can't use the line 115's `downloadBatchStatus` as it is a copy of the underlying status. In order to update the `notificationSeen` we need the one that is modified and persisted by the `download-manager`, so we pull it from our source of truth which is our map. 

This knowledge about `copied` vs `real` download batch status is a bit hidden, we should spend sometime updating the `DownloadBatchStatusCallback` to make this relationship more explicit.

Dorvaryn:
Nice one ^^ As you mention there is a fair bit of hidden knowledge here with the cached vs updated values. But that can be worked on as an improvement. As a fix this is :+1:

Dorvaryn (approved):

MechaRyRy:
Thanks @Dorvaryn 😄 I've created a ticket to document this issue in a bit more detail and recommend some changes. https://github.com/novoda/download-manager/issues/489

zegnus (approved):

PR #490, opened 2019-04-02
2.3.1 Develop to Release

MechaRyRy:
We want to release the following:

#488 Downloaded notification re-shown
#479 Quick Returns 

Version 2.3.1 because this is patching a bug and not much else.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 allprojects {
-    version = '2.2.0'
+    version = '2.2.1'
 }
 
 def teamPropsFile(propsFile) {
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 allprojects {
-    version = '2.2.0'
+    version = '2.3.0'
 }
 
 def teamPropsFile(propsFile) {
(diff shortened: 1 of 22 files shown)


MechaRyRy:
There's an excessive number of commits because of a squash that was previously used. In case you are wondering.

zegnus:
@Mecharyry a squash collapses several commits into one, how come there is an excessive number of commits because of a previous squash?

MechaRyRy:
@zegnus I don't honestly know but it's the only thing that has changed between the past release and all others as far as I can tell. The changes are correct but the reported commits are just all present.

Dorvaryn:
If those previous changes were squashed into release before and release was not merged back into develop subsequently then those commits still exist into develop, hence why this PR shows them all. We can merge as is or we could try and merge release into develop first to see if this workflow would prevent such issues in the future.

zegnus:
Ah yes, I was thinking that same scenario

MechaRyRy:
I've tried merging release into develop thinking that might solve the issue, it did not 😄

Dorvaryn (approved):

Dorvaryn:
No point blocking this since the diff is correct. But we might need to define a proper workflow if we want to use squashes. Prs like this will make following what is going on a lot more complicated

zegnus (approved):

Issue #491, opened 2019-04-12
Is there any possibility to bypass the initial head request ?


mspmax:
Hi Guys, 
We are facing an issue at the moment as we do high volume downloads (eg: 100 or more per batch) so it takes a considerable amount of time to start downloading. This is mainly due to the fact that head requests are done sequentially. So is there a way to improve this or basically a way to provide the content length before downloading? and to add to this I noticed that it does another HEAD request on every download(parallel item) completion, any reason for that?

Cheers !

zegnus:
Hi @mspmax the problem with passing the length beforehand is that there is no guarantee that the values that you provide are going to be the ones reported by the endpoint causing corrupted downloads, I guess that this could be a risk to consider if we add this value.

In terms of the performance issue, parallel downloads are not currently supported and it would require a big change in the library. Also I am not 100% sure that enabling parallel downloads will speed up the batch because if we assume that one download takes the maximum bandwidth available and that the endpoint is capable of fulfilling this bandwidth, then parallel downloads won't improve much, but it will have a big hit on memory and resources usage of the library.

If you can give us an example where we can load stress the library we will try to think of easy ways to speed it up.

mspmax:
Hi @zegnus, thanks for the prompt reply. I'm bit confused here so when we add few items to a Batch it means those are not done in parallel but sequential? but still, there can be any number of items per batch right?

zegnus:
@mspmax you are correct. Everything is sequential and you can add any number of items per batch and any number of batches; the library will process them one by one until everything is finished and there is no performance impact on the number of items to download in terms of network calls, also you won't overload an endpoint with 100s of opened connections.

mspmax:
Thanks, but the main issue is around the head requests as it takes a considerable amount of time. Basically, for 100 list of items in the batch it sends 100 head requests before starting to download.

MechaRyRy:
> does another HEAD request on every download(parallel item) completion

That's interesting, I'll take a look at that particular problem and open another issue to represent it. 

> 100 list of items in the batch it sends 100 head requests before starting to download.

The issue here is not with the actual download @zegnus it's with the smaller requests that we do at the beginning that could be done in parallel. We could probably do this or allow the clients to specify the size initially and we will trust what they tell us, if it is incorrect from that point then the downloads will fail.

mspmax:
[...]

PR #492, opened 2019-04-17
DLM-491/File Size Requesting

MechaRyRy:
## Problem
As detailed in #491, sometimes clients need to process small requests quickly to avoid performance issues where a batch has a lot of small files to process. 

## Solution
As a workaround, until we investigate having parallel head requests for the file sizes we have opted to add an `Optional` `FileSize` to the initial batch builder. If specified then the head requests will be ignored. This is potentially dangerous, if the size is incorrect the batch will fail during the download. We are leaving this to the clients.

## Screen Capture

Before | Screenshot 2019-04-17 at 08 54 20
--- | ---
After | Screenshot 2019-04-17 at 08 54 44

#### On incorrect size specified
when_size_incorrect


MechaRyRy:
❌ DO NOT MERGE AT THE MOMENT ❌ I need to do some CI work before we merge this.

frapontillo (dismissed):

zegnus (dismissed):

ouchadam:
@Mecharyry is CI work still needed for this? Would be great to have this merged in & released!! 🙏

ouchadam (dismissed):

MechaRyRy:
@ouchadam thanks for reminding me about this! All done now.

danybony (approved):
💯

zegnus (approved):

PR #514, opened 2019-08-20
Delete empty folders

logicalusman:
### Problem
When deleting a download batch, all of the files get deleted but leaving the folders.

### Solution
The solution is to let the `DownloadBatch` delete folders after deleting files - this requires `DownloadBatch` to know the storage root of the batch. Helper and internal classes to `Batch` , such as, `LiteBatchFileBuilder`  knew the formation of batch storage path. That knowledge is now extracted to `BatchStorageRoot` so it is now used by the `DownloadBatch` to delete the batch folders.


library/src/main/java/com/novoda/downloadmanager/BatchStorageRoot.java
```diff
@@ -0,0 +1,21 @@
...
+
+    private BatchStorageRoot(StorageRoot storageRootPath, DownloadBatchId downloadBatchId) {
+        path = storageRootPath.path() + File.separator + downloadBatchId.rawId();
+    }
+
+    public String path() {
+        return path;
+    }
+
+}
```
logicalusman, line 46: Holds knowledge of storage root path for a batch. Previously this knowledge was hidden (private)

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -355,6 +357,7 @@ void delete() {
         for (DownloadFile downloadFile : downloadFiles) {
             downloadFile.delete();
         }
+        deleteDownloadDirs();
```
logicalusman: Once all files are deleted, the leftover dirs are deleted via this call. Remember it only happens for this download batch.

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -370,6 +373,28 @@ void delete() {
...
+                for (File child : nestedDirs) {
+                    deleteDir(child);
+                }
+            }
+        }
+        boolean deleted = batchRootDir.delete();
+        String message = String.format("File or Directory: %s deleted: %s", batchRootDir.getAbsolutePath(), deleted);
+        Logger.d(getClass().getSimpleName(), message);
+    }
+
```
logicalusman, line 414: A dir can only be deleted if it is empty (no files and dirs), which is why recursion is applied.

library/src/main/java/com/novoda/downloadmanager/LiteBatchFileBuilder.java
```diff
@@ -59,8 +57,7 @@ public BatchFileBuilder withSize(FileSize fileSize) {
     @Override
     public BatchBuilder apply() {
         String absolutePath = buildPath(
-                storageRoot.path(),
-                downloadBatchId.rawId(),
+                batchStorageRoot.path(),
                 path.or(""),
```
logicalusman, line 61: Batch storage root is formed by creating a dir, named after the value of `downloadBatchId`, inside `storageRoot`.
This knowledge is now extracted to `BatchStorageRoot`

library/src/main/java/com/novoda/downloadmanager/BatchStorageRoot.java
```diff
@@ -0,0 +1,21 @@
+package com.novoda.downloadmanager;
+
+import java.io.File;
+
+final class BatchStorageRoot {
```
MechaRyRy, line 5: I don't understand the necessity of adding this. The `StorageRoot` and the `BatchId` give you everything you need in order to be able to delete a `folder`. This feels like it is doing the same thing as the `Builder` is already doing but just wrapping it again into another class. ⚾️

library/src/test/java/com/novoda/downloadmanager/LiteBatchFileBuilderTest.java
```diff
@@ -12,22 +11,15 @@
...
     private final BatchFileBuilder liteBatchFileBuilder = new LiteBatchFileBuilder(
-            storageRoot,
-            downloadBatchId,
+            batchStorageRoot,
             ANY_NETWORK_ADDRESS
     ).withParentBuilder(batchBuilder);
 
-    @Before
-    public void setUp() {
-        given(downloadBatchId.rawId()).willReturn("my-movie");
```
MechaRyRy, line 25: ⚾️ This feels like we are introducing the `BatchStorageRoot` for no real benefit. Is the intention just to highlight explicitly that a batch root is composed of a `StorageRoot` and a `DownloadBatchId`?

library/src/main/java/com/novoda/downloadmanager/BatchStorageRoot.java
MechaRyRy: Ok. So this exists to keep the `storage root` and `downloadBatchId` tied together. Fair enough.

MechaRyRy: Should this be auto-valued given it is a model class?

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -370,6 +373,28 @@ void delete() {
                          + ", should be deleting");
     }
 
+    private void deleteDownloadDirs() {
+        BatchStorageRoot batchStorageRoot = BatchStorageRoot.with(() -> downloadBatchStatus.storageRoot(), downloadBatchStatus.getDownloadBatchId());
```
MechaRyRy: You can use a method reference for the `downloadBatchStatus.storageRoot`.

library/src/test/java/com/novoda/downloadmanager/LiteBatchFileBuilderTest.java
logicalusman: The construction of the batch root path was hidden inside `LiteBatchFileBuilder`. This would make it explicit and also it can be used/constructed somewhere if needed.

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -370,6 +373,28 @@ void delete() {
...
+        File batchRootDir = new File(batchStorageRoot.path());
+        if (batchRootDir.exists()) {
+            deleteDir(batchRootDir);
+        }
+    }
+
+    private void deleteDir(File batchRootDir) {
```
MechaRyRy: Avoid abbreviations to prevent confusion. 
```suggestion
    private void deleteDirectory(File batchRootDirectory) {
```

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
MechaRyRy: Deleting the files and then deleting the directories is probably overkill. You can just delete the directories because it will grab all of the files inside a directory and attempt to delete them.

MechaRyRy: Ah wait, this isn't possible because we need to update the DB persistence layer 🤔

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
logicalusman: ok

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -370,6 +373,28 @@ void delete() {
...
+            File[] nestedDirs = batchRootDir.listFiles();
+            if (nestedDirs != null) {
+                for (File child : nestedDirs) {
+                    deleteDir(child);
+                }
+            }
+        }
+        boolean deleted = batchRootDir.delete();
+        String message = String.format("File or Directory: %s deleted: %s", batchRootDir.getAbsolutePath(), deleted);
+        Logger.d(getClass().getSimpleName(), message);
```
MechaRyRy: Can we use `Logger.v` and follow the format for the other logs in this class, that way we follow the already accepted scheme.

library/src/test/java/com/novoda/downloadmanager/BatchStorageRootTest.java
```diff
@@ -0,0 +1,17 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+public class BatchStorageRootTest {
+
+    private final StorageRoot storageRoot = () -> "path/to/the/root/downloads";
+    private final DownloadBatchId downloadBatchId = () -> "batch_1";
+
+    @Test
+    public void path() {
```
MechaRyRy:
```suggestion
    public void appendsDownloadBatchIdToStorageRoot() {
```

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
logicalusman: Not really. It still need to go through same level of effort - A directory cannot be deleted if it has even a single file/directory inside.

library/src/test/java/com/novoda/downloadmanager/BatchStorageRootTest.java
```diff
@@ -0,0 +1,17 @@
...
+        BatchStorageRoot batchStorageRoot = BatchStorageRoot.with(storageRoot, downloadBatchId);
+        String expected = "path/to/the/root/downloads/batch_1";
+        Assert.assertEquals(batchStorageRoot.path(), expected);
```
MechaRyRy: We use Truth assertions in `downloadManager`.

library/src/test/java/com/novoda/downloadmanager/BatchStorageRootTest.java
```diff
@@ -0,0 +1,17 @@
...
+
+public class BatchStorageRootTest {
+
+    private final StorageRoot storageRoot = () -> "path/to/the/root/downloads";
+    private final DownloadBatchId downloadBatchId = () -> "batch_1";
+
+    @Test
+    public void path() {
+        BatchStorageRoot batchStorageRoot = BatchStorageRoot.with(storageRoot, downloadBatchId);
+        String expected = "path/to/the/root/downloads/batch_1";
```
MechaRyRy: I would inline this, it doesn't have any benefit extracting it.

library/src/test/java/com/novoda/downloadmanager/BatchStorageRootTest.java
```diff
@@ -0,0 +1,17 @@
...
+import org.junit.Test;
+
+public class BatchStorageRootTest {
+
+    private final StorageRoot storageRoot = () -> "path/to/the/root/downloads";
+    private final DownloadBatchId downloadBatchId = () -> "batch_1";
+
+    @Test
+    public void path() {
+        BatchStorageRoot batchStorageRoot = BatchStorageRoot.with(storageRoot, downloadBatchId);
```
MechaRyRy, line 14: Spacing between each line in this test so that it reads like

```
given

when

then
```

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
MechaRyRy: This has issues if the file is downloading and you decide to delete it 😄 if we are still downloading then it will try to write to a file that no longer exists or is in the process of being deleted, which will result in an error on the client side. 

error_when_writing

MechaRyRy: Getting this error will be hit and miss because it will be a race condition.

MechaRyRy (changes requested):
I'm worried about the potential for a race condition when performing a delete now.

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
logicalusman: I used the one similar to  `PathBasedFilePersistence.delete()` , so all deleted files/dir path logs to same level i.e. `debug`

library/src/test/java/com/novoda/downloadmanager/BatchStorageRootTest.java
logicalusman: ok

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
MechaRyRy: Ok. Fair enough 😄

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
logicalusman: ok

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
logicalusman: I see. However, this is the not handled by the sdk originally and not issue with this PR. I shall liaise with @zegnus .

library/src/test/java/com/novoda/downloadmanager/BatchStorageRootTest.java
logicalusman: ok

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
MechaRyRy: I don't understand your comment. We wouldn't be getting this error before because of how the states operate. Can you please elaborate.

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
```diff
@@ -370,6 +373,28 @@ void delete() {
...
+        BatchStorageRoot batchStorageRoot = BatchStorageRoot.with(downloadBatchStatus::storageRoot, downloadBatchStatus.getDownloadBatchId());
+        File batchRootDir = new File(batchStorageRoot.path());
+        if (batchRootDir.exists()) {
+            deleteDirectory(batchRootDir);
+        }
+    }
+
+    private void deleteDirectory(File batchRootDirectory) {
+        if (batchRootDirectory.isDirectory()) {
+            File[] nestedDirs = batchRootDirectory.listFiles();
```
MechaRyRy: `nestedDirectories`.

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
logicalusman: Oh, I know what you meant. Ignore my earlier comment. Need to check the download state before deleting.

library/src/main/java/com/novoda/downloadmanager/BatchStorageRoot.java
zegnus: I think we didn't add AutoValue in this project no?

MechaRyRy: Oh 🤔 probably not actually. Well in that case, `equals` and `hashcode` and `toString` 😂

logicalusman: Added those three methods.

library/src/main/java/com/novoda/downloadmanager/DownloadBatch.java
logicalusman: Fixed.

library/src/test/java/com/novoda/downloadmanager/BatchStorageRootTest.java
```diff
@@ -0,0 +1,18 @@
...
+    private final StorageRoot storageRoot = () -> "path/to/the/root/downloads";
+    private final DownloadBatchId downloadBatchId = () -> "batch_1";
+
+    @Test
+    public void appendsDownloadBatchIdToStorageRoot() {
+        BatchStorageRoot batchStorageRoot = BatchStorageRoot.with(storageRoot, downloadBatchId);
+
+        assertThat(batchStorageRoot.path()).isEqualTo("path/to/the/root/downloads/batch_1");
+    }
+}
```
zegnus: needs a new line

zegnus (dismissed):

library/src/test/java/com/novoda/downloadmanager/BatchStorageRootTest.java
logicalusman: @zegnus  you mean new line after last `}` ?

logicalusman: Fixed.

MechaRyRy (approved):

PR #523, opened 2019-10-11
Allow for identifying when download failed due to violating a Downloa…

Konrad-Morawski:
…dBatchRequirementRule (as well as the specific rule that caused the error)

### DIFF
--- a/demo-simple/src/main/java/com/novoda/downloadmanager/demo/DownloadBatchSizeRequirementRule.java
+++ b/demo-simple/src/main/java/com/novoda/downloadmanager/demo/DownloadBatchSizeRequirementRule.java
@@ -5,6 +5,8 @@ import com.novoda.downloadmanager.DownloadBatchStatus;
 
 public class DownloadBatchSizeRequirementRule implements DownloadBatchRequirementRule {
 
+    public final static int CODE = DownloadBatchRequirementRule.class.hashCode();
+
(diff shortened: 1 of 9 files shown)


hal9002:

demo-simple/src/main/java/com/novoda/downloadmanager/demo/DownloadBatchSizeRequirementRule.java
```diff
@@ -5,6 +5,8 @@
 
 public class DownloadBatchSizeRequirementRule implements DownloadBatchRequirementRule {
 
+    public final static int CODE = DownloadBatchRequirementRule.class.hashCode();
```
Konrad-Morawski, line 8: For example (it's just a demo)

library/src/main/java/com/novoda/downloadmanager/DownloadError.java
```diff
@@ -33,35 +45,37 @@ public String message() {
         return message;
     }
 
+    @Nullable
+    public Integer violatedRuleCode() {
```
Konrad-Morawski, line 49: At the end of the chain, we'd retrieve the info (about a rule violation) from here

Konrad-Morawski:
A false start on my part, disregard it for now please. Sorry

CC @hal9002

Issue #521, opened 2019-09-30
Migrate to WorkManager


ouchadam:
Evernote android-job has been deprecated (and has some internal ANR issues)

https://github.com/evernote/android-job

[WorkManager](https://developer.android.com/topic/libraries/architecture/workmanager) is the recommended replacement

PR #532, opened 2020-06-28
NOS-521: Migrate to WorkManager

ouchadam:
Moves away from https://github.com/evernote/android-job in favour of `WorkManager` #521

**Breaking change** clients must provide a `DelegatingWorkerFactory` for the library to inject its own factory into, otherwise network recovery will not work

`ConfigurationException` will be thrown if the `DelegatingWorkerFactory` is not found on the application.

```java
public class MyApplication extends Application implements Configuration.Provider {
    private final  Configuration workManagerConfig = new Configuration.Builder()
            .setWorkerFactory(new DelegatingWorkerFactory())
            .build();
    @Override
    public Configuration getWorkManagerConfiguration() {
        return workManagerConfig;
    }
}
```

This functionality has been used in production for about 6 months without issue from my my side 🌻 

| Network recovery |
| --- | 
after-dlm


library/src/main/java/com/novoda/downloadmanager/DownloadManagerBuilder.java
```diff
@@ -33,7 +39,7 @@
 import static com.novoda.downloadmanager.DownloadBatchStatus.Status.ERROR;
 import static com.novoda.downloadmanager.DownloadBatchStatus.Status.PAUSED;
 
-@SuppressWarnings("PMD.ExcessiveImports")
+@SuppressWarnings({"PMD.ExcessiveImports", "PMD.GodClass"})
```
ouchadam, line 42: not ideal but the change to fix this would be bigger than the workmanager migration itself 😅

zegnus (approved):

danybony (approved):
Good one, thanks @ouchadam ! <3

PR #533, opened 2020-07-29
Removing Jetifier

ouchadam:
Same as https://github.com/novoda/merlin/pull/200 

Removes the `jetifier` config and allows clients of the library to avoid _jetificiation_ 

- Inlines the `MerlinBeard` usage as `merlin` current enforces the jetifier but the library hasn't been updated in a while, happy to revert the change if the above PR is merged and released 😛 

- Removes the stetho dependency within the library, this wasn't being used and probablyyy shouldn't be part of the production classpath

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -72,8 +72,6 @@ dependencies {
     annotationProcessor "androidx.room:room-compiler:$roomVersion"
     api "androidx.work:work-runtime:2.2.0"
 
-    implementation 'com.novoda:merlin:1.2.1'
-    implementation 'com.facebook.stetho:stetho:1.5.1'
     implementation 'com.squareup.okhttp3:okhttp:4.2.2'
     implementation "androidx.room:room-runtime:$roomVersion"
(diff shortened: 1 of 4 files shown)


library/build.gradle
```diff
@@ -72,8 +72,6 @@ dependencies {
     annotationProcessor "androidx.room:room-compiler:$roomVersion"
     api "androidx.work:work-runtime:2.2.0"
 
-    implementation 'com.novoda:merlin:1.2.1'
-    implementation 'com.facebook.stetho:stetho:1.5.1'
```
blundell, line 76: use `debugImplementation` just incase someone is using this?

ouchadam: I'm not sure a library can declare and export variant dependencies, it would be up to the client to include `debugImplementation download-manager`

I'd argue that the library itself shouldn't include stetho at all!

zegnus (approved):

ouchadam:
thanks! 💯

PR #536, opened 2022-12-16
Update sdk versions

JozefCeluch:
Updates target and compile SDK to 33 and minSdk to 21 with the smallest amount of changes necessary to make the demo app work.

I updated min SDK to 21 because that's the minimum version [required by OkHttp](https://developer.squareup.com/blog/okhttp-3-13-requires-android-5/). 

The update of target/compile SDK required to update the `androidx.work:work-runtime` to latest version which then forced the change of the provider configuration in the manfiest.

The demo app seems to work fine with these changes.


danybony (dismissed):
👌

build.gradle
```diff
@@ -8,29 +8,30 @@ def teamPropsFile(propsFile) {
 }
 
 ext {
-    checkstyleVersion = '8.5'
-    findbugsVersion = '3.0.1'
-    pmdVersion = '5.8.1'
+    checkstyleVersion = '8.8'
+//    findbugsVersion = '3.0.1'
```
kenwolf: Can we delete this?

kenwolf (dismissed):

